PR #1, opened 2020-09-23
[PF-109] initial RBS hello world spring APP

yonghaoy:
Cloned from kernel-service-poc repo:
https://github.com/DataBiosphere/kernel-service-poc


src/main/java/bio/terra/TEMPLATE/app/Main.java
```diff
@@ -0,0 +1,14 @@
+package bio.terra.TEMPLATE.app;
```
wchamber, line 1: should we go ahead and fix all the TEMPLATE in the initial commit?

src/main/java/bio/terra/TEMPLATE/app/configuration/ApplicationConfiguration.java
```diff
@@ -0,0 +1,142 @@
...
+  private String dbPassword;
+  private String dbUri;
+  // These properties control code in the StartupInitializer. We would not use these in production,
+  // but they
+  // are handy to set for development and testing. There are only three interesting states:
+  // 1. initialize is true; upgrade is irrelevant - initialize and recreate an empty database
+  // 2. initialize is false; upgrade is true - apply changesets to an existing database
+  // 3. initialize is false; upgrade is false - do nothing to the database
+  private boolean dbInitializeOnStart;
+  private boolean dbUpgradeOnStart;
```
wchamber, line 44: can we use the Janitor's variable naming? I find that more intuitive. Maybe we should contribute back to the kernel-poc.

src/test/resources/application.properties
```diff
@@ -0,0 +1,6 @@
+server.port=8080
```
wchamber, line 1: let's also use the yaml file and naming conventions that Greg changed to the Janitor. Also consider modifying POC first, then cloning.

wchamber (changes requested):
Consider whether we should create the shared library and start using that for RBS. That's a bigger task, but I think we should consider taking it on now. It only gets harder to do the more services we add first.
At least, let's update the poc with some of the improvements we've made to the janitor.

src/main/java/bio/terra/TEMPLATE/app/configuration/ApplicationConfiguration.java
wchamber: Also consider an enum with 3 values instead of 2 booleans with 3 valid combinations.

src/main/java/bio/terra/TEMPLATE/app/Main.java
yonghaoy: ah, right. We can do it now

src/test/resources/application.properties
yonghaoy: done for rbs

src/main/java/bio/terra/TEMPLATE/app/configuration/ApplicationConfiguration.java
yonghaoy: Changed to Janitor style. And for the enum vs booleans. I'd prefer do it later. There ares some things I am not sure yet:
1: How the config would looks like. Boolean is easier. And Enum would become String, and our configuration need extra logic about parsing string to enums and validation check.
2: What would be the 3 enum naming and meaning? Boolean with the javadoc is easer to understand, but naming enums to reflect the three things is challenging for me, I might need your help.
3: Not sure we need the 3): initialize is false; upgrade is false - do nothing to the database
Anyway, changed to Janitor structure for now?

yonghaoy:
Hi @wchamber I changed to Janitor as template with swager codegen, new config style, and all classes in the code.
And as synced offline, we will not modify poc code

src/main/java/bio/terra/TEMPLATE/app/configuration/ApplicationConfiguration.java
wchamber: Leaving it alone sounds good. (I think Spring automatically tries to do Enum.valueOf from the string).

build.gradle
```diff
@@ -0,0 +1,137 @@
...
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+}
+
+plugins {
+    id 'java'
+    id 'idea'
+    id 'com.google.cloud.tools.jib' version '2.5.0'
```
wchamber, line 12: do we need the jib stuff somewhere (w/ or w/out the profiler agent stuff)? Totally fine to follow up in a later PR - just not sure this builds an image without it. Up to you.

src/main/resources/api/service_openapi.yaml
```diff
@@ -0,0 +1,97 @@
...
+info:
+  title: RBS API
+  description: |
+    Resource Buffering Service to allocate and buffer cloud resources
+  version: 0.0.1
+
+servers:
+  # When developing locally/with Skaffold, add local entry with url prefix here:
+  # E.g. url: /env-terra0rbs
+  - url: http://localhost:8080
```
wchamber, line 11: This is different then the janitor where we have `- url: /`. I don't know which is supposed to be better.

README.md
```diff
@@ -1,2 +1,49 @@
...
+# Deployment
+## Run Locally
+Set executable permissions:
+```
+chmod +x gradlew
+```
+
+To spin up the local postgres, run:
+```
+local-dev/run_postgres.sh start
```
wchamber, line 13: Add a todo to add the local-dev files to run the  postgres docker and run_local?

wchamber (approved):
Thanks!

README.md
yonghaoy: Ah, sorry, I had those scripts in my local, the poc's `.gitignore` is not good. Updated `.gitginore` and uploaded all local-dev scripts

src/main/resources/api/service_openapi.yaml
yonghaoy: I think they are the same. Changed to `- url: /`

build.gradle
yonghaoy: I planed to add jib stuff when we add profiler agent later?

PR #4, opened 2020-09-29
[PF-125] Initial Pool Config for RBS

yonghaoy:
1.  Add initial gcp project schema and a Aou resource config
2. Add a unit test to verify configs in all environment are correct.


build.gradle
```diff
@@ -90,16 +95,44 @@ swaggerSources {
...
+sourceSets {
+    main {
+        java {
+            srcDir "${openapiTargetDir}/src/main/java"
+        }
+    }
+    test {
+        // Allow unit tests to directly load config files.
+        resources {
+            srcDir "config/"
```
wchamber, line 132: I think the config directory should just go under `resources`. e.g. `resources/config`

src/main/config/resource_schema.yaml
```diff
@@ -0,0 +1,48 @@
...
+      required:
+        - projectIdPrefix
+      properties:
+        projectIdPrefix:
+          type: string
+        parentFolderId:
+          type: string
+        billingAccount:
+          type: string
+        enabledApis:
```
wchamber, line 44: add descriptions for all of the properties

src/main/resources/api/service_openapi.yaml
```diff
@@ -78,6 +78,37 @@ components:
                 items:
                   type: string
 
+    Pools:
```
wchamber, line 81: why is the pools in service_openapi, but the resource configs in resource schema?

src/test/java/bio/terra/rbs/resource/PoolSchemaTest.java
```diff
@@ -0,0 +1,78 @@
...
+      for (PoolConfig poolConfig : pools.getPoolConfigs()) {
+        if (!resourceNameVersionMap.containsKey(poolConfig.getResourceConfigName())
+            || !resourceNameVersionMap
+                .get(poolConfig.getResourceConfigName())
+                .equals(poolConfig.getVersion())) {
+          fail(
+              String.format(
+                  "ResourceConfig not found for name: %s, version:%s, folder: %s",
+                  poolConfig.getResourceConfigName(), poolConfig.getVersion(), folderName));
+        }
```
wchamber, line 71: I think most of this ends up in non-test code eventually, but it's ok to start with it here in tests.

wchamber (changes requested):

build.gradle
yonghaoy: I started with resources/ then I checked the AoU code with Greg. This is what they have, and a better pattern I think based on Greg's feedback. I am fine with any. Let's check with @gjuggler ?

src/main/resources/api/service_openapi.yaml
yonghaoy: Pools is special thing for RBS. `ResourceConfig` is terra common thing. 
In my mind model, RBS just uses the common` terra resource config` to construct RBS pool.

src/test/java/bio/terra/rbs/resource/PoolSchemaTest.java
yonghaoy: Yup, I will move some of things in my next loading config PR

build.gradle
gjuggler: Sorry, I think we'd gotten our lines crossed. I think I was trying to make the point that Isaac was recently re-working Workspace Manager to house its application YAML config files under the resouces/config directory for tidiness, rather than directly under the resources directory ([link](https://github.com/DataBiosphere/terra-workspace-manager/blob/afb7ed97d4bd17fb30118030aff9fe8f1fb6c66d/src/main/resources/config/application-alpha.yml)). Not relevant to this PR.

yonghaoy: ah, sorry, I heard you mentioned the config/ dir and I found that is AoU has here:
https://github.com/all-of-us/workbench/blob/master/api/build.gradle#L312
And 
https://github.com/all-of-us/workbench/tree/master/api/config? 
Let me change it back. Thanks!

yonghaoy: Done

gjuggler: Oh yeah – that's a totally different thing. AoU uses a mostly-homegrown setup for loading much of its application config. It has some benefits & drawbacks... wld be happy to chat through those at some point, though I think sticking with what Spring supports by default is still probably the right approach.

src/main/config/resource_schema.yaml
yonghaoy: Done

src/test/java/bio/terra/rbs/resource/PoolSchemaTest.java
```diff
@@ -28,7 +28,7 @@
 public class PoolSchemaTest {
   /** PList of pool config folders for all environments, e.g. prod, staging, dev. */
   private static final List<String> POOL_CONFIG_FOLDERS =
-      ImmutableList.of("src/main/config/pool-config/dev/");
+      ImmutableList.of("src/main/resources/config/dev/");
```
wchamber, line 31: I think `"config/dev"` should work here. I think it's generally better to access what's under `resources/` without any preceeding directories.

wchamber (approved):

src/test/java/bio/terra/rbs/resource/PoolSchemaTest.java
yonghaoy: Having different folders for pool config and resource config will make our code logic simpler. If we put everything right under `config/dev`. Our code need to have the file name check like:
```
if(fileName.equals(poolConfig)) {
   // poolSchema process
} else {
  // resource config stuff.
}
```
And not sure if we need to introduce some extra naming patterns to restrict resource config name then.
In the current model it will work if we have extra stuff(e.g., config pull scripts, other yaml files) we want to put under dev/ directory. resource-config will only contains resource configs. And `dev/` root may have pool config with extra stuff.

wchamber: I think you misunderstood my comment. I was suggesting replacing `"src/main/resources/config/dev/"` with `"config/dev/". That's all.

I'm not really worried about our ability to distinguish pool/resource configs from all the extra stuff. Once we have extra stuff, let's worry about that with more directories or naming conventions.

yonghaoy: Done

src/main/resources/api/service_openapi.yaml
```diff
@@ -78,6 +78,37 @@ components:
...
+        - size
+        - resourceConfigName
+      properties:
+        poolName:
+          type: string
+        version:
+          type: string
+        size:
+          type: integer
+        resourceConfigName:
```
gjuggler, line 108: [no action] Just thinking aloud. I suspect eventually we may find it convenient for a pool config to be able to specify a JSON-like structure of overrides that will get merged with the named resource config in order to produce the final pool configuration. This may eventually give us the right balance of customizability vs. conciseness. But this should become clear enough once things are in actual usage anyway...

src/main/resources/api/service_openapi.yaml
```diff
@@ -78,6 +78,37 @@ components:
...
+    PoolConfig:
+      description: |-
+        Terra RBS Reource Pool Config
+      required:
+        - poolName
+        - version
+        - size
+        - resourceConfigName
+      properties:
+        poolName:
```
gjuggler, line 102: [non-opt] Can you add a description field to each of these properties? This should be the canonical place where these fields are defined, so it's important to document what they mean sooner than later.

(Concrete question I have: for the "version" field, is that meant to be a version string for the pool config itself, or is it meant to specify which version of the resource config to load? Ideally a description string would clarify that.)

src/main/resources/config/dev/pool_schema.yml
```diff
@@ -0,0 +1,7 @@
+# RBS Pools Schema in dev encironment
```
gjuggler, line 1: Terminology nit: the object is named PoolConfig, so for consistency this file should probably be named pool_config.yml and this comment should be updated.

src/main/resources/api/service_openapi.yaml
```diff
@@ -78,6 +78,37 @@ components:
                 items:
```
gjuggler, line 78: Kind of silly question on naming: do we have a clear pattern for kebab-case vs. snake_case for YAML files like this? Is there a reason you went one way or another? Just curious...

src/main/resources/config/dev/resource-config/aou_ws_0_0_1.yml
```diff
@@ -0,0 +1,27 @@
...
+    - "cloudtrace.googleapis.com"
+    - "containerregistry.googleapis.com"
+    - "dataflow.googleapis.com"
+    - "dataproc.googleapis.com"
+    - "genomics.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
+    - "dns.googleapis.com"
```
gjuggler, line 27: nit: alphabetize this entry

.github/workflows/test.yml
```diff
@@ -0,0 +1,74 @@
+# This workflow will build a Java project with Gradle
```
gjuggler, line 1: Was this workflow file copied from some source (e.g. in POC repo or Janitor)? If so, it may be worth adding a comment line indicating its provenance, just for posterity.

src/main/resources/config/resource_schema.yaml
```diff
@@ -0,0 +1,60 @@
+openapi: 3.0.0
+info:
+  title: Terra Resoure Buffer Service Resource Config
```
gjuggler, line 3: spelling nit: s/resoure/resource/

src/main/resources/config/dev/resource-config/aou_ws_0_0_1.yml
```diff
@@ -0,0 +1,27 @@
+# All Of Us buffered workspace template
+---
+configName: "aou_ws_resource_config"
+version: "0.0.1"
+gcpProjectConfig:
+  projectIdPrefix: "aou_rw_dev"
```
gjuggler, line 6: [nit] Not that it matters a whole lot at this stage, but AoU uses dashes in its prefix. And their dev environment is called test. So probably "aou-rw-test" would be best.

src/main/resources/config/resource_schema.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+    GcpProjectConfig:
+      description: |-
+        Represents a GCP Project resource with extra cloud resource setup when creating project.
+      required:
+        - projectIdPrefix
+      properties:
+        projectIdPrefix:
+          description: |-
+            The created project name prefix.
```
gjuggler, line 44: nit: Google calls this "project ID", so we should snap to that terminology. It may also be helpful to give an example prefix and corresponding full project ID, e.g. "aou-rw-test" and "aou-rw-test-abc123" (or whatever we think our initial naming scheme may be... though this may change, as we're still having some discussion w/ product on what this should look like).

src/main/resources/config/resource_schema.yaml
```diff
@@ -0,0 +1,60 @@
...
+          description: |-
+            The parent folder id of created project.
+          type: string
+        billingAccount:
+          description: |-
+            The billing account id the project associated to.
+          type: string
+        enabledApis:
+          description: |-
+            List of APIs enable
```
gjuggler, line 56: wording nit: "APIs to enable"

gjuggler (approved):
Handful of nitpicks, but this looks like a good start!

src/main/resources/config/resource_schema.yaml
yonghaoy: Good point! Done

src/main/resources/config/dev/resource-config/aou_ws_0_0_1.yml
yonghaoy: Done

.github/workflows/test.yml
yonghaoy: Done

src/main/resources/api/service_openapi.yaml
yonghaoy: Good point, I will keep that in mind! For now, as we discussed, rbs loads folder of files, and the folder is the env variable.

src/main/resources/api/service_openapi.yaml
yonghaoy: Done for description. And as we discussed, for now, I remove version from both resource config and pool config.
So updating pool config workflow is something like:
1. Add new resource config file with different name.
2. Add new pool config like 
```
poolName: pool_v2
resourceConfigName: newConfigName
```
3. RBS treat this is creating pool not updating existing pool. 

That is what I current thought, but that need more thoughts and discussion. I owe a README section, and will add when final decision is made.

PR #5, opened 2020-10-02
[PF-126] Load PoolConfig from file and support pool creation based on config change

yonghaoy:


src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,148 @@
...
+    Map<String, Integer> activatePoolNameSizeMap =
+        activatePools.stream().collect(Collectors.toMap(Pool::name, Pool::size));
+    Map<String, Integer> parsedPoolNameSizeMap =
+        parsedPoolConfigs.keySet().stream()
+            .collect(Collectors.toMap(PoolConfig::getPoolName, PoolConfig::getSize));
+    MapDifference<String, Integer> mapDifference =
+        Maps.difference(activatePoolNameSizeMap, parsedPoolNameSizeMap);
+
+    Map<String, Integer> poolsToCreate = mapDifference.entriesOnlyOnRight();
+    Map<String, Integer> poolsToDelete = mapDifference.entriesOnlyOnLeft();
```
wchamber, line 70: Could we just use `Set<String>` of sets of pool names instead of `Map<String, Integer>`? I know eventually we need to update pool sizes, but I think it's still simpler to look at the 3 cases (in parsed only, in database only, in both) and handle them differently.

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,148 @@
...
+  public void initialize() {
+    logger.info("Pool config update enabled: " + poolConfiguration.isUpdatePoolOnStart());
+    if (poolConfiguration.isUpdatePoolOnStart()) {
+      initializeFromConfig();
+    }
+  }
+
+  @VisibleForTesting
+  void initializeFromConfig() {
+    List<Pool> activatePools = rbsDao.retrievePools(PoolStatus.ACTIVE);
```
wchamber, line 57: I think this means we can never re-activate a pool. Is that what we want? That could lead to weird states where multiple pools have the same names.
(should unique names be a database constraint?)

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,148 @@
...
+    Map<PoolConfig, ResourceConfig> parsedPoolConfigs =
```
wchamber, line 58: I'm not sure this `Map` is the most helpful. in createPools we have ` Map<String, Integer> poolsToCreate, Map<PoolConfig, ResourceConfig> parsedPools` where we need to go from the key of poolsToCreate to the key+value of parsed pools.
WDYT about List<PoolConfig> and also a Map<ResourceConfigId, ResourceConfig>? Or do we want to make even stronger constraints on PoolConfig name uniqueness?

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,148 @@
...
+                .build();
+        pools.add(createdPool);
+      }
+    }
+    rbsDao.createPools(pools);
+  }
+
+  /** Parse and load {@link PoolConfig} and {@link ResourceConfig} from file. */
+  @VisibleForTesting
+  public static Map<PoolConfig, ResourceConfig> loadPoolConfig(String folderName) {
```
wchamber, line 105: Why is this `public`? Should PoolSchemaTest be in the same package as this?

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,148 @@
...
+                .resourceType(typeVisitor.accept(poolConfig.getValue()))
+                .status(PoolStatus.ACTIVE)
+                .build();
+        pools.add(createdPool);
+      }
+    }
+    rbsDao.createPools(pools);
+  }
+
+  /** Parse and load {@link PoolConfig} and {@link ResourceConfig} from file. */
```
wchamber, line 103: It feels a little funny to me to have this deserialization logic in the same class that reconciles what's loaded with what's in the database. Deserialization feels like a lower level concern, and crossing levels is a bit of a smell. Maybe split the folder/file/parsing into a separate class? Fine to rely on the concrete implementation for now, but I think eventually we'd want to inject the parsed config into `PoolsService` for testing reconciling parsed/database pools.

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,148 @@
...
+    ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory()).findAndRegisterModules();
+    // Parse PoolConfig
+    Pools pools;
+    try {
+      pools =
+          objectMapper.readValue(
+              classLoader.getResourceAsStream(folderName + "/" + POOL_SCHEMA_NAME), Pools.class);
+    } catch (IOException e) {
+      throw new RuntimeException(
+          String.format("Failed to parse pool schema for folder %s", folderName), e);
```
wchamber, line 116: I think we should throw IOException from our load/parse function. If we want to wrap the message/exception with what file/folder operation failed that's fine. But let's a higher level decide if it should be an unchecked exception.

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+    rbsDao.createPools(pools);
+  }
+
+  /** Parse and load {@link PoolConfig} and {@link ResourceConfig} from file. */
+  @VisibleForTesting
+  public static Map<PoolConfig, ResourceConfig> loadPoolConfig(String folderName) {
+    ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+    ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory()).findAndRegisterModules();
+    // Parse PoolConfig
```
wchamber, line 108: `// Do something` comments are generally a good indicator that you want helper functions, e.g.
`Pools parsePools(String folderName)`
ditto others

src/main/java/bio/terra/rbs/service/pool/ResourceConfigTypeVisitor.java
```diff
@@ -0,0 +1,25 @@
...
+@Component
+public class ResourceConfigTypeVisitor implements ResourceConfigVisitor<ResourceType> {
+  @Override
+  public ResourceType visit(GcpProjectConfig resource) {
+    return ResourceType.GOOGLE_PROJECT;
+  }
+
+  @Override
+  public ResourceType noResourceVisited(ResourceConfig resource) {
+    throw new InvalidPoolConfigException(String.format("Invalid resource config for %s", resource));
```
wchamber, line 19: This smells to me. The ResourceConfigVisitor<R> signature promises that we can always return an R, but this throws an unchecked exception. This is surprising behavior from just looking at signatures. Consider whether this should implement `ResourceConfigVisitor<Optional<ResourceType>>` or if we want to change ResourceConfigVisitor to do something else for this null case.

src/main/java/bio/terra/rbs/service/pool/ResourceConfigTypeVisitor.java
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  public ResourceType accept(ResourceConfig resourceConfig) {
+    return ResourceConfigVisitor.visit(resourceConfig, this);
```
wchamber, line 23: Is `new ResourceConfigTypeVisitor().accept(resourceConfig)` any easier than `ResourceConfigVisitor.visit(resourceConfig, new ResourceConfigTypeVisitor())`? Do we want a function with a default implementation on the interface?

wchamber (changes requested):

src/main/java/bio/terra/rbs/service/pool/PoolService.java
yonghaoy: Good point, after re-activate a pool(or just create a pool with the duplicate name), there would be multiple pools with the same name. 
Changed name in db to unique

src/main/java/bio/terra/rbs/service/pool/PoolService.java
yonghaoy: I feel resourceSchemaTest shouldn't be this package because that test is to validate resource configs are valid for all environment. The actual package it want to test is `src/main/resources`

src/main/java/bio/terra/rbs/service/pool/PoolService.java
yonghaoy: I think I am still not sure the name and package this class should be. It's name is PoolService and will handle Pool Update/Delete/Create.
The only function it does in the short term is the `initialize` method, and it will:

1. deserialize config
2. Calculate the config change
3. Do pool update. 
 deserialize is here because I don't feel this is lower level since it is will just used by this function. 
Another options is we change PoolService to PoolCompoment just like the StairwayCompoment, and has` initialize()` function.  I have it here because I feel it is similar to MigirateService, might be good to chat offline about this?

src/main/java/bio/terra/rbs/service/pool/ResourceConfigTypeVisitor.java
yonghaoy: ah, good point, having this because i though that is a standard Java visitor pattern... Removed this.

src/main/java/bio/terra/rbs/service/pool/PoolService.java
yonghaoy: created a helper class to have file processing logic.

yonghaoy:
@wchamber  Hi Will, all comments are addressed, and things change a lot.
Chang log:
1. Remove UUID PoolId and rename PoolName to poolId, now we have PoolId as string and loaded from file
2. Change the calculation logic to Set<Ids> allIds as what we discussed
3. Different class to handle PoolConfig Loading
PTAL, thanks

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,116 @@
...
+          // Note that we already verify pool configs' resource_config_name matches inside
+          // ResourceConfig name when loading from file.
+          throw new RuntimeException(
+              String.format(
+                  "Updating ResourceConfig on existing pool(id= %s) is not allowed, "
+                      + "please create a new pool config instead",
+                  id));
+        } else if (dbPool.size() != (configPool.poolConfig().getSize())) {
+          // Exists in both places but need to update size.
+          poolsToUpdateSize.put(dbPool.id(), configPool.poolConfig().getSize());
```
wchamber, line 75: It's an odd assymetry that the other two cases add pools to a list ,bt this one does validation and adds it to a list. Instead of having the third list be `poolsToUpdatesSize`, what if we had `duplicatePools` with a function to `reconcileDuplicates` that did the throw or update size if necesary?

src/main/java/bio/terra/rbs/service/pool/PoolWithResourceConfig.java
```diff
@@ -0,0 +1,17 @@
...
+/** Represents a {@link PoolConfig} with corresponding {@link ResourceConfig}. */
+@AutoValue
+abstract class PoolWithResourceConfig {
+  abstract PoolConfig poolConfig();
+
+  abstract ResourceConfig resourceConfig();
+
+  static PoolWithResourceConfig create(PoolConfig poolConfig, ResourceConfig resourceConfig) {
+    return new AutoValue_PoolWithResourceConfig(poolConfig, resourceConfig);
+  }
```
wchamber, line 16: Consider expanding the PoolConfig into this class, i.e. have the PoolId, size, and ResourceConfig. Right now the resourceConfigName in PoolConfig is duplicate information with wwhat's in ResourceConfig. If we're already making a class, let's make it a simpler class that has exactly what we need.

src/main/resources/api/service_openapi.yaml
```diff
@@ -94,13 +94,13 @@ components:
...
-        - poolName
+        - poolId
         - size
         - resourceConfigName
       properties:
-        poolName:
+        poolId:
           description: |
-            The name of the pool.
+            The unique id of the pool.
```
wchamber, line 103: nit: `A unique id for the pool. No two pools can ever have the same id. Any string is allowed.`

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+  private static final String RESOURCE_CONFIG_NAME = "aou_ws_resource_v1";
+
+  private static ResourceConfig newResourceConfig(GcpProjectConfig gcpProjectConfig) {
+    return new ResourceConfig().configName(RESOURCE_CONFIG_NAME).gcpProjectConfig(gcpProjectConfig);
+  }
+
+  @Test
+  public void initialize_createPool() throws Exception {
+    String poolId = "poolId";
```
wchamber, line 33: Make this a `PoolId` instead of a string. That's how it's always used. Dito other test.

src/test/java/bio/terra/rbs/service/pool/PoolConfigLoaderTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+import bio.terra.rbs.common.BaseUnitTest;
+import bio.terra.rbs.generated.model.PoolConfig;
+import bio.terra.rbs.generated.model.Pools;
+import bio.terra.rbs.generated.model.ResourceConfig;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import org.junit.jupiter.api.Test;
+
+public class PoolConfigLoaderTest extends BaseUnitTest {
```
wchamber, line 14: also have a success case test?

wchamber (changes requested):

src/main/java/bio/terra/rbs/service/pool/PoolService.java
wchamber: Hmm, I guess the ordering is important. We want to finish all validation before doing any database updating. Maybe just add a comment to that effect and leave it. And add a test that nothing is changed for each validation failure case.

wchamber: Or wrap it all in a single DB transaction to handle this for us and be less error-prone.

yonghaoy: (1) The code structure like this is because we are using one for loop to calculate the all conditions. Some condition is allowed and needs DB operation, some conditions is not allowed and we should error it out. 
Like the last else {} can be 
```
else if(poolId=configId && poolResourceConfig != parsedConfig) {
} else if(poolId=configId && poolSize != parsedSize) {}
```
From this point, I think it make sense. 
 It is just for now, only have validation for duplicated Pools.
(2) Added some comments. 
(3) Sorry, why we want all in one DB transaction? After validation, would we want insert fail also rollback update operation? 
(4) I already have test for assertThrows(RuntimeException) Add one line below to verify Pool stays the same.

src/main/java/bio/terra/rbs/service/pool/PoolWithResourceConfig.java
yonghaoy: Having this class is just an optimization of using Pair. I don't think we gona need this class anywhere else after parse it.  And I hope there is not too many similar objects because that is confusing.
Expanding this object and add more customized logic feels overkill and add unnecessary complexity. 
WDTY?

src/test/java/bio/terra/rbs/service/pool/PoolConfigLoaderTest.java
yonghaoy: I feel we already have PoolSchemaTest in which test success logic already. 
Do we want to add another success case which start from middle of that test logic?

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
yonghaoy: ditto the question of having `PoolId`: I don't feel we get too much value of having this object after we have `Pool`

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,116 @@
...
+      List<PoolWithResourceConfig> parsedPoolConfigs;
+      parsedPoolConfigs = loadPoolConfig(poolConfiguration.getConfigPath());
+      initializeFromConfig(parsedPoolConfigs);
+    }
+  }
+
+  @VisibleForTesting
+  void initializeFromConfig(List<PoolWithResourceConfig> parsedPoolConfigs) {
+    Map<String, Pool> activatePoolMap =
+        Maps.uniqueIndex(rbsDao.retrievePools(PoolStatus.ACTIVE), pool -> pool.id().toString());
```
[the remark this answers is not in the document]
yonghaoy: Done

src/main/java/bio/terra/rbs/service/pool/PoolConfigLoader.java
```diff
@@ -0,0 +1,99 @@
...
+    ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory()).findAndRegisterModules();
+    Map<String, ResourceConfig> resourceConfigNameMap = new HashMap<>();
+
+    PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
+    try {
+      Resource[] resources =
+          resolver.getResources(folderName + "/" + RESOURCE_CONFIG_SUB_DIR_NAME + "/*.yml");
+      for (Resource resource : resources) {
+        ResourceConfig resourceConfig =
+            objectMapper.readValue(resource.getInputStream(), ResourceConfig.class);
```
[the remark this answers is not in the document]
wchamber: That's not always going to be sufficient, (project prefix only has specific characters, valid region), but we can wait until we get there.

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,116 @@
...
+    Map<String, PoolWithResourceConfig> parsedPoolConfigMap =
+        Maps.uniqueIndex(parsedPoolConfigs, config -> config.poolConfig().getPoolId());
+
+    Set<String> allPoolIds = Sets.union(activatePoolMap.keySet(), parsedPoolConfigMap.keySet());
```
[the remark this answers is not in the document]
wchamber: I disagree that `PoolId` is a db layer thing. I'd like `PoolId` to be an RBS business logic thing. When a request comes in asking for a resource from a pool name, we turn that to a `PoolId`. Yes, our DB layer returns things in terms of our business logic terminology: that's convenient. The DB layer translates that business logic layer to its internal representation - text column in SQL row.
We're going to want to uniquely identify pools across RBS. Let's not stringly type that idea, let's give it a concrete type `PoolId`.

src/main/java/bio/terra/rbs/service/pool/PoolService.java
wchamber: 3) Maybe? Otherwise we get into a really weird halfway state where the DB has some but not all pools updated. Maybe I only want to increase the size of a pool because I'm removing an old one.
All or nothing is certainly easier to reason about. WDYT?

src/main/java/bio/terra/rbs/service/pool/PoolWithResourceConfig.java
wchamber: I don't think there's enough custom logic or complexity for it to really be overkill. Up to you ultimately, but I really like my types to constrain me to good behavior. Right now we can have disagreeing resouceConfigName values. The less possible you make it to represent confusing states. the only "logic" is setting some fields in this create method. We don't need to change it's signature.

src/test/java/bio/terra/rbs/service/pool/PoolConfigLoaderTest.java
wchamber: The PoolSchemaTest is for verifying that the committed configs parse. That's an end2end tet kind of independent of the impl. For PoolConfigLoaderTest, we should have a unit test that verifies a success. Let' shave good unit test coverage in addition to end2end tests.

src/main/java/bio/terra/rbs/service/pool/PoolWithResourceConfig.java
yonghaoy: The 'overkill' here I mean is having a object similar to Pool but different from PoolConfig we have in openapi yaml.
The janitor's many objects looks too complicated than it should be I feel. The use of the class is just to hold PoolConfig with ResourceConfig without further processing. Unwraping `PoolConfig` into this class before wrapped into `Pool` soon to me feels like an intermediate steps we can avoid. So I'd like to keep this class simple.

src/main/java/bio/terra/rbs/service/pool/PoolConfigLoader.java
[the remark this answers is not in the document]
yonghaoy: Yeah, we can decide until we are there. 
But a side note, do we want to have config validation logic to verify all configs? And how detailed we want to reach? For example,  like billingAccountName valid, enabledApis are actual valid API names, a lot more VPC network stuff like ip address, network thing? And if we decide we want a full detailed validation, the code logic requires a lot more work than I thought before. For each field on each type, that would be big chunk of code. 
Also, I don't think we want to verify project prefix because that is upto the client, RBS will just use whatever the prefix it gives to RBS.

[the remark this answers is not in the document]
wchamber: Always better to check configs earlier than later. I sure want to know if I typo'd my enabled API before I merge the config, otherwise I have to make a new config version name.
It's always a tradeoff between how hard it is to validate and the pain of getting it wrong. So yes, it's a chunk of code, but it could be a useful chunk of code depending on the complexity of each field.
Project ids can't contain unicode for example.

src/main/java/bio/terra/rbs/service/pool/PoolConfigLoader.java
```diff
@@ -0,0 +1,99 @@
...
+    try {
+      return objectMapper.readValue(
+          classLoader.getResourceAsStream(folderName + "/" + POOL_SCHEMA_NAME), Pools.class);
+    } catch (IOException e) {
+      throw new BadPoolConfigException(
+          String.format("Failed to parse pool schema for folder %s", folderName), e);
+    }
+  }
+
+  /** Deserialize {@link ResourceConfig} from config folder. */
```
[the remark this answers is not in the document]
wchamber: doesn't look done?

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -0,0 +1,132 @@
...
+      } else {
+        Pool dbPool = allDbPoolsMap.get(id);
+        PoolWithResourceConfig configPool = parsedPoolConfigMap.get(id);
+        if (dbPool.status().equals(PoolStatus.INACTIVE)) {
+          throw new RuntimeException(
+              String.format(
+                  "An existing inactive pool with duplicate id(id= %s) found, "
+                      + "please consider change the pool id.",
+                  id));
+        }
```
wchamber, line 79: is there a test for this? A TODO in the test file is fine until we actually delete pools.

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    rbsDao.createPools(ImmutableList.of(pool));
+
+    PoolWithResourceConfig parsedPoolConfig =
+        PoolWithResourceConfig.create(
+            new PoolConfig()
+                .poolId(poolId.toString())
+                .size(10)
+                .resourceConfigName(RESOURCE_CONFIG_NAME),
+            newResourceConfig(new GcpProjectConfig().projectIDPrefix("aou-rw-test111")));
```
wchamber, line 82: Add a comment to draw more attention to this line. It's pretty hard to pick out this is what's important.
Or make variables for each ResourceConfig to make it clear there's two and they're distinct.

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+  TransactionStatus transactionStatus;
+
+  private static final String RESOURCE_CONFIG_NAME = "aou_ws_resource_v1";
+
+  private static ResourceConfig newResourceConfig(GcpProjectConfig gcpProjectConfig) {
+    return new ResourceConfig().configName(RESOURCE_CONFIG_NAME).gcpProjectConfig(gcpProjectConfig);
+  }
+
+  @Test
+  public void initialize_createPool() throws Exception {
```
wchamber, line 34: can we also test the case where the pool already exists in the database?

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+    assertEquals(1, pools.size());
+    Pool createdPool = pools.get(0);
+    assertEquals(poolId, createdPool.id());
+    assertEquals(ResourceType.GOOGLE_PROJECT, createdPool.resourceType());
+    assertEquals(PoolStatus.ACTIVE, createdPool.status());
+    assertEquals(resourceConfig, createdPool.resourceConfig());
+  }
+
+  @Test
+  public void initialize_resourceConfigUpdate() throws Exception {
```
wchamber, line 61: nit: I had a hard time figuring out what this test was verifying from it's name. Consider including the expected outcome as well as the condition.
https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/best-practices.md?cl=head&polyglot=java#naming

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+            .creation(Instant.now())
+            .id(poolId)
+            .resourceType(ResourceType.GOOGLE_PROJECT)
+            .size(1)
+            .resourceConfig(
+                newResourceConfig(new GcpProjectConfig().projectIDPrefix("aou-rw-test")))
+            .status(PoolStatus.ACTIVE)
+            .build();
+
+    rbsDao.createPools(ImmutableList.of(pool));
```
wchamber, line 74: WDYT about calling `poolService.initializeFromConfig` twice instead of creating pools directly with the dao? Maybe instead of `initializeFromConfig` it should be named `updateFromConfig`

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+                .resourceConfigName(RESOURCE_CONFIG_NAME),
+            newResourceConfig(new GcpProjectConfig().projectIDPrefix("aou-rw-test111")));
+
+    assertThrows(
+        RuntimeException.class,
+        () ->
+            poolService.initializeFromConfig(
+                ImmutableList.of(parsedPoolConfig), transactionStatus));
+
+    assertEquals(rbsDao.retrievePools().get(0), pool);
```
wchamber, line 90: nit: use `assertThat(rbsDao.reqtrievePools(), Matchers.containsInAnyOrder(pool))` for better verification and error messages.

src/test/resources/application-test.yml
```diff
@@ -5,3 +5,5 @@ rbs:
     update-db-on-start: true
     uri: jdbc:postgresql://127.0.0.1:5432/testdb
     username: dbuser
+  pool:
+    update-pool-on-start: false
```
wchamber, line 9: add a comment about why this is desirable

wchamber (approved):

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
yonghaoy: Done

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
yonghaoy: Done

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
yonghaoy: Done

src/main/java/bio/terra/rbs/service/pool/PoolService.java
yonghaoy: Right, it would be better to add test once we have updatePool in RbsDao later. Added a TODO in test

src/main/java/bio/terra/rbs/service/pool/PoolConfigLoader.java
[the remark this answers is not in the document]
yonghaoy: Make sense, I may start a doc to list all validation rules for each of those field. And how we will validate them. 
And probably bring this up on tomorrow's standup.

PR #6, opened 2020-10-02
[PF-129] Setup RBS dao to support create pool and get pool

yonghaoy:
That would be used in my following PR


src/main/java/bio/terra/rbs/db/PoolStatus.java
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * The state of the {@link Pool}.
+ *
+ * <p>This is persisted as a string in the database, so the names of the enum values should not be
+ * changed.
+ */
+public enum PoolStatus {
+  ACTIVE,
+  INACTIVE,
```
wchamber, line 11: comment what these states mean, and what the Pool is & isn't allowed to do/look like when it's in these states.

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -0,0 +1,109 @@
...
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Isolation;
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+@Component
+public class RbsDao {
```
wchamber, line 24: class comment

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -0,0 +1,109 @@
...
+      NamedParameterJdbcTemplate jdbcTemplate,
+      @Qualifier(OBJECT_MAPPER) ObjectMapper objectMapper) {
+    this.jdbcTemplate = jdbcTemplate;
+    this.objectMapper = objectMapper;
+  }
+
+  /**
+   * Creates the pool record and adding labels.
+   *
+   * <p>Note that we assume int input {@code cloudResourceUid} is valid.
```
wchamber, line 39: copy paste error?

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -0,0 +1,109 @@
...
+              .name(rs.getString("name"))
+              .resourceConfig(deserialize(rs.getString("resource_config")))
+              .resourceType(ResourceType.valueOf(rs.getString("resource_type")))
+              .status(PoolStatus.valueOf(rs.getString("status")))
+              .size(rs.getInt("size"))
+              .creation(rs.getObject("creation", OffsetDateTime.class).toInstant())
+              .build();
+
+  /** Serializes {@link ResourceConfig} into json format string. */
+  private static String serialize(ResourceConfig resourceConfig) {
```
wchamber, line 91: Why does Pool get an intermediate Java class but ResourceConfig doesn't?

src/main/java/bio/terra/rbs/db/ResourceType.java
```diff
@@ -0,0 +1,14 @@
...
+ * Enums to represent cloud resource type we supported in DB schema.
+ *
+ * <p>These enums are recorded as strings in the database and should therefore not be removed or
+ * modified.
+ *
+ * <p>It is ok to add new values. Also update the BackwardsCompatibilityTest.
+ */
+public enum ResourceType {
+  GOOGLE_PROJECT,
+  GKE_CLUSTER,
```
wchamber, line 13: Is it very helpful to have a second resource type for testing? Maybe make a TEST_VALUE or DUMMY enum since we're a very long ways from supporting this resource type. At least comment why we have a value we're not going to support anytime soon.

src/test/java/bio/terra/rbs/db/BackwardsCompatibilityTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit tests to verify that changes to Java classes do not break backwards compatibility with
+ * values that might already exist in databases.
+ *
+ * <p>If your change causes one of these tests to fail, consider whether it is backwards compatible
+ * instead of modifying the test to pass.
+ */
+public class BackwardsCompatibilityTest extends BaseUnitTest {
```
wchamber, line 15: let's also (eventually) have some backwards compatibility tests for ResourceConfig. Could be in another file, or a TODO + ticket to do later since we'll be iterating on ResourceConfig for a while where we don't care about backwards compatibility.

wchamber (changes requested):
Thanks for breaking this out, few nits

src/main/java/bio/terra/rbs/db/PoolStatus.java
yonghaoy: done

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: done

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: That is similar to what we discussed before about Janitor. Just like CloudResourceUid, ResourceConfig is also an external things stored in RBS db as json string. api/db layers will just use the same object after deserialize from db

src/main/java/bio/terra/rbs/db/ResourceType.java
yonghaoy: The GKE_CLUSTER is not helpful at all. Removed it.

src/test/java/bio/terra/rbs/db/BackwardsCompatibilityTest.java
yonghaoy: what you mean by backwards compatibility for ResourceConfig? That is external object and RBS will just serialize/deserialize it. And our test:https://github.com/DataBiosphere/terra-rbs/blob/master/src/test/java/bio/terra/rbs/resource/PoolSchemaTest.java#L58
will verify rbs's resource config yaml is compatible with resource config schema. 
But my following up PR will add a ResourceConfigVisistor with a test to verify we have the ResourceType supported listed in config

wchamber (approved):
You missed a nit about copy/paste comment in RbsDao. Otherwise lgtm.

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: done

PR #8, opened 2020-10-06
[PF-112] initial stairway setup with status endpoint

yonghaoy:
Will delete ping test in later PR


src/main/resources/application.yml
```diff
@@ -7,3 +7,13 @@ rbs:
     username: ${DATABASE_USER}
   server:
     port: 8080
+  stairway:
+    db:
+      password: ${STAIRWAY_DATABASE_USER_PASSWORD}
+      uri: jdbc:postgresql://127.0.0.1:5432/${STAIRWAY_DATABASE_NAME}
+      username: ${STAIRWAY_DATABASE_USER}
```
wchamber, line 14: instead of hardcoding environment variables in the applicatoin.yml, let's use Spring's automatic env variable to application variable mapping. Ditto for above. I think that's inline with Greg's config doc. (We should also update Janitor eventually, but we can wait until we have a new model for all config there.)

So, set nothing here, and set RBS_STAIRWAY_DB_PASSWORD in `run_local.sh` and elsewhere. Including helm chart changes.

src/main/java/bio/terra/rbs/service/stairway/StairwayComponent.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public void initialize() {
+    logger.warn("stairway username {}", stairwayJdbcConfiguration.getUsername());
+    try {
+      // TODO(CA-941): Determine if Stairway and rbs database migrations need to be coordinated.
+      stairway.initialize(
+          stairwayJdbcConfiguration.getDataSource(),
+          stairwayConfiguration.isForceCleanStart(),
+          stairwayConfiguration.isMigrateUpgrade());
+      // TODO(CA-941): Get obsolete Stairway instances from k8s for multi-instance stairway.
```
wchamber, line 68: new ticket for multi-instance stairway for RBS?
In Janitor, I think the ticket should now be PF-35 fwiw

wchamber (changes requested):

src/main/resources/application.yml
wchamber: Or we could wait, but I have a mild preference for starting RBS closer to where we want to be.

yonghaoy: Sorry, I just checked Greg's config doc, didn't find auto env variable and how it works. Can you give me the pointer. If that is not the broad agreed decision, I'd prefer we can wait until we discuss the upcoming change with other teams. 
I don't think migrate for RBS later gives us much more work compare with doing it now.

wchamber: the application path turns into upper case with underscores.
e.g.
```
# application.yaml
foo:
  bar:
   baz: 2
``` 
can be set with `FOO_BAR_BAZ=2`.

Janitor example, overriding enabling stackdriver.
https://github.com/DataBiosphere/terra-resource-janitor/blob/68edb71d13ac04b4e1cb76e922206023481089f2/local-dev/run_local.sh#L6
https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/main/resources/application.yml#L22

which is what Greg is talking about in this comment chain: https://docs.google.com/document/d/1VfgL8D6PpJg7irbL_gKnmyZ5_GH5OYGfK8GPN45odUs/edit?disco=AAAAG95t27k

wchamber (approved):

src/main/resources/application.yml
yonghaoy: Ah, I missed that part. Thanks. Updated

PR #9, opened 2020-10-07
[PF-126][PF129] Support update pool size and deactivate pool

yonghaoy:


src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -123,10 +124,10 @@ private void createPools(List<PoolWithResourceConfig> poolsToCreate) {
   }
 
   private void deletePools(List<Pool> poolsToDelete) {
```
wchamber, line 126: maybe let's rename this to `deactivatePools` as well since we're not actually deleting them.

src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -123,10 +124,10 @@ private void createPools(List<PoolWithResourceConfig> poolsToCreate) {
...
-    // TODO: Delete pool
+    rbsDao.deactivatePools(poolsToDelete.stream().map(Pool::id).collect(Collectors.toList()));
   }
 
   private void updatePoolSize(Map<PoolId, Integer> poolsToUpdateSize) {
-    // TODO: Update pool size
+    rbsDao.updatePoolsSize(poolsToUpdateSize);
```
wchamber, line 131: optional nit: maybe `poolSizes` instead of `poolsToUpdateSize`?

src/test/java/bio/terra/rbs/db/RbsDaoTest.java
```diff
@@ -30,35 +32,59 @@ public void setup() {
...
+  }
+
+  @Test
+  public void updatePoolSize() {
+    PoolId poolId = PoolId.create("poolId");
+    Pool pool = newPool(poolId);
+
+    rbsDao.createPools(ImmutableList.of(pool));
+    Pool retrievedPool = rbsDao.retrievePools().get(0);
+    Pool newPool = pool.toBuilder().size(retrievedPool.size() + 10).build();
```
wchamber, line 85: nit: `resizedPool` instead of `newPool`

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
```diff
@@ -111,11 +106,86 @@ public void updateFromConfig_updateResourceConfigOnExistingPool_throwsException(
...
+
+    assertThrows(
+        RuntimeException.class,
+        () -> poolService.updateFromConfig(ImmutableList.of(parsedPoolConfig), transactionStatus));
   }
 
   @Test
-  public void updateFromConfig_duplicatePoolId() throws Exception {
-    // TODO: Implement this once Dao support update status.
+  public void updateFromConfig_deletePool_updatePoolStatusSuccess() throws Exception {
```
wchamber, line 137: We're getting a little inconsistent in our terminology about whether pools are "deleted" or "inactivate" or "deactivated." Let's settle on one term, and make sure we use it in variable names, method names, and test method names.
I think "deactivated/deactivate" might be best. Mind taking a pass through and looking for "inactive" and "delete?"

wchamber (approved):
lgtm, just some comments on naming

src/test/java/bio/terra/rbs/service/pool/PoolServiceTest.java
yonghaoy: Ah, right, that is confusing. 
Use deactivated/deactivate everythere now

PR #11, opened 2020-10-07
[PF-130] RbsDao supports retrieve pool with corresponding number of resources.

yonghaoy:


src/main/java/bio/terra/rbs/db/PoolAndResourceCount.java
```diff
@@ -0,0 +1,15 @@
+package bio.terra.rbs.db;
+
+import com.google.auto.value.AutoValue;
+
+/** Represents a {@link Pool} with number of activate resources in the pool. */
+@AutoValue
+abstract class PoolAndResourceCount {
+  abstract Pool pool();
+
+  abstract int resourceCount();
```
wchamber, line 10: maybe `int activeResources()`? (are we calling them ACTIVE or READY? class comment says "active." consistent terminology is nice again.)

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -118,6 +131,19 @@ public void updatePoolsSize(Map<PoolId, Integer> poolsToUpdateSize) {
...
+      (rs, rowNum) ->
+          PoolAndResourceCount.create(
+              Pool.builder()
+                  .id(PoolId.create(rs.getString("id")))
+                  .resourceConfig(deserialize(rs.getString("resource_config")))
+                  .resourceType(ResourceType.valueOf(rs.getString("resource_type")))
+                  .status(PoolStatus.valueOf(rs.getString("status")))
+                  .size(rs.getInt("size"))
+                  .creation(rs.getObject("creation", OffsetDateTime.class).toInstant())
+                  .build(),
```
wchamber, line 144: you could use a POOL_ROW_MAPPER here.

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -72,6 +72,19 @@ public void createPools(List<Pool> pools) {
     return jdbcTemplate.query(sql, POOL_ROW_MAPPER);
   }
 
+  /** Retrieves all pools and resource count in each pool. */
+  @Transactional(propagation = Propagation.SUPPORTS)
+  public List<PoolAndResourceCount> retrievePoolAndResourceCount() {
+    // TODO: Add filter
+    String sql =
+        "select p.id, p.resource_config, p.resource_type, p.creation, p.size, p.status, "
+            + "(select count (*) from resource where pool_id = p.id) as resource_count "
```
wchamber, line 81: this selects all resources counts, but `PoolAndResourceCount` talks about active resources. What do we want?
One thing we can do is have `PoolAndResourceCount` use a `Multiset<ResourceState>` to get counts of all the different resource states. Not sure what's going to be helfpul in the scheduler.

src/main/java/bio/terra/rbs/service/resource/FlightScheduler.java
```diff
@@ -0,0 +1,4 @@
+package bio.terra.rbs.service.resource;
+
+/** Schedulling services periodically scans */
+public class FlightScheduler {}
```
wchamber, line 4: optional nit: this file doesn't do anything yet.

wchamber (changes requested):

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: Ah, right, thanks

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: I think we only cares about the READY resource state. So add a where clause should be enough?

src/main/java/bio/terra/rbs/db/PoolAndResourceCount.java
yonghaoy: Changed to activeResourcesCount

src/main/java/bio/terra/rbs/db/PoolStatus.java
```diff
@@ -10,7 +10,8 @@
   /** Active pool, able to handout resources. */
   ACTIVE,
   /**
-   * Deactivated pool, all resources are deactivated or being deactivated, not able to handout resources.
+   * Deactivated pool, all resources are deactivated or being deactivated, not able to handout
+   * resources.
```
yonghaoy, line 14: Modified by spotlessApply

src/main/java/bio/terra/rbs/db/RbsDao.java
wchamber: I'm not sure that's true for the scheduler. We don't want to overschedule resources if there are some resource creations in flight. i.e. we have 10 READY, 2 CREATING, for a pool of size 20. How many flights should we make?
We could make have 2 int fields, resourceReadyCount, resourceCreatingCount, but a `SetMultimap` might scale better.

src/main/java/bio/terra/rbs/db/ResourceState.java
```diff
@@ -0,0 +1,16 @@
...
+ * <p>This is persisted as a string in the database, so the names of the enum values should not be
+ * changed.
+ */
+public enum ResourceState {
+  /** Resource is creating. */
+  CREATING,
+  /** Resource is ready to handout. */
+  READY,
+  /** Resource is handed out. */
+  HANDEDOUT,
```
wchamber, line 15: nit: underscored two words, HANDED_OUT?

wchamber (changes requested):
I do think we need to know about the count of CREATING resources for scheduling. Isn't that what we're going to use PoolAndResourceCount for?

src/main/java/bio/terra/rbs/db/ResourceState.java
wchamber: also, I don't mind 'handed out' but it is a little awkward to use. any other naming ideas before we commit to the database forever? We could do "USED"? idk, no strong preference.

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: I think your are right. Updated the sql query and code little bit. PTAL Thanks

src/main/java/bio/terra/rbs/db/ResourceState.java
yonghaoy: Okay, changed to used.

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -72,6 +74,30 @@ public void createPools(List<Pool> pools) {
     return jdbcTemplate.query(sql, POOL_ROW_MAPPER);
   }
 
+  /** Retrieves all pools and READY resource count in each pool. */
+  @Transactional(propagation = Propagation.SUPPORTS)
+  public List<PoolAndResourceStates> retrievePoolAndResourceStatesCount() {
+    // TODO: Add filter
+    String sql =
+        "select count (*) FILTER (WHERE r.state='READY') as ready_count, "
```
yonghaoy, line 82: Not sure if this is the best approach, PTAL, thanks

wchamber: I think it's simpler if we get the counts for all of the resource states, even if we're not examining the handed out state yet. I think a GROUP BY r.state is also more idiomatic than using select subexpressions for each r.state value.
```
select count(*) as resource_count, r.state,
  p.id, p.resource_config, p.resource_type, p.creation, p.size, p.status
from pool p
join resource r on r.pool_id = p.id
group by p.id, r.state;
```
I think that works, but test to verify.
We do need to change the row mapper to be a ResultSetExtractor to group the resource state rows into a single object.
In Janitor, we do this to group the many rows of a TrackedResource with different label key/values into a single object per TrackedResource with all label values. So, not exactly what we're trying to do here, but similar. Let me know if you want to talk more.
https://github.com/DataBiosphere/terra-resource-janitor/blob/e4da842229aacd1e99bfcb1fffb9ac1eadb611dc/src/main/java/bio/terra/janitor/db/JanitorDao.java#L360-L381

src/main/java/bio/terra/rbs/db/ResourceState.java
```diff
@@ -0,0 +1,16 @@
...
+ * <p>This is persisted as a string in the database, so the names of the enum values should not be
+ * changed.
+ */
+public enum ResourceState {
+  /** Resource is creating. */
+  CREATING,
+  /** Resource is ready to handout. */
+  READY,
+  /** Resource is handed out. */
+  USED,
```
wchamber, line 15: I really don't know if USED or HANDED_OUT is better. WDYT?

wchamber (changes requested):

src/main/java/bio/terra/rbs/db/ResourceState.java
yonghaoy: `USED` sounds wired. I vote for `HANDED_OUT`

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -72,6 +77,19 @@ public void createPools(List<Pool> pools) {
     return jdbcTemplate.query(sql, POOL_ROW_MAPPER);
   }
 
+  /** Retrieves all pools and resource count for each state. */
+  @Transactional(propagation = Propagation.SUPPORTS)
+  public List<PoolAndResourceStates> retrievePoolAndResourceStatesCount() {
```
wchamber, line 82: nit: `retrievePoolAndResourceStates()`

wchamber (approved):

PR #12, opened 2020-10-08
[PF-164] Support retrieve and create resource in RbsDao

yonghaoy:


src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -107,34 +134,117 @@ public void updatePoolsSize(Map<PoolId, Integer> poolsToUpdateSize) {
...
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public void createResource(Resource resource) {
+    String sql =
+        "INSERT INTO resource (id, pool_id, creation, state) values "
+            + "(:id, :pool_id, :creation, :state)";
+    MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue("id", resource.id().id())
+            .addValue("pool_id", resource.poolId().id())
+            .addValue("creation", resource.creation().atOffset(ZoneOffset.UTC))
```
yonghaoy, line 147: I was little bit concerned the creation time in resource creation flight. In a Flight, 1st step creates a Resource with Creation state. The final step is to update the state to READY. When the creation time will be updated. Step 1, final step or both? 
I tentative to set it in the 1st step, I don't think we need this time to accurate. And feel it is better to always let the  entity in db has creation time. Will, wdyt?

src/main/resources/api/service_openapi.yaml
```diff
@@ -90,6 +90,26 @@ components:
           type: string
       type: object
 
+    ResourceInfo:
```
yonghaoy, line 93: this is little bit too earlier to add. But I think this is what we need eventually. And it is the correct to let rbs generate `CloudResourceUid`

src/main/java/bio/terra/rbs/db/Resource.java
```diff
@@ -0,0 +1,53 @@
...
+  public abstract CloudResourceUid terraResourceUid();
+
+  @Nullable
+  public abstract RequestHandoutId requestHandoutId();
+
+  @Nullable
+  public abstract Instant creation();
+
+  @Nullable
+  public abstract Instant handoutTime();
```
wchamber, line 27: I think most of these fields are close to self-explanatory, but it might still be nice to have /** javadoc */ comments for all of them. Also very much worth explaning what it means for the nullable fields to be null or not null. Or a quick comment on the life cycle of a resource based on its fields.

src/main/java/bio/terra/rbs/db/Resource.java
```diff
@@ -0,0 +1,53 @@
...
+@AutoValue
+public abstract class Resource {
+  public abstract ResourceId id();
+
+  public abstract PoolId poolId();
+
+  public abstract ResourceState state();
+
+  @Nullable
+  public abstract CloudResourceUid terraResourceUid();
```
wchamber, line 18: If we want to have a generic TerraResourceUid that can have "Terra resources" that aren't just cloud resources, I think it should be a new class that has a CloudResourceUid field. If we want to keep it simple and only have a CloudResourceUid for now, let's keep the return type and field name the same as `cloudResourceUid()` and change that when it changes.

src/main/resources/api/service_openapi.yaml
```diff
@@ -90,6 +90,26 @@ components:
...
+      properties:
+        poolId:
+          description: |
+            The pool id the resource belongs to.
+          type: string
+        requestHandoutId:
+          description: |
+            The unique identifier used for handout resource. RBS will handout the same resource for the same id.
+          type: string
+        cloudResourceUid:
```
wchamber, line 109: can we also add a `description` for this field, or is that not possible when we have a `$ref`?

src/main/resources/api/service_openapi.yaml
wchamber: I agree, but my optional suggestion is to only add it once we use it.

src/test/java/bio/terra/rbs/db/RbsDaoTest.java
```diff
@@ -91,16 +98,46 @@ public void updatePoolSize() {
...
     Pool pool2 = newPool(PoolId.create("poolId2"));
-    rbsDao.createPools(ImmutableList.of(pool1, pool2));
+    Pool pool3 = newPool(PoolId.create("poolId3"));
+
+    // Pool1 has 1 CREATING, 2 READY, Pool2 has 1 READY, 1 USED, Pool3 is empty
+    Resource pool1CreatingResource1 = newResource(pool1.id(), ResourceState.CREATING);
+    Resource pool1ReadyResource1 = newResource(pool1.id(), ResourceState.READY);
+    Resource pool1ReadyResource2 = newResource(pool1.id(), ResourceState.READY);
+    Resource pool2UsedResource1 = newResource(pool2.id(), ResourceState.USED);
+    Resource pool2ReadyResource1 = newResource(pool2.id(), ResourceState.READY);
```
wchamber, line 110: I think we can inline these Resources in the `rbsDao.createResource`

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -133,34 +134,109 @@ public void updatePoolsSize(Map<PoolId, Integer> poolsToUpdateSize) {
...
```
wchamber, line 147: technically, resource.creation() is `@Nullable.` If RbsDao is enforcing that creation() can't be null when the Resource is added with createResource, I'm not sure that resource.creation() should be `@Nullable`.

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -133,34 +134,109 @@ public void updatePoolsSize(Map<PoolId, Integer> poolsToUpdateSize) {
...
+              .terraResourceUid(
+                  rs.getString("terra_resource_uid") == null
+                      ? null
+                      : deserializeResourceUid(rs.getString("terra_resource_uid")))
+              .state(ResourceState.valueOf(rs.getString("state")))
+              .requestHandoutId(
+                  rs.getString("request_handout_id") == null
+                      ? null
+                      : RequestHandoutId.create(rs.getString("request_handout_id")))
+              .creation(rs.getObject("creation", OffsetDateTime.class).toInstant())
```
wchamber, line 192: ditto is creation `@Nullable`?

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -133,34 +134,109 @@ public void updatePoolsSize(Map<PoolId, Integer> poolsToUpdateSize) {
...
+          .setSerializationInclusion(JsonInclude.Include.NON_NULL)
+          .writeValueAsString(resourceUid);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(
+          String.format("Failed to serialize ResourceConfig: %s", resourceUid), e);
+    }
+  }
+
+  /** Deserializes {@link CloudResourceUid} into json format string. */
+  private static CloudResourceUid deserializeResourceUid(String terraResourceUid) {
```
wchamber, line 234: ditto naming terraResourceUid vs CloudResourceUid consistency.

wchamber (changes requested):

src/main/resources/api/service_openapi.yaml
yonghaoy: We need to add some object to reference to CloudResourceUid yaml. I think that is something we need to have now. So I added this before hand.

build.gradle
```diff
@@ -23,12 +23,17 @@ sourceCompatibility = '1.8'
...
     maven {
         url 'https://broadinstitute.jfrog.io/broadinstitute/libs-release/'
     }
+    maven {
+        url 'https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot-local/'
+    }
 }
 
+// The version of CRL schema library
+def crlSchemaVersion = "0.0.2-SNAPSHOT"
```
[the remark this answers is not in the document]
yonghaoy: https://broadworkbench.atlassian.net/browse/PF-165

Also created one for Janitor client, but I don't think that would be needed recently, just something keep in mind
https://broadworkbench.atlassian.net/browse/PF-166

src/main/java/bio/terra/rbs/db/Resource.java
yonghaoy: Agree that we may want `TerraResourceUid` instead of `CloudResourceUid` directly. For now I renamed to cloudResourceUid.

src/main/resources/api/service_openapi.yaml
yonghaoy: I think it will get compiled. But I got intellij complains. "Property 'description' is not allowed". I think what open_api want is go to the $ref to look up the description.

src/test/java/bio/terra/rbs/db/RbsDaoTest.java
yonghaoy: Done.

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: Yup. `creation` is not `nullable` now. I was thinking about the question I mentioned above when shall we set the Creation. If that happens when we create the resource. It will not be `nullable`.

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: not nullable now

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: done

src/main/java/bio/terra/rbs/db/Resource.java
yonghaoy: done

build.gradle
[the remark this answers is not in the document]
wchamber: thanks!

wchamber (approved):
lgtm, just needs to be updated for changes to yyu-PF-130.

PR #13, opened 2020-10-09
[PF-130] Flight scheduler how to submit creates/deactivates flights

yonghaoy:


src/test/java/bio/terra/rbs/service/resource/FlightSchedulerTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+@AutoConfigureMockMvc
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+public class FlightSchedulerTest extends BaseUnitTest {
+  // Construct a FlightScheduler manually instead of Autowired for ease of testing.
+  private FlightScheduler flightScheduler;
+
+  @Autowired RbsDao rbsDao;
+  @Autowired StairwayComponent stairwayComponent;
+  @MockBean FlightManager flightManager;
```
wchamber, line 32: Since we don't have a FlightManager yet, the interaction testing here is ok. But long term it would be better to check that resources are created or deleted as appropriate instead of just testing the interaction between FlightScheduler and FlightManager. go/mocks#testing-interactions

src/test/resources/application-test.yml
```diff
@@ -5,6 +5,8 @@ rbs:
     update-db-on-start: true
     uri: jdbc:postgresql://127.0.0.1:5432/testdb
     username: dbuser
+  primary:
+    scheduler-enabled: false
   pool:
     update-pool-on-start: false
```
wchamber, line 11: add some comment on why these are helpful default values for testing.

wchamber (approved):

src/test/java/bio/terra/rbs/service/resource/FlightSchedulerTest.java
yonghaoy: I was thinking scheduling flight to actually invoke real FlightManager in our integration test once we have those flights. Does that SG?

wchamber: Yep!

src/main/java/bio/terra/rbs/app/configuration/PrimaryConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+  private boolean schedulerEnabled;
+
+  /** How often to query for flights to submit. */
+  private Duration flightSubmissionPeriod = Duration.ofMinutes(1);
+
+  /**
+   * How many resource creation flights for a pool to process because we don't want a pool eats all
+   * flights.
+   */
+  private int resourceCreationPerPoolLimit = 100;
```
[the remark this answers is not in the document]
yonghaoy: I think we know some of the problems is the cloud qps is 1/s. And scheduler runs every 1 minutes. 
So I kind of give it a larger number(>60, but not double).
And we also want to have some control on the limit of resource deletion number, because I feel creating is more important. 
So I come up with the initial number here. It is very early and subject to change. But I think have it here wouldn't hurt. I am sure no limit and submit 100k flight to create project would fail 90% if that runs within 1 minutes even with 5 times retry.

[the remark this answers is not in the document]
wchamber: SGTM

PR #14, opened 2020-10-09
[PF-162] Add two more enums in resource state for delete resource

yonghaoy:
Those are needed in delete resource flight


wchamber (approved):
Generally it's easier to evaluate these kind of changes if they're associated with the code that starts to use them. Consider waiting until they're used before adding them next time.

PR #15, opened 2020-10-09
[PF-148] Flight Manager to submit creation/deletion resource flight

yonghaoy:
Hi Will, I don't think there is anything interesting need to test in FlightManager now. And my plan is to add integration test which has cloud API calls.  

If you think I should have some unit test, maybe an `dummyFlight` which has `initialSetp`(), `finalStep`(), and verify Flight status? Not sure if that is necessary once we have the integration test I mentioned above. Let me know if we should add unit test. 
Thanks

### DIFF
--- a/src/main/java/bio/terra/rbs/service/resource/FlightFactory.java
+++ b/src/main/java/bio/terra/rbs/service/resource/FlightFactory.java
@@ -1,0 +1,13 @@
+package bio.terra.rbs.service.resource;
+
+import bio.terra.rbs.db.ResourceType;
+import bio.terra.stairway.Flight;
+import org.springframework.stereotype.Component;
+
+/** An interface getting {@link Flight} from {@link ResourceType}. */
+@Component
+public interface FlightFactory {
+  Class<? extends Flight> getCreationFlightClass(ResourceType type);
+
+  Class<? extends Flight> getDeletionFlightClass(ResourceType type);
+}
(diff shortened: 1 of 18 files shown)


src/main/java/bio/terra/rbs/service/resource/FlightScheduler.java
```diff
@@ -105,9 +105,17 @@ private void scheduleCreationFlights(Pool pool, int number) {
         pool.id(),
         flightToSchedule);
 
+    int successSubmitNum = 0;
     while (flightToSchedule-- > 0) {
-      flightManager.submitCreationFlight(pool);
+      boolean submissionSuccessful = flightManager.submitCreationFlight(pool);
+      if (submissionSuccessful) {
```
wchamber, line 111: optional nit: inline `submissionSuccessful`

src/main/java/bio/terra/rbs/service/resource/FlightFactory.java
```diff
@@ -0,0 +1,13 @@
+package bio.terra.rbs.service.resource;
+
+import bio.terra.rbs.db.ResourceType;
```
wchamber, line 3: not this PR, but lets' consider moving some of the enums/autovalue classes out of db and into common since we do want them to be used across RBS.

wchamber (approved):

src/main/java/bio/terra/rbs/service/resource/FlightFactory.java
yonghaoy: ah, okay, I will do it after my first project creation PR.

PR #16, opened 2020-10-13
[PF-127] Create GCP projects

yonghaoy:


src/main/java/bio/terra/rbs/service/resource/FlightManager.java
```diff
@@ -21,17 +22,20 @@
 
   private final FlightFactory flightFactory;
   private final Stairway stairway;
+  private CloudResourceManagerCow rmCow;
```
wchamber, line 25: final?

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -171,6 +171,29 @@ public void createResource(Resource resource) {
...
+    jdbcTemplate.update(sql, params);
+  }
+
+  /** Delete the resource match the {@link ResourceId}. */
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public void deleteResource(ResourceId id) {
+    String sql = "DELETE FROM resource WHERE id = :id";
+    MapSqlParameterSource params = new MapSqlParameterSource().addValue("id", id.id());
+
+    jdbcTemplate.update(sql, params);
```
wchamber, line 194: maybe return a bool if anything was deleted?

src/main/java/bio/terra/rbs/service/resource/flight/CreateGoogleProjectStep.java
```diff
@@ -0,0 +1,61 @@
...
+  @Override
+  public StepResult doStep(FlightContext flightContext) {
+    String projectId = flightContext.getWorkingMap().get(GOOGLE_PROJECT_ID, String.class);
+    Project project =
+        new Project()
+            .setProjectId(projectId)
+            .setParent(
+                new ResourceId().setType("folder").setId(gcpProjectConfig.getParentFolderId()));
+    try {
+      Operation operation = rmCow.projects().create(project).execute();
```
wchamber, line 38: I think you could run into a problem if this step gets when the project was already created. Maybe check that the project doesn't exist first?

src/main/java/bio/terra/rbs/service/resource/flight/CreateGoogleProjectStep.java
```diff
@@ -0,0 +1,61 @@
...
+      OperationCow<Operation> operationCow = rmCow.operations().operationCow(operation);
+      OperationUtils.pollUntilComplete(operationCow, Duration.ofSeconds(5), Duration.ofSeconds(30));
+    } catch (IOException | InterruptedException e) {
+      logger.error("Error when creating GCP project", e);
```
wchamber, line 42: I'm not sure this warrants `.error`: we expect this to happen some amount with GCP. Could logger.info, or just let our CRL logging take care of it.

src/main/java/bio/terra/rbs/service/resource/flight/CreateGoogleProjectStep.java
```diff
@@ -0,0 +1,61 @@
...
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    try {
+      rmCow
+          .projects()
+          .delete(flightContext.getWorkingMap().get(GOOGLE_PROJECT_ID, String.class))
+          .execute();
```
wchamber, line 54: you should account for the project having failed to have been created.

src/main/java/bio/terra/rbs/service/resource/flight/FinalCreateResourceStep.java
```diff
@@ -0,0 +1,37 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+
+/**
+ * The initial step to create resource. It creates a entity in resource db table with CREATING
+ * state.
+ */
+public class FinalCreateResourceStep implements Step {
```
wchamber, line 15: think the name and comment are out of date. Maybe `ReadyResourceStep` or `FinishResourceCreationStep` or `MarkResourceReadyStep`?

src/main/java/bio/terra/rbs/service/resource/flight/FinalCreateResourceStep.java
```diff
@@ -0,0 +1,37 @@
...
+
+    rbsDao.updateResourceAfterCreation(
+        workingMap.get(FlightMapKeys.RESOURCE_ID, ResourceId.class),
+        workingMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class));
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    // Nothing need to do.
```
wchamber, line 34: worth explaining why. I think this is your doc that we shared with dan that there's nothing to do undo here because we either updated the RbsDao and are finished, or we didn't and we don't need to undo something. I think this is a little tricky and worth a longer comment.

src/main/java/bio/terra/rbs/service/resource/flight/GenerateGoogleProjectIdStep.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.UUID;
+
+/** Generates Project Id and put it in working map. */
+public class GenerateGoogleProjectIdStep implements Step {
+  public GenerateGoogleProjectIdStep() {}
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) {
+    FlightMap workingMap = flightContext.getWorkingMap();
+    // TODO(yonghao): Use Terra Project Id generator.
```
wchamber, line 18: nit: make a ticket?

src/main/java/bio/terra/rbs/service/resource/flight/GenerateGoogleProjectIdStep.java
```diff
@@ -0,0 +1,36 @@
...
+import static bio.terra.rbs.service.resource.FlightMapKeys.CLOUD_RESOURCE_UID;
+import static bio.terra.rbs.service.resource.FlightMapKeys.GOOGLE_PROJECT_ID;
+
+import bio.terra.rbs.generated.model.CloudResourceUid;
+import bio.terra.rbs.generated.model.GoogleProjectUid;
+import bio.terra.stairway.*;
+import java.util.UUID;
+
+/** Generates Project Id and put it in working map. */
+public class GenerateGoogleProjectIdStep implements Step {
```
wchamber, line 12: optoinal nit: I don't know that we need the `Google`. We don't have any other project ids, and when we start making more resources, their resource specific flights can end up in different packages.

src/main/java/bio/terra/rbs/service/resource/flight/GenerateResourceIdStep.java
```diff
@@ -0,0 +1,27 @@
+package bio.terra.rbs.service.resource.flight;
+
+import bio.terra.rbs.db.*;
+import bio.terra.rbs.service.resource.FlightMapKeys;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import java.util.UUID;
+
+/** The step to create resourceId and put it in Stairway working map. state. */
```
wchamber, line 10: nit: extra step?

src/main/resources/application.yml
```diff
@@ -1,4 +1,8 @@
 rbs:
+  crl:
+    # Configurations to use Terra Cloud Resource Library. Don't need to set anything here, but CRL client needs to setup
+    # to use Janitor in test.
```
wchamber, line 4: I don't even know if I would include this since we're not setting anything

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.stream.Collectors;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+
+@AutoConfigureMockMvc
+public class RbsIntegrationTest extends BaseIntegrationTest {
+  // The AOU pool id defined in pool config.
+  // Currently
```
wchamber, line 27: formatting?

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -0,0 +1,87 @@
...
+  private static PoolId AOU_POOL_ID = PoolId.create("aou_ws_test_v1");
```
wchamber, line 28: Hmmm, I don't know about using the auo_ws_test_v1 pool. I think I'd feel better if it wasn't associated with "aou" at all. Maybe just "ws_test_v1"? and `TEST_POOL_ID`?

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Autowired
+  @Qualifier(GOOGLE_RM_COW)
```
wchamber, line 31: Is there any way to not need these qualifiers? I don't think we want to have multiple CloudResourceManagerCows or Client configs floating around.

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Autowired
+  @Qualifier(CRL_CLIENT_CONFIG)
+  ClientConfig clientConfig;
+
+  @Autowired RbsDao rbsDao;
+
+  @Test
+  public void testCreateGoogleProject() throws Exception {
+    List<Resource> resources = pollUntilResourceCreated(AOU_POOL_ID, 2, Duration.ofSeconds(10), 10);
```
wchamber, line 42: It's really odd to me that this test doesn't initiate the resource creation in anyway. Is it happening automatically by virtue of the server starting up? That seems really dangerous for testing where we might restart the server many times for any given test.

wchamber (changes requested):

src/main/resources/application.yml
yonghaoy: I thought I have it here with the comment as a reminder not to forget to set those values in test, especially when other servers take the code as an example to use CRL?

yonghaoy (commented):
@wchamber Hi Will, comments are addressed, there are some open questions left. Happy to chat offline. 
Also add a test for verify ResourceCreation in data base's rollback. 
PTAL. Thanks

src/main/java/bio/terra/rbs/db/PoolId.java
```diff
@@ -1,18 +1,34 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.auto.value.AutoValue;
 
 /** Wraps the id in db pool table. */
 @AutoValue
+@JsonSerialize(as = PoolId.class)
+@JsonDeserialize(builder = AutoValue_PoolId.Builder.class)
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.WRAPPER_ARRAY)
```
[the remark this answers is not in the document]
wchamber: We don't need to use Gson in the middle. `create` is the function for serializing from a string.
```
class PoolId {
  private static final String POOL_ID_MAP_KEY "poolIdKey";
  static PoolId retrieve(FlightMap map) {
       return PoolId.create(map.get(POOL_ID_MAP_KEY, String.class);
  }
  FlightMap store(FlightMap map) {
    map.put(POOL_ID_MAP_KEY, id());
    return map;
  }
```
I think the FlightMapKeys pattern is awkward. It divorces the keys from what they serialize. You can't just look at the keys and figure out how to `map.put` and `map.get`; you have to go look at an existing usage. Defining the flight map key next to to the functions about how to `put` and `get` encapsulates that logic in a single place.
Not going to die on this hill, but I'd prefer explicit to implicit serialization into FlightMap .

src/main/java/bio/terra/rbs/service/resource/flight/CreateGoogleProjectStep.java
wchamber: Sorry, I didn't mean duplication, I meant retries of this step. Let's say we create the project, but fail out polling the operation. Now we retry this step, and fail to create the project because we already created it, and we have to abort the step altogether.

We can write a helper function that does the 403 interpreting correctly, e.g. https://github.com/DataBiosphere/terra-workspace-manager/blob/1322b1cf98e094f0b68cf38b3c1db38e5655ccec/src/main/java/bio/terra/workspace/service/workspace/flight/CreateProjectStep.java#L73-L84

src/main/java/bio/terra/rbs/service/resource/flight/CreateGoogleProjectStep.java
wchamber: I strongly disagree. That's relying on us to look at the log and care about the volume of this error because it's obscuring other important errors. We don't actually care about this step failing, we care about it failing frequently. The right way to alert for that is custom metrics looking at step success rate, or probably flight failure rate. And what's "frequent enough for us to care?" 1/minute? 1/10 steps?
Error means somebody should care about looking at this log. This is more like it was interesting that the step failed, but that's expected behavior. go/java-logging-best-practices#logging-levels-and-avoiding-spam https://g3doc.corp.google.com/base/g3doc/logging.md?cl=head#c-logging-levels

src/main/java/bio/terra/rbs/service/resource/flight/CreateGoogleProjectStep.java
wchamber: No, the undo starts on the same step. Otherwise a step would have to know about the actions of the subsequential step.
In undo, we need to be prepared for the `do` to have failed at any point, or even succeeded.

src/main/java/bio/terra/rbs/service/resource/flight/GenerateGoogleProjectIdStep.java
wchamber: YAGNI. Start simple, then add more disambiguation as needed. A rename isn't hard, but in the meantime the longer name takes up more space and eye-time.

README.md
```diff
@@ -2,6 +2,12 @@
 Cloud resource buffering server for Terra. 
 
 # Deployment
+## Configs Rendering
+Local Testing and Github Action tests require credentials to be able to call GCP, run
+``` local-dev/render-config.sh``` first for local testing. It generates:
+* A Google Service Account Secret to create/delete cloud resources in test.
+* A Google Service Account Secret to publish message to 'prod' Janitor instance.
```
wchamber, line 9: nit: remove 'prod', just publish to a Janitor instance.

src/main/java/bio/terra/rbs/app/configuration/CrlConfiguration.java
```diff
@@ -0,0 +1,123 @@
...
+  /** How long to keep the resource before Janitor do the cleanup. */
+  public static final Duration TEST_RESOURCE_TIME_TO_LIVE = Duration.ofMinutes(30);
+
+  /**
+   * Whether we're running RBS in test mode with Cloud Resource Library. If so, we enable to the
+   * Janitor to auto-delete all created cloud resources.
+   */
+  private boolean testingMode = false;
+
+  /** Credential file path to be able to publish message to Janitor test env (toolsalpha). */
```
wchamber, line 41: not `toolsalpha`. finish fixing this comment.

src/main/java/bio/terra/rbs/app/configuration/CrlConfiguration.java
```diff
@@ -0,0 +1,123 @@
...
+  private String janitorClientCredentialFilePath;
+
+  /** pubsub project id to publish track resource to Janitor */
+  private String janitorTrackResourceProjectId;
+
+  /** pubsub topic id to publish track resource to Janitor */
+  private String janitorTrackResourceTopicId;
```
wchamber, line 48: I think it's worth noting that we only use these fields when `testingMode` is true.

src/main/java/bio/terra/rbs/service/resource/FlightManager.java
```diff
@@ -28,25 +29,26 @@ public FlightManager(FlightFactory flightFactory, StairwayComponent stairwayComp
     this.stairway = stairwayComponent.get();
   }
 
-  public boolean submitCreationFlight(Pool pool) {
-    // TODO: Add input into FlightMap
-    return submitToStairway(
-        flightFactory.getCreationFlightClass(pool.resourceType()), new FlightMap());
+  public Optional<String> submitCreationFlight(Pool pool) {
```
wchamber, line 32: needs a comment about what this does and what the return type is.

src/main/java/bio/terra/rbs/service/resource/FlightManager.java
```diff
@@ -28,25 +29,26 @@ public FlightManager(FlightFactory flightFactory, StairwayComponent stairwayComp
...
+    FlightMap flightMap = new FlightMap();
+    flightMap.put(FlightMapKeys.POOL_ID, pool.id());
+    flightMap.put(FlightMapKeys.RESOURCE_CONFIG, pool.resourceConfig());
```
wchamber, line 35: I'd kind of rather this happens as a part of the flightFactory, like Janitor where the Flight class is also allowed to set up the FlightMap. Then we can have a static function on the Flight class for initializing the Flight map. Otherwise, this code that sets up the FlightMap is far away from the steps that care about the input FlightMap.
https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/main/java/bio/terra/janitor/service/cleanup/FlightSubmissionFactoryImpl.java#L14-L19

src/main/java/bio/terra/rbs/service/resource/flight/FinishResourceCreationStep.java
```diff
@@ -0,0 +1,37 @@
...
+import bio.terra.rbs.generated.model.CloudResourceUid;
+import bio.terra.rbs.service.resource.FlightMapKeys;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+
+/**
+ * The initial step to create resource. It creates a entity in resource db table with CREATING
+ * state.
```
wchamber, line 13: fix the comment

src/main/java/bio/terra/rbs/service/resource/flight/FinishResourceCreationStep.java
```diff
@@ -0,0 +1,37 @@
...
+
+    rbsDao.updateResourceAsReady(
+        workingMap.get(FlightMapKeys.RESOURCE_ID, ResourceId.class),
+        workingMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class));
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    // Nothing need to do.
```
wchamber, line 34: still need to expand on this, see previous PR review comment on same line.

src/test/java/bio/terra/rbs/db/PoolIdTest.java
```diff
@@ -0,0 +1,25 @@
+package bio.terra.rbs.db;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import bio.terra.rbs.common.BaseUnitTest;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Test;
+
+/** Unit tests to verify {@link PoolId} is serializable/deserializable by Jackson. */
```
wchamber, line 9: It's not just that's its serializable. We critically need to maintain backwards compatibility since this gets stored in Stairway's database. Include that in the comment and the deserialize test function name.

src/test/java/bio/terra/rbs/db/RbsDaoTest.java
```diff
@@ -128,14 +131,33 @@ public void retrievePoolWithResourceState() {
...
+  public void createRetrieveDeleteResource() {
     PoolId poolId = PoolId.create("poolId");
     Pool pool = newPool(poolId);
     rbsDao.createPools(ImmutableList.of(pool));
     Resource resource = newResource(poolId, ResourceState.CREATING);
 
     rbsDao.createResource(resource);
     assertEquals(resource, rbsDao.retrieveResource(resource.id()).get());
+
+    rbsDao.deleteResource(resource.id());
```
wchamber, line 143: `assertTrue(rbsDao.deleteResource...`

src/test/java/bio/terra/rbs/db/ResourceIdTest.java
```diff
@@ -0,0 +1,30 @@
+package bio.terra.rbs.db;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import bio.terra.rbs.common.BaseUnitTest;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
+
+/** Unit tests to verify {@link ResourceId} is serializable/deserializable by Jackson. */
```
wchamber, line 10: ditto

src/test/java/bio/terra/rbs/service/resource/FlightManagerTest.java
```diff
@@ -0,0 +1,140 @@
...
+  @Autowired StairwayComponent stairwayComponent;
+
+  @Autowired RbsDao rbsDao;
+
+  /**
+   * Test inserted resource is deleted from DB if the following steps fail. Note that this doesn't
+   * cover the case RBS/Stairway crashes during the flight.
+   */
+  @Test
+  public void undoCreateResourceEntityStep_shouldDeleteDbRecord() throws Exception {
```
wchamber, line 42: I don't see why this is a FlightManager test. It seems like it's testing a single step.

Can we test all of the project creation steps individually? Or the project creation flight class directly? Or at least all of the project creation flight class steps together with interspsersed latch and error?
I don't think this is doing a good job of testing do/undo for the flight.

wchamber (changes requested):
Next time break it up before sending it out. 900 lines is too much. Possible pieces you could have done for this PR.
* config changes including rendering.
* RbsDao changes.
* Related steps, but not the whole flight, e.g. the resource id generation and store in dao steps. or the project id creation and project creation steps.

src/main/java/bio/terra/rbs/db/PoolId.java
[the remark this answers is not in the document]
yonghaoy: I see what you mean. Changed to the way you suggest. But I also keep the Jackson annotations because I thought that is something good to have to let Jackson seralize/deseralize autovalue object?

src/main/java/bio/terra/rbs/db/PoolId.java
```diff
@@ -1,23 +1,21 @@
...
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.auto.value.AutoValue;
 
-/**
- * Wraps the id in db pool table.
- *
- * <p>we're storing this as Json for the sake of Stairway FlightMap serialization/deserialization.
- * Stairway FlightMap uses Jackson, so making this serializable/deserializable by Jackson.
- */
+/** Wraps the id in db pool table. */
```
wchamber, line 11: nit: not new, but maybe - `The unique identifier for every Pool.`
it's not just about what's in the db.

src/main/java/bio/terra/rbs/db/PoolId.java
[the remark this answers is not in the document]
wchamber: I don't think we should have jackson annotations we're not using. That's confusing until we need them. We're not using jackson with the FlightMap functions, so let's remove them (and the builder) for now.

src/main/java/bio/terra/rbs/service/resource/FlightManager.java
```diff
@@ -29,13 +29,15 @@ public FlightManager(FlightFactory flightFactory, StairwayComponent stairwayComp
     this.stairway = stairwayComponent.get();
   }
 
+  /** Submit Stairway Flight to create resource. */
   public Optional<String> submitCreationFlight(Pool pool) {
     FlightMap flightMap = new FlightMap();
-    flightMap.put(FlightMapKeys.POOL_ID, pool.id());
+    flightMap = pool.id().store(flightMap);
```
wchamber, line 35: we don't need the re-assignment. We're only returning the flightMap so that it can be used fluently, e.g.
`pool.id().store(flightMap).put(FlightMapKeys.RESOURCE_CONFIG, poolresourceConfig();`
The input flight map is mutated.
maybe just stop returning the FlightMap if that doesn't seem very helpful.

src/main/java/bio/terra/rbs/service/resource/flight/CreateGoogleProjectStep.java
```diff
@@ -29,17 +31,21 @@ public CreateGoogleProjectStep(CloudResourceManagerCow rmCow, GcpProjectConfig g
...
+      }
+      Project project =
+          new Project()
+              .setProjectId(projectId)
+              .setParent(
+                  new ResourceId().setType("folder").setId(gcpProjectConfig.getParentFolderId()));
+      OperationCow<?> operation =
+          rmCow.operations().operationCow(rmCow.projects().create(project).execute());
+      pollUntilSuccess(operation, Duration.ofSeconds(10), Duration.ofMinutes(5));
+    } catch (IOException | RetryException e) {
```
wchamber, line 47: I don't think you want to catch a RetryException and re-wrap it in a retry StepResult. It's ok to not catch the Retry and allow the doStep to throw RetryException.
as-is, you're double wrapping exceptions.
The other approach would be to have pollUntilSuccess throw IOExceptoin | InterruptedException then wrap them both here.

src/main/java/bio/terra/rbs/service/resource/flight/FinishResourceCreationStep.java
```diff
@@ -24,14 +24,16 @@ public StepResult doStep(FlightContext flightContext) {
...
         workingMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class));
     return StepResult.getStepResultSuccess();
   }
 
   @Override
   public StepResult undoStep(FlightContext flightContext) {
-    // Nothing need to do.
+    // Nothing need to do. Unto a resource creation flight needs to delete the resource from Cloud
+    // and delete entity
+    // from db. Both steps are handled in previous steps undo method.
```
wchamber: Actually, don't we want to put something in the workingMap after we've updated the resource as READY and check that in *every* undo step before deleting anything? I thought that was how we were going to avoid handing out a resource and then deleting it.
This PR or the next, worth checking that behavior.

src/main/java/bio/terra/rbs/service/resource/flight/GoogleProjectCreationFlight.java
```diff
@@ -24,7 +24,7 @@ public GoogleProjectCreationFlight(FlightMap inputParameters, Object application
         new RetryRuleFixedInterval(/* intervalSeconds =*/ 180, /* maxCount =*/ 5);
     addStep(new GenerateResourceIdStep());
     addStep(new CreateResourceDbEntityStep(rbsDao));
-    addStep(new GenerateGoogleProjectIdStep());
+    addStep(new GenerateProjectIdStep());
     addStep(new CreateGoogleProjectStep(rmCow, gcpProjectConfig), retryRule);
```
wchamber, line 28: I guess I'll make the same argument that this should be `CreateProjectStep`

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -20,32 +29,28 @@
...
-        Pool.builder()
-            .id(poolId)
-            .status(PoolStatus.ACTIVE)
-            .size(2)
-            .creation(Instant.now())
-            .resourceType(ResourceType.GOOGLE_PROJECT)
-            .resourceConfig(resourceConfig)
-            .build();
-    rbsDao.createPools(ImmutableList.of(pool));
+    poolService.updateFromConfig(loadPoolConfig("test/config"), transactionStatus);
```
wchamber, line 50: uh, what's up with the null transactionStatus? worth a comment about what's going on at aleast.

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -68,13 +73,50 @@ public void testCreateGoogleProject() throws Exception {
...
+                new ResourceConfig()
+                    .configName("configName")
+                    .gcpProjectConfig(new GcpProjectConfig().projectIDPrefix("prefix")))
+            .status(PoolStatus.ACTIVE)
+            .creation(Instant.now())
+            .build();
+
+    rbsDao.createPools(ImmutableList.of(pool));
+    assertTrue(rbsDao.retrieveResources(ResourceState.CREATING, 1).isEmpty());
+    String flightId = manager.submitCreationFlight(pool).get();
```
wchamber, line 100: the first test in this class tests that the pools are filled automatically. This test is about how the CreateProjectFlight reacts to an error. I think these should be in different classes. Maybe add a `CreateProjectFlightIntegrationTest`?

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -85,9 +127,61 @@ public void testCreateGoogleProject() throws Exception {
...
+      throws InterruptedException, DatabaseOperationException {
+    Duration maxWait = Duration.ofSeconds(10);
+    Duration waited = Duration.ZERO;
+    while (waited.compareTo(maxWait) < 0) {
+      if (!stairwayComponent.get().getFlightState(flightId).isActive()) {
+        return;
+      }
+      int pollMs = 100;
+      waited.plus(Duration.ofMillis(pollMs));
+      TimeUnit.MILLISECONDS.sleep(pollMs);
```
wchamber, line 140: nit: define your variable as a Duration instead of an int and convert it to an int for the sleep. Always prefer Duration objects to primitive int values.
```
Duration poll = Duration.ofMillis(100);
TimeUnit.MILLISECONDS.sleep(poll.toMillis())
```

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -68,13 +73,50 @@ public void testCreateGoogleProject() throws Exception {
...
+    String flightId = manager.submitCreationFlight(pool).get();
+    // Resource is created in db
+    Resource resource =
+        pollUntilResourceExists(ResourceState.CREATING, poolId, 1, Duration.ofSeconds(10), 10)
+            .get(0);
+
+    LatchStep.releaseLatch();
+    blockUntilFlightComplete(flightId);
+    // Resource is deleted.
+    assertFalse(rbsDao.retrieveResource(resource.id()).isPresent());
```
wchamber, line 109: check the project is being deleted?

wchamber (changes requested):

src/main/java/bio/terra/rbs/service/resource/FlightManager.java
yonghaoy: Let store return void instead.

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
yonghaoy: The project is never created. We want to verify the rollback can success even project doesn't exist.
Also, it's hard to get generated project id, ResourceUid is populated at the end of the flight. 
So I only verify the entire flight rollback successfully(by checking db).

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
yonghaoy: Done

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
yonghaoy: Added some comment above and rename it.

src/main/java/bio/terra/rbs/service/resource/flight/FinishResourceCreationStep.java
yonghaoy: Okay, I will improve it here in next PR. Just to confirm. If resource state is handout, we will return fatal in undo?

src/main/java/bio/terra/rbs/service/resource/flight/CreateProjectStep.java
```diff
@@ -44,7 +44,7 @@ public StepResult doStep(FlightContext flightContext) {
       OperationCow<?> operation =
           rmCow.operations().operationCow(rmCow.projects().create(project).execute());
       pollUntilSuccess(operation, Duration.ofSeconds(10), Duration.ofMinutes(5));
-    } catch (IOException | RetryException e) {
+    } catch (IOException | InterruptedException | RetryException e) {
```
wchamber: don't catch `RetryException`?

src/test/java/bio/terra/rbs/integration/IntegrationUtils.java
```diff
@@ -0,0 +1,36 @@
...
+
+/** Utilities used in integration test. */
+public class IntegrationUtils {
+  public static List<Resource> pollUntilResourceExists(
+      RbsDao rbsDao,
+      ResourceState state,
+      PoolId poolId,
+      int expectedResourceNum,
+      Duration period,
+      int maxNumPolls)
```
wchamber, line 20: 6 arguments is too many. Can we just use default values for `period` and `maxNumPolls`? Do these really need to vary? 
Let's put the `int expectedResourceNum` next to the `ResourceState`. Or do we want to use a Multiset<ResourceState> or a Map<ResourceState, Integer>? 
`pollUntilResourceExists` doesn't sound like it's examining the `ResourceState`. maybe `pollUntilResourcesMatch` ? `pollUntilResourceCountsAre`?

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -38,16 +25,15 @@
 
   @Autowired RbsDao rbsDao;
   @Autowired PoolService poolService;
-  @Autowired StairwayComponent stairwayComponent;
 
-  TransactionStatus transactionStatus;
-  FlightFactory flightFactory = new TestingFlightFactory();
+  // Requreid by the method but not used in test.
+  TransactionStatus unusedTransactionStatus;
```
wchamber, line 30: This smells weird to me. It's a PoolService implementation detail that the transactionStatus can be null. Maybe we should instead annotate the PoolService methods with `@Transactional` probably work for a follow up PR. Does this look odd to you?
if we're never initializing it, maybe just inline it as null.

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -73,42 +59,6 @@ public void testCreateGoogleProject() throws Exception {
...
-
-    LatchStep.releaseLatch();
-    blockUntilFlightComplete(flightId);
-    // Resource is deleted.
-    assertFalse(rbsDao.retrieveResource(resource.id()).isPresent());
-  }
-
   private List<Resource> pollUntilResourceExists(
       ResourceState state, PoolId poolId, int expectedResourceNum, Duration period, int maxNumPolls)
       throws Exception {
```
wchamber: duplicated by IntegrationTestUtils?

src/test/java/bio/terra/rbs/service/resource/flight/ErrorCreateProjectStep.java
```diff
@@ -6,10 +6,9 @@
 import bio.terra.stairway.StepResult;
 import bio.terra.stairway.StepStatus;
 
-/** Dummy {@link CreateGoogleProjectStep} which fails in doStep. */
-public class ErrorCreateGoogleProjectStep extends CreateGoogleProjectStep {
-  public ErrorCreateGoogleProjectStep(
-      CloudResourceManagerCow rmCow, GcpProjectConfig gcpProjectConfig) {
+/** Dummy {@link CreateProjectStep} which fails in doStep. */
+public class ErrorCreateProjectStep extends CreateProjectStep {
```
wchamber, line 10: Can we include this in the test file where it's used? I don't think we should need to have this in multiple test files...

src/test/java/bio/terra/rbs/service/resource/flight/ErrorCreateProjectStep.java
```diff
@@ -6,10 +6,9 @@
 import bio.terra.stairway.StepResult;
 import bio.terra.stairway.StepStatus;
 
-/** Dummy {@link CreateGoogleProjectStep} which fails in doStep. */
-public class ErrorCreateGoogleProjectStep extends CreateGoogleProjectStep {
-  public ErrorCreateGoogleProjectStep(
-      CloudResourceManagerCow rmCow, GcpProjectConfig gcpProjectConfig) {
+/** Dummy {@link CreateProjectStep} which fails in doStep. */
```
wchamber, line 9: `but still runs undoStep?`

src/main/java/bio/terra/rbs/service/resource/flight/FinishResourceCreationStep.java
```diff
@@ -0,0 +1,38 @@
...
+        ResourceId.retrieve(flightContext.getWorkingMap()),
+        workingMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class));
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    // Nothing need to do. Unto a resource creation flight needs to delete the resource from Cloud
+    // and delete entity from db. Both steps are handled in previous steps undo method.
+    return StepResult.getStepResultSuccess();
```
wchamber, line 36: (can't figure out how to reply to the comment chaing)
You tell me what your doc says to do. At least TODO the ticket to implement this correctly.

src/test/java/bio/terra/rbs/service/resource/flight/LatchStep.java
```diff
@@ -0,0 +1,37 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import java.util.concurrent.CountDownLatch;
+
+/**
+ * A step to block until the latch is released to be used for testing.
+ *
+ * <p>This step relies on in-memory state and does not work across services or after a service
+ * restart. It is only useful for testing.
```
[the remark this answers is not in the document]
wchamber: probably still good to have. we might want some stairway-contrib-testing-utils one day.

wchamber (changes requested):

src/test/java/bio/terra/rbs/integration/IntegrationUtils.java
yonghaoy: make period and maxNumPolls constant now. 
And put the int expectedResourceNum next to the ResourceState.
I didn't merge those in `Map` or `Multiset` because I think the method only need to check single state. Add those extra logic to process Map in caller and here is unnecessary. `Pair` would be a good option through...

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
yonghaoy: Inlined as null for now here. 
I thought TransactionalStatus might be helpful if we want it. But seems we are not using that one for now. Will check with you offline about how we will future change the poolService's transactional method stuff later as follow up

src/main/java/bio/terra/rbs/service/resource/flight/FinishResourceCreationStep.java
yonghaoy: Added a TODO for now. I can not create ticket under any epics now. I will create a ticket then once that issue is resolved and when I worked on it. 
BTW, can you still create tickets under Janitor/RBS epic?

src/main/java/bio/terra/rbs/service/resource/flight/FinishResourceCreationStep.java
```diff
@@ -0,0 +1,37 @@
...
+
+    rbsDao.updateResourceAsReady(
+        ResourceId.retrieve(flightContext.getWorkingMap()),
+        workingMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class));
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    // TODO(yonghao): Fail the flight if resource is already handed out.
```
wchamber, line 34: just TODO(PF-127). We're not done with Stairway project creation until undo works correctly.

wchamber (approved):

PR #19, opened 2020-10-16
[PF-127] Combines flight map and flight class in one factory. 

yonghaoy:
1. Use `FlightSubmision` which contains flight map and flight class


src/main/java/bio/terra/rbs/service/pool/PoolService.java
```diff
@@ -44,13 +40,13 @@ public void initialize() {
...
-      transactionTemplate.execute(status -> updateFromConfig(parsedPoolConfigs, status));
+      updateFromConfig(parsedPoolConfigs);
     }
   }
 
   @VisibleForTesting
-  public boolean updateFromConfig(
-      List<PoolWithResourceConfig> parsedPoolConfigs, TransactionStatus status) {
+  @Transactional
+  public boolean updateFromConfig(List<PoolWithResourceConfig> parsedPoolConfigs) {
```
wchamber, line 49: Spring annotations only work when they're called across beans. This means this method will not be used transactionally anymore when it's called from initialize (within the same bean).
I think you should internalize the transaction wrapping within this method, instead of doing it in initialize.

src/main/java/bio/terra/rbs/service/resource/FlightSubmissionFactoryImpl.java
```diff
@@ -1,15 +1,18 @@
...
+import bio.terra.stairway.FlightMap;
 import com.google.common.collect.ImmutableMap;
 import java.util.Map;
 import org.springframework.stereotype.Component;
 
 @Component
-public class FlightFactoryImpl implements FlightFactory {
+public class FlightSubmissionFactoryImpl implements FlightSubmissionFactory {
   /** Supported resource creation flight map. */
   Map<ResourceType, Class<? extends Flight>> CREATION_FLIGHT_MAP =
```
wchamber, line 17: nit: make both of these maps `private static final ImmutableMap`. these are constants and it's better to use ImmutableMap for constants.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -65,8 +66,8 @@ public void testCreateGoogleProject_errorDuringProjectCreation() throws Exceptio
   @Test
   public void errorCreateProject_noRollbackAfterResourceReady() throws Exception {
     // Verify project and db entity won't get deleted if resource id READY, even the flight fails.
-    TestingFlightFactory.setFlightClassToUse(ErrorAfterCreateResourceFlight.class);
-    FlightManager manager = new FlightManager(flightFactory, stairwayComponent);
+    TestingSubmissionFlightFactory.setFlightClassToUse(ErrorAfterCreateResourceFlight.class);
+    FlightManager manager = new FlightManager(flightSubmissionFactory, stairwayComponent);
```
wchamber, line 70: consider instantiating and inlining the TestingSubmissionFlightFactory instead of having a shared one that gets modified. That's clearer lifetimes.

`FlightManager manager = new FlightManager(new TestingSubmissionFlightFactory(ErrorAfterCreateResourceFlight.class));`

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -144,22 +145,25 @@ private Pool preparePool() {
         .creation(Instant.now())
         .build();
   }
-  /** A {@link FlightFactory} used in test. */
-  public static class TestingFlightFactory implements FlightFactory {
+  /** A {@link FlightSubmissionFactory} used in test. */
+  public static class TestingSubmissionFlightFactory implements FlightSubmissionFactory {
```
wchamber, line 149: `Stub` is more descriptive than `Testing`.

src/test/java/bio/terra/rbs/service/resource/FlightSubmissionFactoryImplTest.java
```diff
@@ -0,0 +1,35 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import bio.terra.rbs.common.*;
+import bio.terra.rbs.service.resource.flight.GoogleProjectCreationFlight;
+import bio.terra.stairway.FlightMap;
+import java.util.UUID;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class FlightSubmissionFactoryImplTest {
```
wchamber, line 12: I think this is overkill to test a stub. Also, I don't think this is being run since it's not annotated as a test class or extending BaseUnitTest.

wchamber (changes requested):
Next time, split independent changes into independent PRs.

src/test/java/bio/terra/rbs/service/resource/FlightSubmissionFactoryImplTest.java
yonghaoy: okay deleted.

wchamber (approved):

PR #21, opened 2020-10-16
[PF-183] Billing info setup 

yonghaoy:


yonghaoy:
Actually we need to enable billing api for the project we are using in test. Having a terraform for that...

src/main/java/bio/terra/rbs/service/resource/flight/GoogleProjectCreationFlight.java
```diff
@@ -26,6 +29,10 @@ public GoogleProjectCreationFlight(FlightMap inputParameters, Object application
     addStep(new CreateResourceDbEntityStep(rbsDao));
     addStep(new GenerateProjectIdStep());
     addStep(new CreateProjectStep(rmCow, gcpProjectConfig), retryRule);
+    if (gcpProjectConfig.getBillingAccount() != null
+        && !gcpProjectConfig.getBillingAccount().isEmpty()) {
+      addStep(new SetBillingInfoStep(billingCow, gcpProjectConfig));
```
wchamber, line 34: maybe always add the step and check in the step if anything needs to be done based on the gcpProjectConfig? That might be a little simpler for us to debug Stairway things which care about the number of steps.

src/main/java/bio/terra/rbs/service/resource/flight/SetBillingInfoStep.java
```diff
@@ -0,0 +1,37 @@
...
+import static bio.terra.rbs.service.resource.FlightMapKeys.GOOGLE_PROJECT_ID;
+
+import bio.terra.cloudres.google.billing.CloudBillingClientCow;
+import bio.terra.rbs.generated.model.GcpProjectConfig;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import com.google.cloud.billing.v1.ProjectBillingInfo;
+
+/** Creates the basic GCP project. */
```
wchamber, line 12: update

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -50,16 +56,23 @@ public void testCreateGoogleProject() throws Exception {
...
+  private void assertProjectMatch(CloudResourceUid resourceUid, GcpProjectConfig gcpProjectConfig)
+      throws Exception {
     Project project =
         rmCow.projects().get(resourceUid.getGoogleProjectUid().getProjectId()).execute();
+    assertEquals(
+        gcpProjectConfig.getBillingAccount(),
+        billingCow
+            .getProjectBillingInfo("projects/" + project.getProjectId())
+            .getBillingAccountName());
     assertEquals("ACTIVE", project.getLifecycleState());
```
wchamber, line 76: group the project get with the project lifecycle assertion - don't have the billing check in the middle.

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
```diff
@@ -50,16 +56,23 @@ public void testCreateGoogleProject() throws Exception {
...
-  private void assertProjectMatch(CloudResourceUid resourceUid) throws Exception {
+  private void assertProjectMatch(CloudResourceUid resourceUid, GcpProjectConfig gcpProjectConfig)
+      throws Exception {
     Project project =
         rmCow.projects().get(resourceUid.getGoogleProjectUid().getProjectId()).execute();
+    assertEquals(
+        gcpProjectConfig.getBillingAccount(),
+        billingCow
+            .getProjectBillingInfo("projects/" + project.getProjectId())
+            .getBillingAccountName());
```
wchamber, line 75: we're starting to get more detailed about checking that the project creation flight creates the project in the correct way. I think that kind of check is more appropriate for the `CreateProjectFlightIntegrationTest`. I also want a test that having a null billing account still allows a project to be created.

The tests in this file are just about the end2end, not the particulars how created projects.

wchamber (changes requested):

src/main/java/bio/terra/rbs/service/resource/flight/GoogleProjectCreationFlight.java
yonghaoy: done

src/main/java/bio/terra/rbs/service/pool/PoolWithResourceConfig.java
```diff
@@ -7,9 +7,9 @@
 /** Represents a {@link PoolConfig} with corresponding {@link ResourceConfig}. */
 @AutoValue
 public abstract class PoolWithResourceConfig {
-  public abstract PoolConfig poolConfig();
+   abstract PoolConfig poolConfig();
 
-  public abstract ResourceConfig resourceConfig();
+   abstract ResourceConfig resourceConfig();
```
wchamber: nit: it's funny that this class is public but it's methods are package private. I know they aren't currently used somewhere, but maybe keep them public for simplicity and so we don't have to change it later?

src/test/java/bio/terra/rbs/integration/RbsIntegrationTest.java
yonghaoy: okay, moved to CreateProjectFlightIntegrationTest.

src/main/java/bio/terra/rbs/service/resource/flight/SetBillingInfoStep.java
```diff
@@ -21,6 +21,11 @@ public SetBillingInfoStep(CloudBillingClientCow billingCow, GcpProjectConfig gcp
 
   @Override
   public StepResult doStep(FlightContext flightContext) {
+    // Skip if billing account is not set.
+    if (gcpProjectConfig.getBillingAccount() != null
+        && !gcpProjectConfig.getBillingAccount().isEmpty()) {
+      return StepResult.getStepResultSuccess();
```
wchamber, line 27: uh, this looks backwards to me. Shouldn't we skip if the billing account is null or is an empty string?
How are the tests passing? Am I reading this wrong?

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -31,10 +34,35 @@
...
+    String flightId = manager.submitCreationFlight(pool).get();
+    blockUntilFlightComplete(flightId);
+    assertProjectCreated(pool);
+  }
+
+  @Test
+  public void testCreateGoogleProject_withBillingAccount() throws Exception {
+    // Basic GCP project with billing setup.
+    FlightManager manager = new FlightManager(flightSubmissionFactoryImpl, stairwayComponent);
+    Pool pool = preparePool(newBasicGcpConfig().billingAccount(BILLING_ACCOUNT_NAME));
```
wchamber, line 60: the basic project has a billing account, so we need a test for one without a billing account

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -129,22 +155,51 @@ private void blockUntilFlightComplete(String flightId)
...
+            .get(resource.cloudResourceUid().getGoogleProjectUid().getProjectId())
+            .execute();
+    assertEquals("ACTIVE", project.getLifecycleState());
+    if (pool.resourceConfig().getGcpProjectConfig().getBillingAccount() != null) {
+      assertEquals(
+          pool.resourceConfig().getGcpProjectConfig().getBillingAccount(),
+          billingCow
+              .getProjectBillingInfo("projects/" + project.getProjectId())
+              .getBillingAccountName());
+    }
```
wchamber, line 200: nit: I think it'd be better to turn this into multiple asserts. I don't like the conditional asserting based on input parameters - that's too much logic for a test assertion.
maybe `void assertProjectExists(Resource resource)` and `assertBillingIs(Resource resource, String billingAccount)`?

wchamber (changes requested):

src/main/java/bio/terra/rbs/service/resource/flight/SetBillingInfoStep.java
yonghaoy: oh, right. Test does fails actually. Test passing now...

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: changed here to: Basic doesn't have billing account. We only set billing account in this test here.

terra-rbs-client/build.gradle
```diff
@@ -0,0 +1,108 @@
+plugins {
```
wchamber, line 1: uh, I think adding the `terra-rbs-client/build.gradle` should be a separate PR. Please remove. Careful with your commits and branches.

wchamber (approved):

terra-rbs-client/build.gradle
yonghaoy: ah, strange, it is in his PR. My local branchs got messed up somehow. Removed

PR #22, opened 2020-10-19
[PF-139] publish client library for rbs

yonghaoy:
https://broadinstitute.jfrog.io/broadinstitute/webapp/#/artifacts/browse/tree/General/libs-snapshot-local/bio/terra/rbs/terra-rbs-client


wchamber (approved):
It looks like the other clients are published without a prefix folder, e.g. `terra-rbs-client` instead of `rbs/terra-rbs-client`. Can we match that style?
https://broadinstitute.jfrog.io/broadinstitute/webapp/#/artifacts/browse/tree/General/libs-snapshot-local/bio/terra/rbs/terra-rbs-client

PR #23, opened 2020-10-19
[PF-183]: Enable services support

yonghaoy:


src/main/java/bio/terra/rbs/service/resource/flight/StepUtils.java
```diff
@@ -47,4 +47,9 @@ public static boolean isResourceReady(FlightContext flightContext) {
     return workingMap.get(RESOURCE_READY, Boolean.class) != null
         && workingMap.get(RESOURCE_READY, Boolean.class);
   }
+
+  /** Converts project id to name. */
+  public static String projectIdToName(String projectId) {
```
wchamber, line 52: I don't know that this is useful outside of the billing api. Let's keep it as a private method in there for now before making it widely shared.

`StepUtils` is way to general. It's just going to be a dumping ground for everything. Consider how to break up utilities classes so that they stay focused. Could have OperationUtils, or GoogleUtils and distinguish from resource creation utils.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -65,6 +80,21 @@ public void testCreateGoogleProject_withBillingAccount() throws Exception {
...
+    // Basic GCP project with billing setup and api enabled.
+    FlightManager manager = new FlightManager(flightSubmissionFactoryImpl, stairwayComponent);
+    Pool pool =
+        preparePool(
+            newBasicGcpConfig().billingAccount(BILLING_ACCOUNT_NAME).enabledApis(ENABLED_SERVICES));
+
+    String flightId = manager.submitCreationFlight(pool).get();
+    blockUntilFlightComplete(flightId);
+    Project project = assertProjectExists(pool);
+    assertBillingIs(project, pool.resourceConfig().getGcpProjectConfig().getBillingAccount());
```
wchamber, line 94: I don't think this test needs to check billing

src/test/java/bio/terra/rbs/integration/IntegrationUtils.java
```diff
@@ -30,4 +32,12 @@
     }
     throw new InterruptedException("Polling exceeded maxNumPolls");
   }
+
+  /**
+   * Create a string matching the service name on {@link GoogleApiServiceusageV1Service#getName()},
+   * e.g. projects/123/services/serviceusage.googleapis.com.
+   */
+  public static String serviceName(Project project, String apiId) {
```
wchamber, line 40: ditto, don't make thees public util functions until there's a need to use them broadly.
Also consider if we want to add these as static functions to CRL.

wchamber (commented):
why are the tests failing?

src/main/java/bio/terra/rbs/service/resource/flight/StepUtils.java
yonghaoy: this is used in ServiceUsage and Billing now. And I think we will need this for other steps as well. Created a GoogleUtils for this

src/test/java/bio/terra/rbs/integration/IntegrationUtils.java
yonghaoy: right. moved to where it is get used.
As for static functions, not sure letting CRL having those simpler helper functions is necessary... I didn't find too many useful utils classes in Google's APIs. We can discuss more.

wchamber (approved):

PR #24, opened 2020-10-20
[PF-183] Add IAM policy for project

yonghaoy:


src/main/java/bio/terra/rbs/service/resource/flight/SetIamPolicyStep.java
```diff
@@ -0,0 +1,65 @@
...
+      gcpProjectConfig
+          .getIamBindings()
+          .forEach(
+              iamBinding ->
+                  policy
+                      .getBindings()
+                      .add(
+                          new Binding()
+                              .setRole(iamBinding.getRole())
+                              .setMembers(iamBinding.getMembers())));
```
wchamber, line 48: if we want to use a stream-like for each, let's split this up so there's not so much indentation
```
getIamBindings().stream()
.map(iamBinding -> new Binding().setRole...setMembers)
.forEach(policy.getBindings()::add)
```
or do a for loop.

src/main/resources/config/resource_schema.yaml
```diff
@@ -53,4 +53,22 @@ components:
           type: array
           items:
             type: string
+        iamBindings:
+          description: |-
+            List of APIs to enable
```
wchamber, line 58: fix copy paste

src/main/resources/config/resource_schema.yaml
```diff
@@ -53,4 +53,22 @@ components:
...
+          items:
+            $ref: '#/components/schemas/IamBinding'
+      type: object
+    IamBinding:
+      description: Associates `members` with a `role`.
+      properties:
+        members:
+          description: "GCP identities, see https://cloud.google.com/resource-manager/reference/rest/Shared.Types/Binding"
+          items:
+            type: string
```
wchamber, line 69: add an example?

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -48,8 +52,22 @@
   private static final List<String> ENABLED_SERVICES =
       Arrays.asList(
           "bigquery.googleapis.com", "compute.googleapis.com", "cloudbilling.googleapis.com");
+
   private static final String ENABLED_FILTER = "state:ENABLED";
 
+  /**
+   * The groups used to test IAM policy sets up on a group. It doesn't matter what the users are for
+   * the purpose of this test. They just need to exist for Google. Currently, those groups are
+   * manually created, for Broad development, they are created via BITs service portal.
```
wchamber, line 61: nit: `these groups were manually created for Broad development via the BITs service portal.`

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -48,8 +52,22 @@
...
+   */
+  private static final String TEST_GROUP_NAME = "terra-rbs-test@broadinstitute.org";
+
+  private static final String TEST_GROUP_VIEWER_NAME = "terra-rbs-viewer-test@broadinstitute.org";
+  private static final List<IamBinding> IAM_BINDINGS =
```
wchamber, line 66: consider inlining this in the test if we're not using it elsewhere. Then all the context of the test is there and there's no need to page around to see exacetly what the test does.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -239,6 +269,20 @@ private void assertEnableApisContains(Project project, List<String> enabledApis)
...
+    Policy policy =
+        rmCow.projects().getIamPolicy(project.getProjectId(), new GetIamPolicyRequest()).execute();
+
+    List<Binding> bindings =
+        iamBindings.stream()
+            .map(
+                iamBinding ->
+                    new Binding().setRole(iamBinding.getRole()).setMembers(iamBinding.getMembers()))
+            .collect(Collectors.toList());
+    Matchers.hasItems(bindings.toArray());
```
wchamber, line 283: this isn't doing any assertion: it's creating the Matcher object.

wchamber (changes requested):

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -263,24 +260,28 @@ private void assertEnableApisContains(Project project, List<String> enabledApis)
...
             .collect(Collectors.toList());
-    Matchers.hasItems(bindings.toArray());
+    assertThat(
+        new ArrayList<>(
+            rmCow
+                .projects()
+                .getIamPolicy(project.getProjectId(), new GetIamPolicyRequest())
+                .execute()
+                .getBindings()),
+        Matchers.hasItems(bindings.toArray()));
```
wchamber, line 284: seems funny that we have to conver to an array list to use the matcher. getBindings() doesn't return a list?

wchamber (approved):

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: It returns List<>, but, I got complains saying "Cannot resolve method...." if I use `getBindings` directly...

PR #25, opened 2020-10-20
[PF-182] Initial VPC resource config

yonghaoy:
I checked deployment manager template, and didn't find any values we need to customize. So I just add two boolean for now, one fore high security network, one for privateIpGoogleAccess. Let me know if there is anything we need to customize. Thanks! 
https://github.com/broadinstitute/gcp-dm-templates/blob/eeee90fa4619b273d07206a867b01914cdeb0a30/firecloud_project.py#L100


src/main/resources/config/resource_schema.yaml
```diff
@@ -59,6 +59,15 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/IamBinding'
+        enableHightSecurityNetwork:
+          description: |-
+            Whether to create Virtual Private Cloud. If enabled, it will create network, subNets, and firewall rules, enable flow logs on the network.
```
yonghaoy, line 64: not sure if we want to have flow logs as separate boolean or just always enable it.

wchamber: no idea.

src/main/resources/config/resource_schema.yaml
```diff
@@ -59,6 +59,15 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/IamBinding'
+        enableHightSecurityNetwork:
```
wchamber, line 62: do we want to group these related fields, like both within a network object? That might be helpful for top level organization.
(can we do that inline, or do we need to create a separate object?)

gjuggler: +1, it should be easy to do inline if needed.

src/main/resources/config/resource_schema.yaml
gjuggler: Here was my takeaway from last Friday:
- We don't need a highSecurityNetwork option, since this should happen for all projects by default. This will be a change for Terra but it sounded like Doug is okay with biting off that change (since it's something we've wanted for Terra for awhile anyway).
- The key remaining difference is that for AoU we want (1) flow logs turned on, and (2) enable private google access, which causes GCS traffic to be easily identified within the network. Both of these features are directly related to AoU's requirement that network traffic is measured. So we can combine these into one boolean parameter, maybe name it `enableNetworkMetering` or `monitorNetworkTraffic`. I think I like the 2nd one; @dvoet or @wchamber any opinions?

gjuggler, line 64: nitpick on the description text: it's not quite accurate to say the option would "create Virtual Private Cloud". All GCP projects have a VPC network no matter what – the key difference here is that we're deleting the default network and manually creating all subnets & firewall rules, etc.

(But I think this option will go away based on my comment above – so feel free to mostly ignore.)

src/main/resources/config/resource_schema.yaml
gjuggler: That said... my comment above is recommending we collapse this into a single option like `enableNetworkMonitoring`, which then probably isn't worth bumping into a nested `network` object if we only have one option there.

wchamber: If this were a protobuf, I would say definitely nest it as we're likely to add more fields in the future. In JSON schema, if we add another networking field in the future, would we want to move this into that object? I think yes.

src/main/resources/config/resource_schema.yaml
wchamber: One boolean parameter sgtm. `monitorNetworkTraffic` seems fine.
I could see us  in the future getting to a place where we want to have different flavors of network traffic measurement. I think when we get there, we deprecate this field and replace it with more fields as needed. In the meanwhile, naming the parameters by what we want them to accomplish instead of the cloud mechanisms seems good. I think that does set us towards feature gating params for non-backwards compatible changes to what it means to `monitorNetworkTraffic`.

src/main/resources/config/resource_schema.yaml
yonghaoy: Thanks! Inlined.

src/main/resources/config/resource_schema.yaml
```diff
@@ -73,3 +76,12 @@ components:
           description: Role that is assigned to `members`. For example, `roles/viewer`, `roles/editor`, or `roles/owner`.
           type: string
       type: object
+
+    Network:
+      description: Network setup for the project.
+      properties:
+        enableNetworkMonitoring:
+          description: |-
+            IF enabled, turn on flow logs, enable provate google access.
```
wchamber, line 85: I don't know if we want to add more about what it means to enable network monitoring. I don't know what the right words for that would be. @gjuggler ?
typos: `If` and `private`

wchamber (approved):

src/main/resources/config/resource_schema.yaml
yonghaoy: Added a short description, not sure that is accurate. @gjuggler

gjuggler: That's pretty accurate. It may help to either link to GCP docs, and/or the AoU design doc http://broad.io/aou-egress-alerting and https://docs.google.com/document/d/1ccz2kzDL68CPofZ-b95ykQIudCgh7OCSFJ4Ym4Oz0r0/edit for more context on how & why these features are used to measure network traffic.

gjuggler (approved):

PR #28, opened 2020-10-21
[PF-131] Add handout endpoint api interface 

yonghaoy:


src/main/resources/api/service_openapi.yaml
```diff
@@ -24,7 +24,51 @@ paths:
           description: Service is broken
           $ref: '#/components/responses/StatusResponse'
 
+  '/api/pool/v1/{poolId}/resource':
+    parameters:
+      - $ref: '#/components/parameters/PoolId'
+    get:
```
wchamber, line 30: should this be a `put` (or maybe `post`?) since we're modifying state? I think the request id should also be a path parameter because REST likes to encode parameters that identify a resource in the path instead of query or body.

src/main/resources/api/service_openapi.yaml
```diff
@@ -24,7 +24,51 @@ paths:
...
+      summary: Get a resource from pool.
+      operationId: handoutResource
+      tags:
+        - rbs
+      parameters:
+        - $ref: '#/components/parameters/HandoutRequestId'
+      responses:
+        '200':
+          description: OK
```
wchamber, line 39: Explain the semantics of using the same request id multiple times.

src/main/resources/api/service_openapi.yaml
```diff
@@ -24,7 +24,51 @@ paths:
           description: Service is broken
           $ref: '#/components/responses/StatusResponse'
 
+  '/api/pool/v1/{poolId}/resource':
+    parameters:
+      - $ref: '#/components/parameters/PoolId'
+    get:
+      summary: Get a resource from pool.
```
wchamber, line 31: there's a little more going on here. the request id is assigned to the resource no more than once, we hand it out, etc.

wchamber (changes requested):

src/main/resources/api/service_openapi.yaml
yonghaoy: Changed to put and path params.
But put also feels wired, 
From client point, it is getting resource from RBS, and updating state is some internal side effect in RBS? But from the logical operation, put also make sense because we are modifying resource?

wchamber: If we think about it as the client putting a resource reservation in RBS, I think it's ok. It's definitely not a `get` because it's not a read, the client wants something to change. Yeah, this is more of an RPC than a REST endpoint, but whatever.

src/main/resources/api/service_openapi.yaml
```diff
@@ -24,19 +24,18 @@ paths:
...
-        - rbs
       parameters:
+        - $ref: '#/components/parameters/PoolId'
         - $ref: '#/components/parameters/HandoutRequestId'
+      tags:
+        - rbs
       responses:
         '200':
-          description: OK
+          description: OK, using the same handoutrequestId will also get 200 back with the same ResourceInfo.
```
wchamber: nit: camelcase s/handourequestId/handoutRequestId

src/main/resources/api/service_openapi.yaml
```diff
@@ -62,7 +61,7 @@ components:
 
     HandoutRequestId:
       name: handoutRequestId
-      in: query
+      in: path
       description: |-
         The identity present by client when asking for a resource. Using the same handoutRequestId in the same pool would ge the same resource back.
```
wchamber, line 66: nit: The unique identifier presented by the client for a resource request. Using the same...

wchamber (approved):

PR #29, opened 2020-10-21
[PF-131] Add NotFoundException

yonghaoy:
Will be used in handout resource endpoints

### DIFF
--- a/src/main/java/bio/terra/rbs/app/controller/GlobalExceptionHandler.java
+++ b/src/main/java/bio/terra/rbs/app/controller/GlobalExceptionHandler.java
@@ -1,7 +1,6 @@
 package bio.terra.rbs.app.controller;
 
 import bio.terra.rbs.common.exception.ErrorReportException;
-import bio.terra.rbs.common.exception.NotFoundException;
 import bio.terra.rbs.generated.model.ErrorReport;
 import java.util.List;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/rbs/app/controller/GlobalExceptionHandler.java
```diff
@@ -47,6 +48,11 @@
     return buildErrorReport(ex, HttpStatus.INTERNAL_SERVER_ERROR, null);
   }
 
+  @ExceptionHandler(NotFoundException.class)
+  public ResponseEntity<ErrorReport> notFoundHandler(Exception ex) {
+    return buildErrorReport(ex, HttpStatus.NOT_FOUND, null);
```
wchamber, line 53: shouldn't need this if we extend ErrorReportException, like we do for BadRequestException: https://github.com/DataBiosphere/terra-rbs/blob/8f04e7631342b8d3e65a605e699482b6afe084f7/src/main/java/bio/terra/rbs/common/exception/BadRequestException.java

src/main/java/bio/terra/rbs/common/exception/NotFoundException.java
```diff
@@ -0,0 +1,12 @@
+package bio.terra.rbs.common.exception;
+
+/** This exception maps to HttpStatus.NOT_FOUND in the GlobalExceptionHandler. */
+public class NotFoundException extends RuntimeException {
```
wchamber, line 4: ^ should extend ErrorReportException, e.g. https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/common/exception/NotFoundException.java

wchamber (changes requested):

wchamber (approved):

PR #30, opened 2020-10-21
[PF-200] Add few basic validations for GCP resource Config

yonghaoy:
Now we always create networks/subnets for a project. And it requires billing account and enabling compute api. 
So having a check here to catch error earlier.

### DIFF
--- a/src/main/java/bio/terra/rbs/service/pool/AlwaysPassValidator.java
+++ b/src/main/java/bio/terra/rbs/service/pool/AlwaysPassValidator.java
@@ -3,7 +3,7 @@ package bio.terra.rbs.service.pool;
 import bio.terra.rbs.generated.model.ResourceConfig;
 
 /** The default validator which does not no restrictions. */
-public class DefaultResourceConfigValidator implements ResourceConfigValidator {
+public class AlwaysPassValidator implements ResourceConfigValidator {
   @Override
   public void validate(ResourceConfig config) {
     // Always pass.
--- a/src/main/java/bio/terra/rbs/service/pool/AlwaysPassValidator.java
+++ b/src/main/java/bio/terra/rbs/service/pool/AlwaysPassValidator.java
@@ -2,7 +2,7 @@ package bio.terra.rbs.service.pool;
 
 import bio.terra.rbs.generated.model.ResourceConfig;
 
-/** The default validator which does not no restrictions. */
+/** The validator which does not no restrictions. */
 public class AlwaysPassValidator implements ResourceConfigValidator {
   @Override
   public void validate(ResourceConfig config) {
--- a/src/main/java/bio/terra/rbs/service/pool/DefaultResourceConfigValidator.java
+++ b/src/main/java/bio/terra/rbs/service/pool/DefaultResourceConfigValidator.java
@@ -1,0 +1,11 @@
+package bio.terra.rbs.service.pool;
+
+import bio.terra.rbs.generated.model.ResourceConfig;
+
+/** The default validator which does not require any restrictions on the config. */
+public class DefaultResourceConfigValidator implements ResourceConfigValidator {
+  @Override
+  public void validate(ResourceConfig config) {
+    // Always pass.
+  }
+}
--- a/src/main/java/bio/terra/rbs/service/pool/DefaultResourceConfigValidator.java
+++ b/src/main/java/bio/terra/rbs/service/pool/DefaultResourceConfigValidator.java
@@ -2,7 +2,7 @@ package bio.terra.rbs.service.pool;
 
 import bio.terra.rbs.generated.model.ResourceConfig;
 
-/** The default validator which does not require any restrictions on the config. */
+/** The default validator which does not no restrictions. */
 public class DefaultResourceConfigValidator implements ResourceConfigValidator {
   @Override
   public void validate(ResourceConfig config) {
--- a/src/main/java/bio/terra/rbs/service/pool/GcpResourceConfigValidator.java
+++ b/src/main/java/bio/terra/rbs/service/pool/GcpResourceConfigValidator.java
@@ -6,7 +6,13 @@ import bio.terra.rbs.generated.model.ResourceConfig;
 import com.google.common.collect.ImmutableList;
 import java.util.List;
 
-/** Validates GCP resource configs */
+/**
+ * Validates GCP resource config. The current rules are:
+ * <ul>
+ *     <li> Billing account is required.
+ *     <li> compute.googleapis.com need to be enabled.
+ * </ul>
+ * */
 public class GcpResourceConfigValidator implements ResourceConfigValidator {
   /** List of services required to be enabled. */
   private static final List<String> REQUIRED_SERVICES = ImmutableList.of("compute.googleapis.com");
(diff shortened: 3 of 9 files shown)


src/main/java/bio/terra/rbs/service/pool/DefaultResourceConfigValidator.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.rbs.service.pool;
+
+import bio.terra.rbs.generated.model.ResourceConfig;
+
+/** The default validator which does not no restrictions. */
+public class DefaultResourceConfigValidator implements ResourceConfigValidator {
```
wchamber, line 6: maybe AlwaysPassValidator? `Default` makes me think it's a safe choice everywhere, but I think that over promises.

src/main/java/bio/terra/rbs/service/pool/ResourceConfigValidator.java
```diff
@@ -0,0 +1,8 @@
+package bio.terra.rbs.service.pool;
+
+import bio.terra.rbs.generated.model.ResourceConfig;
+
+/** Validate {@link ResourceConfig} is valid. */
+public interface ResourceConfigValidator {
+  void validate(ResourceConfig config);
```
wchamber, line 7: nit: add a comment saying this throws a RuntimeException if validation does not succeed.

src/main/java/bio/terra/rbs/service/pool/GcpResourceConfigValidator.java
```diff
@@ -0,0 +1,35 @@
...
+import bio.terra.rbs.generated.model.GcpProjectConfig;
+import bio.terra.rbs.generated.model.ResourceConfig;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+/**
+ * Validates GCP resource config. The current rules are:
+ * <ul>
+ *     <li> Billing account is present.
+ *     <li> compute.googleapis.com need to be enabled.
```
wchamber, line 13: probably good to explain why they're required to help future maintainers decide if that's changed

src/main/java/bio/terra/rbs/service/pool/ResourceConfigValidatorFactory.java
```diff
@@ -0,0 +1,20 @@
+package bio.terra.rbs.service.pool;
+
+import bio.terra.rbs.common.ResourceConfigVisitor;
+import bio.terra.rbs.common.ResourceType;
+import bio.terra.rbs.generated.model.ResourceConfig;
+import org.springframework.stereotype.Component;
+
+/** Factory to return {ResourceConfigValidator} by {@link ResourceConfig}. */
+@Component
+public class ResourceConfigValidatorFactory {
```
wchamber, line 10: does this need to be a component if it only has a static function?
We could also get rid of this class and add the static function on ResourceConfigValidator. Factory seems a little overkill here since we never change what comes out of the factory (yet), but up to you.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -268,20 +245,21 @@ private void assertEnableApisContains(Project project, List<String> enabledApis)
...
-    assertThat(
-        new ArrayList<>(
-            rmCow
-                .projects()
-                .getIamPolicy(project.getProjectId(), new GetIamPolicyRequest())
-                .execute()
-                .getBindings()),
-        Matchers.hasItems(bindings.toArray()));
+    // By default we enable some services, and some GCP automatically creates SA accounts and grant
+    // permission.
```
wchamber, line 249: nit: By default we enable some services, which causes GCP to automatically create Service Accounts and grant them permissions on the project.
e.g. ...
So we need to iterate through all bindings and verify they at least contain the members & roles we expect.

src/test/java/bio/terra/rbs/service/pool/ResourceConfigValidatorTest.java
```diff
@@ -0,0 +1,45 @@
...
+public class ResourceConfigValidatorTest extends BaseUnitTest {
+  @Test
+  public void testValidateGcpConfig_success() {
+    ResourceConfig resourceConfig =
+        new ResourceConfig()
+            .configName("testConfig")
+            .gcpProjectConfig(
+                new GcpProjectConfig()
+                    .billingAccount("123")
+                    .addEnabledApisItem("compute.googleapis.com"));
```
wchamber, line 20: optional: consider having a method that returns a basic valid GcpProjectConfig, use it here, then mutate it as necessary for the other tests. That makes it easier to see what's changing between the tests, while not focusing on the necesary fields that are explicitly not important.

src/test/java/bio/terra/rbs/service/pool/ResourceConfigValidatorTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Test
+  public void testValidateGcpConfig_missingBillingAccount() {
+    ResourceConfig resourceConfig =
+        new ResourceConfig()
+            .configName("testConfig")
+            .gcpProjectConfig(new GcpProjectConfig().addEnabledApisItem("compute.googleapis.com"));
+    assertThrows(
+        InvalidPoolConfigException.class,
+        () -> new GcpResourceConfigValidator().validate(resourceConfig));
```
wchamber, line 32: for both of these assertThrows, can we verify that the message is/contains what we expect? While that does make this test more brittle for changing the messages, it makes sure that they are verifying the throwing point they expect. It can be easy to be invalid for the wrong reason and then have bugs in what's not being checked.

wchamber (changes requested):
few nits

src/main/java/bio/terra/rbs/service/pool/ResourceConfigValidatorFactory.java
yonghaoy: ah, right, it is static now. I'd like to keep the Factory, GcpValidatior, AlwaysPassValidator, because that is a well-organized,  need to have thing in the future.

src/test/java/bio/terra/rbs/service/pool/ResourceConfigValidatorTest.java
yonghaoy: Done.

src/main/java/bio/terra/rbs/service/pool/AlwaysPassValidator.java
```diff
@@ -3,7 +3,7 @@
 import bio.terra.rbs.generated.model.ResourceConfig;
 
 /** The default validator which does not no restrictions. */
```
wchamber, line 5: nit: don't include 'default' in the comment

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -245,10 +245,12 @@ private void assertEnableApisContains(Project project, List<String> enabledApis)
 
   private void assertIamBindingsContains(Project project, List<IamBinding> iamBindings)
       throws Exception {
-    // By default we enable some services, and some GCP automatically creates SA accounts and grant
-    // permission.
+    // By default we enable some services, which causes GCP to automatically create Service Accounts
+    // and grant them
+    // permissions on the project.
```
wchamber, line 250: nit: comment whitespacing

wchamber (approved):

PR #31, opened 2020-10-21
[PF-163] Basic network creation Stairway step

yonghaoy:


yonghaoy:
@wchamber

src/main/java/bio/terra/rbs/service/resource/flight/CreateNetworkStep.java
```diff
@@ -0,0 +1,56 @@
...
+import bio.terra.stairway.StepStatus;
+import bio.terra.stairway.exception.RetryException;
+import com.google.api.services.compute.model.Network;
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Craetes Network project */
```
gjuggler, line 21: nit: this should probably be something like "Creates a VPC network within a GCP project created in a prior step."

src/main/java/bio/terra/rbs/service/resource/flight/CreateNetworkStep.java
```diff
@@ -0,0 +1,56 @@
...
+    } catch (IOException | InterruptedException e) {
+      logger.info("Error when creating network", e);
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    return StepResult.getStepResultSuccess();
```
gjuggler, line 54: I'm not sure on general Stairway style, but I feel like it's worth adding a line comment explaining why we don't need to undo anything. Unless this is very common / more often than not the case.

gjuggler (approved):

src/main/java/bio/terra/rbs/service/resource/flight/CreateNetworkStep.java
yonghaoy: Done

src/main/java/bio/terra/rbs/service/resource/flight/CreateNetworkStep.java
```diff
@@ -0,0 +1,57 @@
...
+  @Override
+  public StepResult doStep(FlightContext flightContext) throws RetryException {
+    String projectId = flightContext.getWorkingMap().get(GOOGLE_PROJECT_ID, String.class);
+    Network network = new Network().setName(NETWORK_NAME).setAutoCreateSubnetworks(false);
+    try {
+      OperationCow<?> operation =
+          computeCow
+              .globalOperations()
+              .operationCow(projectId, computeCow.networks().insert(projectId, network).execute());
+      pollUntilSuccess(operation, Duration.ofSeconds(5), Duration.ofMinutes(5));
```
wchamber, line 44: what will happen if this is retried and previously succeeded? Do we need to check if the network already exists, or will this succeed?

yonghaoy: ah, right, creating network usually takes < 60s in my previous observation from CRL tests. So I think 5 minutes should be enough for us to poll a success/fail result unless GCP outrage.. . 
Also, I didn't add retry rule for those steps yet. 

Some options we can do to have better protection if we want to rety stuff: 
GetNetwork and catch 404 notfound? 
CreateNetwork and catch 409 conflict. 

And we may need all similar things for all other steps. 
I feel starting from non-retry and let stairway rollback is safer... 
Wdyt?

wchamber: I'm inclined to start with better protection. I think retry rules for all the external calls (i.e. GCP) probably makes sense (not this PR). We expect GCP to fail occasionally, let's handle it. It'd be a shame to throw away the project because the network api call fails.
You could test for already-created-network by running this step twice.

wchamber: But you could also punt until the next PR.

wchamber (approved):

src/main/java/bio/terra/rbs/service/resource/flight/CreateNetworkStep.java
yonghaoy: added a better protection. But I still think 5 minutes polling time is long enough, and there will be no left-over pending creating networks. During retry. If that happens, that probably would be a GCP outrage, and I don't think our protection here would help as well..
But it might be helpful in some cases when GCP works fine just really really slow.  Not sure if that worth the effort to add those extra logics to handle those wired case.
Throwing away the project because the network api call fails in 0.01% chance sounds not that shameful through... 

Anyway, added the protection with a test case, couldn't hurt

PR #34, opened 2020-10-23
[PF-163] Create SubNetwork

yonghaoy:
This should be equal to:
https://github.com/broadinstitute/gcp-dm-templates/blob/eeee90fa4619b273d07206a867b01914cdeb0a30/firecloud_project.py#L112

### DIFF
--- a/src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
@@ -15,11 +15,10 @@ import bio.terra.stairway.exception.RetryException;
 import com.google.api.services.compute.model.Network;
 import com.google.api.services.compute.model.Subnetwork;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import org.slf4j.Logger;
@@ -44,36 +43,38 @@ import org.slf4j.LoggerFactory;
  */
 public class CreateSubnetsStep implements Step {
   @VisibleForTesting public static final String SUBNETWORK_NAME = "subnetwork";
-  /** All current Google Compute Engine regions, value from: {@code gcloud compute regions list}. */
+  /**
+   * All current Google Compute Engine regions with the default Ip ranges listed (and manually
+   * copied) in: https://cloud.google.com/vpc/docs/vpc#ip-ranges.
+   */
   @VisibleForTesting
-  public static final List<String> SUBNET_REGIONS =
-      ImmutableList.of(
-          "asia-east1",
-          "asia-east2",
-          "asia-northeast1",
-          "asia-northeast2",
-          "asia-northeast3",
-          "asia-south1",
-          "asia-southeast1",
-          "asia-southeast2",
-          "australia-southeast1",
-          "europe-north1",
-          "europe-west1",
-          "europe-west2",
-          "europe-west3",
-          "europe-west4",
-          "europe-west6",
-          "northamerica-northeast1",
-          "southamerica-east1",
-          "us-central1",
-          "us-east1",
-          "us-east4",
-          "us-west1",
-          "us-west2",
-          "us-west3",
-          "us-west4");
-
-  public static final Map<String, String> REGION_TO_IP_RANGE = new HashMap<>();
+  public static final Map<String, String> REGION_TO_IP_RANGE =
+      ImmutableMap.<String, String>builder()
+          .put("asia-east1", "10.140.0.0/20")
+          .put("asia-east2", "10.170.0.0/20")
+          .put("asia-northeast1", "10.146.0.0/20")
+          .put("asia-northeast2", "10.174.0.0/20")
+          .put("asia-northeast3", "10.178.0.0/20")
+          .put("asia-south1", "10.160.0.0/20")
+          .put("asia-southeast1", "10.148.0.0/20")
+          .put("asia-southeast2", "10.184.0.0/20")
+          .put("australia-southeast1", "10.152.0.0/20")
+          .put("europe-north1", "10.166.0.0/20")
+          .put("europe-west1", "10.132.0.0/20")
+          .put("europe-west2", "10.154.0.0/20")
+          .put("europe-west3", "10.156.0.0/20")
+          .put("europe-west4", "10.164.0.0/20")
+          .put("europe-west6", "10.172.0.0/20")
+          .put("northamerica-northeast1", "10.162.0.0/20")
+          .put("southamerica-east1", "10.158.0.0/20")
+          .put("us-central1", "10.128.0.0/20")
+          .put("us-east1", "10.142.0.0/20")
+          .put("us-east4", "10.150.0.0/20")
+          .put("us-west1", "10.138.0.0/20")
+          .put("us-west2", "10.168.0.0/20")
+          .put("us-west3", "10.180.0.0/20")
+          .put("us-west4", "10.182.0.0/20")
+          .build();
 
   private final Logger logger = LoggerFactory.getLogger(CreateSubnetsStep.class);
   private final CloudComputeCow computeCow;
@@ -84,18 +85,6 @@ public class CreateSubnetsStep implements Step {
     this.gcpProjectConfig = gcpProjectConfig;
   }
 
-  /**
-   * Generates Ip ranges programmatically.
-   *
-   * @see <a
-   *     href="https://github.com/broadinstitute/gcp-dm-templates/blob/eeee90fa4619b273d07206a867b01914cdeb0a30/firecloud_project.py#L49">gcp-dm-templates</a>
-   */
-  static {
-    for (int i = 0; i < SUBNET_REGIONS.size(); i++) {
-      REGION_TO_IP_RANGE.put(SUBNET_REGIONS.get(i), "10." + (128 + 2 * i) + ".0.0/20");
-    }
-  }
-
   @Override
   public StepResult doStep(FlightContext flightContext) throws RetryException {
     String projectId = flightContext.getWorkingMap().get(GOOGLE_PROJECT_ID, String.class);
@@ -103,7 +92,8 @@ public class CreateSubnetsStep implements Step {
     List<OperationCow<?>> operationsToPoll = new ArrayList<>();
     try {
       Network network = computeCow.networks().get(projectId, NETWORK_NAME).execute();
-      for (String region : SUBNET_REGIONS) {
+      for (Map.Entry<String, String> entry : REGION_TO_IP_RANGE.entrySet()) {
+        String region = entry.getKey();
         if (cloudObjectExists(
             () -> computeCow.subnetworks().get(projectId, region, SUBNETWORK_NAME).execute())) {
           continue;
@@ -113,7 +103,7 @@ public class CreateSubnetsStep implements Step {
                 .setName(SUBNETWORK_NAME)
                 .setRegion(region)
                 .setNetwork(network.getSelfLink())
-                .setIpCidrRange(REGION_TO_IP_RANGE.get(region))
+                .setIpCidrRange(entry.getValue())
                 .setEnableFlowLogs(networkMonitoringEnabled)
                 .setPrivateIpGoogleAccess(networkMonitoringEnabled);
         operationsToPoll.add(
--- a/src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
@@ -1,5 +1,10 @@
 package bio.terra.rbs.service.resource.flight;
 
+import static bio.terra.rbs.service.resource.FlightMapKeys.GOOGLE_PROJECT_ID;
+import static bio.terra.rbs.service.resource.flight.CreateNetworkStep.NETWORK_NAME;
+import static bio.terra.rbs.service.resource.flight.GoogleUtils.enableNetworkMonitoring;
+import static bio.terra.rbs.service.resource.flight.GoogleUtils.pollUntilSuccess;
+
 import bio.terra.cloudres.google.api.services.common.OperationCow;
 import bio.terra.cloudres.google.compute.CloudComputeCow;
 import bio.terra.rbs.generated.model.GcpProjectConfig;
@@ -8,55 +13,53 @@ import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
 import bio.terra.stairway.StepStatus;
 import bio.terra.stairway.exception.RetryException;
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.services.compute.model.Network;
 import com.google.api.services.compute.model.Subnetwork;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
-import java.time.Duration;
-import java.util.List;
-
-import static bio.terra.rbs.service.resource.FlightMapKeys.GOOGLE_PROJECT_ID;
-import static bio.terra.rbs.service.resource.flight.CreateNetworkStep.NETWORK_NAME;
-import static bio.terra.rbs.service.resource.flight.GoogleUtils.enableNetworkMonitoring;
-import static bio.terra.rbs.service.resource.flight.GoogleUtils.pollUntilSuccess;
-
 /** Craetes Subnetworks for project */
 public class CreateSubnetsStep implements Step {
+  @VisibleForTesting public static final String SUBNETWORK_NAME = "subnetwork";
+  /** All current Google Compute Engine regions, value from: {@code gcloud compute regions list}. */
+  @VisibleForTesting
+  public static final List<String> SUBNETS_REGIONS =
+      ImmutableList.of(
+          "asia-east1",
+          "asia-east2",
+          "asia-northeast1",
+          "asia-northeast2",
+          "asia-northeast3",
+          "asia-south1",
+          "asia-southeast1",
+          "asia-southeast2",
+          "australia-southeast1",
+          "europe-north1",
+          "europe-west1",
+          "europe-west2",
+          "europe-west3",
+          "europe-west4",
+          "europe-west6",
+          "northamerica-northeast1",
+          "southamerica-east1",
+          "us-central1",
+          "us-east1",
+          "us-east4",
+          "us-west1",
+          "us-west2",
+          "us-west3",
+          "us-west4");
+
   private final Logger logger = LoggerFactory.getLogger(CreateSubnetsStep.class);
   private final CloudComputeCow computeCow;
   private final GcpProjectConfig gcpProjectConfig;
-
-  private static final String SUBNET_NAME_PREFIX = "subnetwork_";
-  /** All current Google Compute Engine regions, value from: {@code gcloud compute regions list}. */
-  private static final List<String> SUBNETS_REGIONS = ImmutableList.of(
-  "asia-east1",
-  "asia-east2",
-  "asia-northeast1",
-  "asia-northeast2",
-  "asia-northeast3",
-  "asia-south1",
-  "asia-southeast1",
-  "asia-southeast2",
-  "australia-southeast1",
-  "europe-north1",
-  "europe-west1",
-  "europe-west2",
-  "europe-west3",
-  "europe-west4",
-  "europe-west6",
-  "northamerica-northeast1",
-  "southamerica-east1",
-  "us-central1",
-  "us-east1",
-  "us-east4",
-  "us-west1",
-  "us-west2",
-  "us-west3",
-  "us-west4");
 
   public CreateSubnetsStep(CloudComputeCow computeCow, GcpProjectConfig gcpProjectConfig) {
     this.computeCow = computeCow;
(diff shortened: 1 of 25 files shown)


src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -207,6 +246,35 @@ public MultiNetworkStepFlight(FlightMap inputParameters, Object applicationConte
...
+      addStep(new CreateResourceDbEntityStep(rbsDao));
+      addStep(new GenerateProjectIdStep());
+      addStep(new CreateProjectStep(rmCow, gcpProjectConfig));
+      addStep(new SetBillingInfoStep(billingCow, gcpProjectConfig));
+      addStep(new EnableServicesStep(serviceUsageCow, gcpProjectConfig));
+      addStep(new SetIamPolicyStep(rmCow, gcpProjectConfig));
+      addStep(new CreateNetworkStep(cloudComputeCow, gcpProjectConfig));
+      addStep(new CreateSubnetsStep(cloudComputeCow, gcpProjectConfig));
+      addStep(new CreateSubnetsStep(cloudComputeCow, gcpProjectConfig));
+      addStep(new FinishResourceCreationStep(rbsDao));
```
wchamber, line 274: this is not really maintainable. This is supposed to be the same sequence of steps as the normal creation flight, but it's just copy-pasted and nothing is checking that. Figure out how to make this stay in sync.

wchamber (changes requested):

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -0,0 +1,134 @@
...
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Craetes Subnetworks for project */
+public class CreateSubnetsStep implements Step {
```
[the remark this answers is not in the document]
yonghaoy: I think Google also calls it subnets: https://cloud.google.com/vpc/docs/vpc#vpc_networks_and_subnets

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /**
+   * Generates Ip ranges programmatically.
+   *
+   * @see <a
+   *     href="https://github.com/broadinstitute/gcp-dm-templates/blob/eeee90fa4619b273d07206a867b01914cdeb0a30/firecloud_project.py#L49">gcp-dm-templates</a>
+   */
+  @VisibleForTesting
+  public static String generateIpRange(int i) {
```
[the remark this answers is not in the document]
yonghaoy: My understanding: 
Some valid ipV4 ranges: https://cloud.google.com/vpc/docs/vpc#manually_created_subnet_ip_ranges
For us, it is from 10.128.0.0/20 to 10.176.0.0/20 for now. 
And seems that is the values in DM:
https://console.cloud.google.com/dm/deployments/details/dm-aou-rw-test-aa8646c7?project=terra-deployments-dev&folder&organizationId

@gjuggler for more information.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -137,6 +159,23 @@ public void testCreateGoogleProject_multipleNetworkCreation() throws Exception {
     assertNetworkExists(project);
   }
 
+  @Test
+  public void testCreateGoogleProject_multipleSubnetsCreation() throws Exception {
+    // Verify flight is able to finish successfully when subnets exists
```
gjuggler, line 164: I think you need to add a bit more explanation to this test case. IMO it's important to mention that this scenario may arise when the step partially fails and ends up in a state where some subnets need to be recreated and some are getting created the first time.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -317,6 +385,20 @@ private void assertNetworkExists(Project project) throws Exception {
     assertFalse(network.getAutoCreateSubnetworks());
   }
 
+  private void assertSubnetsExists(Project project, boolean networkMonitoringEnabled)
```
gjuggler, line 388: sp / grammar nit: s/exists/exist/

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Craetes Subnetworks for project */
+public class CreateSubnetsStep implements Step {
+  @VisibleForTesting public static final String SUBNETWORK_NAME = "subnetwork";
+  /** All current Google Compute Engine regions, value from: {@code gcloud compute regions list}. */
+  @VisibleForTesting
+  public static final List<String> SUBNETS_REGIONS =
```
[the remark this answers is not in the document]
gjuggler: See my suggested javadoc comments. tl;dr I agree it's not maintainable if we care about having all possible regions reflected here. However, the other extreme (only create a subnet in Terra's current "primary" region) is almost certainly less preferable in the medium term. So this is a balance, to create some future flexibility in resource regions while not going overboard on infrastructure yet (since we don't have a pressing product need).

For example: it's not hard to imagine some cron job which periodically crawls projects and adds missing subnet regions – if we end up needing something like that.

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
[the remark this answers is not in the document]
gjuggler: See the javadoc comment I suggest above – the scheme in the DM code was meant to replicate the empirical behavior of Google's default network and auto-subnet creation. That said, I implemented that awhile ago, so if Google's IP allocation has changed, maybe we need to think more on this.

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -0,0 +1,134 @@
...
+        operationsToPoll.add(
+            computeCow
+                .regionalOperations()
+                .operationCow(
+                    projectId,
+                    region,
+                    computeCow.subnetworks().insert(projectId, region, subnetwork).execute()));
+      }
+
+      // Batch poll to make it faster
```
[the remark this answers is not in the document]
yonghaoy: ah, right, change the comment to make it clearer

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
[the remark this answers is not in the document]
yonghaoy: Thanks!, Greg, a question(happy to discuss in person): 
If DM tried to replicate the behaviour of auto-subnet, I don't think it ever matches the default ip rage. https://cloud.google.com/vpc/docs/vpc#ip-ranges (region-ip pair).
And the list here even missed some regions, like asia-northeast3,  asia-southeast2, us-west3, us-west4.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: as long as we added CreateSubnetsStep() twice and test passed, we are good. It's necessary to be the same sequence of steps as the normal creation, we have those steps because it is the easiest way to **setup**. i.e., we need billing, compute api, and a network to create subnet. We can remove IAM policy because we don't need it.
I think the problem is code duplication, but I don't know what is the best way of decup the code from CreationStep. Any suggestion?

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -317,6 +385,20 @@ private void assertNetworkExists(Project project) throws Exception {
...
+      throws Exception {
+    Network network = computeCow.networks().get(project.getProjectId(), NETWORK_NAME).execute();
```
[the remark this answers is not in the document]
yonghaoy: we need to verify the subnet selflink line below

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -317,6 +385,20 @@ private void assertNetworkExists(Project project) throws Exception {
...
+    for (int i = 0; i < SUBNETS_REGIONS.size(); i++) {
+      String region = SUBNETS_REGIONS.get(i);
+      Subnetwork subnetwork =
+          computeCow.subnetworks().get(project.getProjectId(), region, SUBNETWORK_NAME).execute();
+      assertEquals(network.getSelfLink(), subnetwork.getNetwork());
+      assertEquals(generateIpRange(i), subnetwork.getIpCidrRange());
```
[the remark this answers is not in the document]
yonghaoy: Changed to use a Map<Region, Ip> now. PTAL

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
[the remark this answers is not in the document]
yonghaoy: I don't want to add two more functions like 
assertNonMornitorSubnet()
assertMornitorSubnet()
assertSubnetInternal()
The test files is getting more complicated now.   I'd like to use boolean with  inlined /**comment*/ in callsiet ...

But I changed to pass Network config instead. Does that LG?

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -0,0 +1,134 @@
...
+        String region = SUBNETS_REGIONS.get(i);
+        try {
+          // Skip this steps if subnet already exists.
+          computeCow.subnetworks().get(projectId, region, SUBNETWORK_NAME).execute();
+          logger.info("Subnets already exists for project {}, region: {}.", projectId, region);
+          continue;
+        } catch (IOException e) {
+          if (e instanceof GoogleJsonResponseException
+              && ((GoogleJsonResponseException) e).getStatusCode() == 404) {
+            // do nothing
```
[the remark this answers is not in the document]
yonghaoy: the part is gone now

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -0,0 +1,134 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Craetes Subnetworks for project */
```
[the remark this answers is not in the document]
yonghaoy: Thanks!

yonghaoy (commented):
Thanks! Comments are addressed, PTAL!

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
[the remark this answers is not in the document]
gjuggler: I clearly didn't look hard enough for that list when I first set up that scheme :) Let's anchor to the actual auto-mode IP ranges. Thanks Will for questioning this.

The missed regions were likely all added since the DM work first began, e.g. us-west3 and us-west4 seem to have been launched in 2020. This highlights the fundamental "flaw" in this approach, but it's a price we're paying to align with CIS best practices, which was a hard requirement for AoU to adopt Terra.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
gjuggler: Here's my cheeky attempt. Seems like it compiles, but I couldn't actually run integration tests to check it works. (I'm getting a 403 trying to pull secrets from Vault. I'll need to chase that down another time.)

```
  public static class MultiSubnetsStepFlight extends GoogleProjectCreationFlight {

    public MultiSubnetsStepFlight(FlightMap inputParameters, Object applicationContext) {
      super(inputParameters, applicationContext);
    }
    
    @Override
    protected void addStep(Step step) {
      super.addStep(step);
      if (step instanceof CreateSubnetsStep) {
        // Create a duplicate entry for any CreateSubnetsStep in the original flight path.
        super.addStep(step);
      }
    }
  }
```

It's obviously not a very flexible way to edit the step list for a pre-existing flight, though. Adding a  `flight.insertStepForTestingOnly(index, step)` method to Stairway might be more generally suitable. Or otherwise some Stairway-level control mechanism which simulates partial step failure and retry, though that would probably require some deeper thought.

gjuggler (approved):
LGTM, but I'd treat Will's approval as the authoritative one here :)

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: Thanks!

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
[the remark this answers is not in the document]
yonghaoy: Changed to use region - ip range map. Unfortunately, I have to manually copy them over..

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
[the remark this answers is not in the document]
wchamber: No, passing the whole Network config makes the test more difficult to understand - it could do anything with the network config, why does it need the whole thing.
If the extra functions seem too heavy weight, how about an enum? 
```
enum NetworkMonitoring { ENABLED; DISABLED }
...
assertSubnetExists(project, NetworkMonitoring.ENABLED);
```
If the enum is also too much, let's at least do argument comments, `assertSubnetExists(project, /*networkMonitoringEnabled= */ true);
https://goto.google.com/java-practices/boolean-parameters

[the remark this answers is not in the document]
yonghaoy: done

src/main/java/bio/terra/rbs/service/resource/flight/GoogleUtils.java
```diff
@@ -29,4 +35,30 @@ public static void pollUntilSuccess(
...
+
+  /** Checks if network monitoring is enabled from config. */
+  public static boolean checkEnableNetworkMonitoring(GcpProjectConfig gcpProjectConfig) {
+    return gcpProjectConfig.getNetwork() != null
+        && gcpProjectConfig.getNetwork().isEnableNetworkMonitoring() != null
+        && gcpProjectConfig.getNetwork().isEnableNetworkMonitoring();
+  }
+
+  /** Checks if cloudObject already exists. */
+  public static boolean cloudObjectExists(CowExecute execute) throws IOException {
```
wchamber, line 47: Nice idea. Few suggestions.

The function name seems a little redundant given this is under `GoogleUtils`. Is `objectExists()` going too far? (I also kind of want to call them resources, `resourceExists()`) WDYT? Not sure what's better.

Let's add a little to the javadoc.
```
<p>Many Google 'get' operations return a {@link GoogleJsonResponseException} 404. This function handles that exception case as false, while throwing any other exception.
```

src/main/java/bio/terra/rbs/service/resource/flight/GoogleUtils.java
```diff
@@ -29,4 +35,30 @@ public static void pollUntilSuccess(
...
+          && ((GoogleJsonResponseException) e).getStatusCode() == 404) {
+        return false;
+      } else {
+        throw e;
+      }
+    }
+  }
+
+  public interface CowExecute {
+    void execute() throws IOException;
```
wchamber, line 62: This needs javadoc. Add `@FunctionalInterface` for as well.
I'm not convinced `Cow` is super relevant here, but maybe some Javadoc will make it seem better.

src/main/java/bio/terra/rbs/service/resource/flight/GoogleUtils.java
```diff
@@ -29,4 +35,30 @@ public static void pollUntilSuccess(
...
+    return gcpProjectConfig.getNetwork() != null
+        && gcpProjectConfig.getNetwork().isEnableNetworkMonitoring() != null
+        && gcpProjectConfig.getNetwork().isEnableNetworkMonitoring();
+  }
+
+  /** Checks if cloudObject already exists. */
+  public static boolean cloudObjectExists(CowExecute execute) throws IOException {
+    try {
+      execute.execute();
+      return true;
```
wchamber, line 50: should we return an Optional<R> instead of boolean? Then we can re-use this where we get project check for existence.
I guess we could have two helper functions, and have the boolean one `return isPresent` the Optional one. Seems like a common pattern. WDYT?

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -0,0 +1,135 @@
...
+        operationsToPoll.add(
+            computeCow
+                .regionalOperations()
+                .operationCow(
+                    projectId,
+                    region,
+                    computeCow.subnetworks().insert(projectId, region, subnetwork).execute()));
+      }
+
+      // Kicking off all the operations first then polling all operations.
```
wchamber, line 118: nit: `then poll all operations`

src/main/java/bio/terra/rbs/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -0,0 +1,135 @@
...
+      logger.info("Error when creating subnets", e);
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    // Flight undo will just need to delete the project on GCP at CreateProjectStep.
+    // doStep methods already checks Sunets exists or not. So no need to delete subnet.
```
wchamber, line 132: typo `Sunets`

src/main/java/bio/terra/rbs/service/resource/flight/GoogleUtils.java
```diff
@@ -29,4 +35,30 @@ public static void pollUntilSuccess(
   public static String projectIdToName(String projectId) {
     return "projects/" + projectId;
   }
+
+  /** Checks if network monitoring is enabled from config. */
+  public static boolean checkEnableNetworkMonitoring(GcpProjectConfig gcpProjectConfig) {
+    return gcpProjectConfig.getNetwork() != null
+        && gcpProjectConfig.getNetwork().isEnableNetworkMonitoring() != null
+        && gcpProjectConfig.getNetwork().isEnableNetworkMonitoring();
```
wchamber, line 43: yay null values. When I see this nonsense, I kind of want to convert our Json config into an AutoValue. Don't do anything in this PR, I'm just bloviating.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -98,6 +101,25 @@ public void testCreateGoogleProject_witIamBindings() throws Exception {
     assertIamBindingsContains(project, iamBindings);
   }
 
+  @Test
+  public void testCreateGoogleProject_enableNetworkMonitoring() throws Exception {
+    // Basic GCP project with billing setup.
```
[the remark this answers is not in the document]
wchamber: +1

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -98,6 +106,25 @@ public void testCreateGoogleProject_witIamBindings() throws Exception {
...
+        preparePool(
+            newBasicGcpConfig()
+                .network(
+                    new bio.terra.rbs.generated.model.Network().enableNetworkMonitoring(true)));
+
+    String flightId = manager.submitCreationFlight(pool).get();
+    blockUntilFlightComplete(flightId);
+    Project project = assertProjectExists(pool);
+    assertBillingIs(project, pool.resourceConfig().getGcpProjectConfig().getBillingAccount());
+    assertEnableApisContains(project, pool.resourceConfig().getGcpProjectConfig().getEnabledApis());
```
wchamber, line 123: I think it's also ok to remove assertions not relevant to the test. We don't need to assert everything about every test. It's not likely that enabling network monitoring is going to disable billing or the apis. As-is, there's extra boilerplate that makes it harder to find out what this test actually cares about, e.g. the network/subnets with their monitoring.
In the basic test, asserting everything is as it should be is good, but we don't need to repeat that for every special case.
(ditto all other tests)

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -137,6 +164,26 @@ public void testCreateGoogleProject_multipleNetworkCreation() throws Exception {
     assertNetworkExists(project);
   }
 
+  @Test
+  public void testCreateGoogleProject_multipleSubnetsCreation() throws Exception {
+    // Verify flight is able to finish successfully when subnets already exists/
+    // this scenario may arise when the step partially fails and ends up in a state where some
+    // subnets need to be
+    // recreated and some are getting created the first time.
```
wchamber, line 172: nit: whitespace

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -180,30 +227,38 @@ public ErrorCreateProjectFlight(FlightMap inputParameters, Object applicationCon
...
+        super.addStep(step);
+      }
+    }
+  }
+
+  /**
+   * A {@link Flight} that has multiple subnets creation steps. So a flight will try to create all
+   * Subnets twice and still success.
+   */
+  public static class MultiSubnetsStepFlight extends GoogleProjectCreationFlight {
```
wchamber, line 249: looks like a general pattern that's useful for testing. Can we parameterize it so that we only need 1 class and not many?
```
public static class MultiInstanceStep<T extends Step> extends GoogleProjectCreationFlight {
...
  if (step instanceof T)
...
```
bonus points if you can also parameterize/generalize for other flights and not just subclass `GoogleProjectCreationFlight`.

wchamber (changes requested):
Last touches. I think the comment additions and REGION_TO_IP_RANGE map are really good changes.

src/main/java/bio/terra/rbs/service/resource/flight/GoogleUtils.java
yonghaoy: added a javadoc. Stilling using CowExecute because I am not sure what would be a better name. maybe `CloudExecute`?

src/main/java/bio/terra/rbs/service/resource/flight/GoogleUtils.java
yonghaoy: I don't think we can use it for getting project because that is 403.
For other things, I use the new Optional function now

src/main/java/bio/terra/rbs/service/resource/flight/GoogleUtils.java
yonghaoy: That would interesting... This config is defined/generated by openapi and parsed by Jackson... Now sure when/where we can plug autoValue in

gjuggler: We struggled with this in AoU as well. All of the swagger-generated classes are simply defined with template code; in theory it should be possible to improve those templates to have better / more fluent handling of nulls like this, but that involves reaching pretty deep into the internals of those templates, so we never bit the bullet on it.

One example of where we did end up doing some template-hacking was to add universal tracing to our server-side OpenAPI classes ([code](https://github.com/all-of-us/workbench/blob/master/api/src/main/resources/swagger3codegen/libraries/okhttp-gson/api.mustache)). Unfortunately it's hard to do this in a nice, maintainable way.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: instanceOf generic type is not allowed. Not sure how we can do this. One possible solution is  addeding a static method in the class something like setClazz(), and each class call it. For other flights, the ErrorCreateProjectFlight inserts Latch and an error steps, not sure how much we will need to modify to make the class more generic to support this

src/main/java/bio/terra/rbs/service/resource/flight/GoogleUtils.java
wchamber: hmmm, consider whether we want to elevate the 403/404 to the function name or as an input parameter.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
wchamber: Last try for multi -instance
```
  private final Class<? extends Step> multiStepClazz;
  public MultiInstanceStepFlight(Class<? extends Step> multiStepClazz) {...}
...
  if (multiStepClazz.isInstance(step))
...
```
I think for the other flights with latches and steps it's worth at least extending GoogleProjectCreationFlight and adding additional steps (all not in this PR please). Consider whether there are other creative solutions like this for overriding `addStep`, or what support for this we could add to the Stairway library.

src/main/java/bio/terra/rbs/service/resource/flight/GoogleUtils.java
```diff
@@ -43,22 +44,32 @@ public static boolean checkEnableNetworkMonitoring(GcpProjectConfig gcpProjectCo
...
         throw e;
       }
     }
   }
 
-  public interface CowExecute {
-    void execute() throws IOException;
+  /** Wrappers how to process a CRL's Cow execution. */
+  @FunctionalInterface
+  public interface CowExecute<R> {
```
wchamber, line 72: I like your `CloudExecute` suggestion. Maybe, `/** A Google cloud operation that's expected to throw a IOException. See {@link GoogleUtils#resourceExists}. */ `

src/main/java/bio/terra/rbs/service/resource/flight/GoogleUtils.java
```diff
@@ -43,22 +44,32 @@ public static boolean checkEnableNetworkMonitoring(GcpProjectConfig gcpProjectCo
...
+   * Checks if cloudObject already exists.
+   *
+   * <p>Many Google 'get' operations return a {@link GoogleJsonResponseException} 404. This function
+   * handles that exception case as false, while throwing any other exception.
+   */
+  public static <R> boolean resourceExists(CowExecute<R> execute) throws IOException {
+    return getResource(execute).isPresent();
+  }
+
+  public static <R> Optional<R> getResource(CowExecute<R> execute) throws IOException {
```
wchamber, line 57: add javadoc for this as well. Ok to have `resourceExists` reference this or vice versa to avoid repeating the same documentation.

wchamber (approved):

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: I don't think way gona work as well. Flights are not instantiated in RBS. All RBS did is tell Stairway the Flight class and inputMap.  And for latches and all stuff, that is only used in one test, and we can further if we have more tests which needs similar setup... For now, I don't think there will be a new test which needs that in the near future.

PR #35, opened 2020-10-26
[PF-163] Creates static custom route when network monitoring enabled

yonghaoy:
https://github.com/broadinstitute/gcp-dm-templates/blob/eeee90fa4619b273d07206a867b01914cdeb0a30/templates/network.py#L79

### DIFF
--- a/src/main/java/bio/terra/rbs/service/resource/flight/CreateRouteStep.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/CreateRouteStep.java
@@ -46,18 +46,19 @@ public class CreateRouteStep implements Step {
 
   @Override
   public StepResult doStep(FlightContext flightContext) throws RetryException {
-    if (!checkEnableNetworkMonitoring(gcpProjectConfig)) {
+    if (!isNetworkMonitoringEnabled(gcpProjectConfig)) {
       return StepResult.getStepResultSuccess();
     }
     String projectId = flightContext.getWorkingMap().get(GOOGLE_PROJECT_ID, String.class);
     try {
-      Network network =
-          getResource(() -> computeCow.networks().get(projectId, NETWORK_NAME).execute(), 404)
-              .get();
-
       if (resourceExists(() -> computeCow.routes().get(projectId, ROUTE_NAME).execute(), 404)) {
         return StepResult.getStepResultSuccess();
       }
+      // Network is already created and checked in previous step so here won't be empty.
+      // If we got NPE, that means something went wrong with GCP, fine to just throw NPE here.
+      Network network =
+          getResource(() -> computeCow.networks().get(projectId, NETWORK_NAME).execute(), 404)
+              .get();
       Route route =
           new Route()
               .setName(ROUTE_NAME)
--- a/src/main/java/bio/terra/rbs/service/resource/flight/CreateRouteStep.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/CreateRouteStep.java
@@ -1,0 +1,86 @@
+package bio.terra.rbs.service.resource.flight;
+
+import static bio.terra.rbs.service.resource.FlightMapKeys.GOOGLE_PROJECT_ID;
+import static bio.terra.rbs.service.resource.flight.GoogleUtils.*;
+
+import bio.terra.cloudres.google.api.services.common.OperationCow;
+import bio.terra.cloudres.google.compute.CloudComputeCow;
+import bio.terra.rbs.generated.model.GcpProjectConfig;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import bio.terra.stairway.exception.RetryException;
+import com.google.api.services.compute.model.Network;
+import com.google.api.services.compute.model.Route;
+import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/rbs/service/resource/flight/CreateRouteStep.java
```diff
@@ -0,0 +1,86 @@
...
+  private final GcpProjectConfig gcpProjectConfig;
+
+  public CreateRouteStep(CloudComputeCow computeCow, GcpProjectConfig gcpProjectConfig) {
+    this.computeCow = computeCow;
+    this.gcpProjectConfig = gcpProjectConfig;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) throws RetryException {
+    if (!checkEnableNetworkMonitoring(gcpProjectConfig)) {
```
wchamber, line 49: missed this last time around, but `check` to me implies a little more than a boolean evaluation. How about `isNetworkMonitoringEnabled`?

src/main/java/bio/terra/rbs/service/resource/flight/CreateRouteStep.java
```diff
@@ -0,0 +1,86 @@
...
+      return StepResult.getStepResultSuccess();
+    }
+    String projectId = flightContext.getWorkingMap().get(GOOGLE_PROJECT_ID, String.class);
+    try {
+      Network network =
+          getResource(() -> computeCow.networks().get(projectId, NETWORK_NAME).execute(), 404)
+              .get();
```
wchamber, line 56: this is doing an unqualified .get() on an optional. worth an explanation on why that's ok and the optional can't be empty.

src/main/java/bio/terra/rbs/service/resource/flight/CreateRouteStep.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) throws RetryException {
+    if (!checkEnableNetworkMonitoring(gcpProjectConfig)) {
+      return StepResult.getStepResultSuccess();
+    }
+    String projectId = flightContext.getWorkingMap().get(GOOGLE_PROJECT_ID, String.class);
+    try {
+      Network network =
```
wchamber, line 54: nit, move this below the if routes exist check to be closer to its use in creating the route

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -119,6 +121,7 @@ public void testCreateGoogleProject_enableNetworkMonitoring() throws Exception {
     Project project = assertProjectExists(pool);
     assertNetworkExists(project);
     assertSubnetsExist(project, NetworkMonitoring.ENABLED);
+    assertRouteExist(project);
```
wchamber, line 124: probably also worth asserting the route doesn't exist in the base case.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -254,6 +257,22 @@ protected void addStep(Step step) {
     }
   }
 
+  /** A {@link Flight} that has multiple route creation steps. */
+  public static class MultiRouteStepFlight extends GoogleProjectCreationFlight {
+    public MultiRouteStepFlight(FlightMap inputParameters, Object applicationContext) {
```
wchamber, line 262: here's another one. what was wrong with my Class<? extends Step> suggestion?

wchamber (approved):
couple nits

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: Tried few ways but seems it is not working. 
I remeber the last thing I tried is 

```
  private final Class<? extends Step> multiStepClazz;
  public MultiInstanceStepFlight(Class<? extends Step> multiStepClazz) {...}
...
  if (multiStepClazz.isInstance(step))
...
```

But like I said  https://github.com/DataBiosphere/terra-rbs/pull/34#discussion_r512176996, Flight is not instinciated in RBS, so I don't know where to passed parameters then... Unless one proposal I asked in that comment threads is to have some static setter like: 
```

    private static Class<? extends Step> multiStepClazz;
    public MultiInstanceStepFlight(FlightMap inputParameters, Object applicationContext) {
      super(inputParameters, applicationContext);
    }
    
    public static void setStepClass(Class<? extends Step>  clazz) {
      multiStepClazz = clazz;
    }
    @Override
    protected void addStep(Step step) {
      super.addStep(step);
      if (multiStepClazz.isInstance(step)) {
        super.addStep(step);
      }
    }
  }
```

DO you think that is okay or is there any other better solutions?

yonghaoy: @wchamber

wchamber: oh, got it. It's kind of awkward Stairway really wants to use constructors here. Registering creation functions would be nicer.
We could put the Class in the `FlightMap inputParameters`.

wchamber (approved):

PR #36, opened 2020-10-26
[PF-143] updateResourceAsDeleted and updateResourceAsDeleting

yonghaoy:
Will be used in delete resource flight


src/test/java/bio/terra/rbs/db/RbsDaoTest.java
```diff
@@ -195,4 +195,26 @@ public void updateResourceAsHandedOut() {
...
+
+  @Test
+  public void updateResourceAsDeleted() {
+    Pool pool = newPool(PoolId.create("poolId"));
+    Resource resource = newResource(pool.id(), ResourceState.DELETING);
+    rbsDao.createPools(ImmutableList.of(pool));
+    rbsDao.createResource(resource);
+
+    rbsDao.updateResourceAsDeleted(resource.id());
+    assertEquals(ResourceState.DELETED, rbsDao.retrieveResource(resource.id()).get().state());
```
wchamber, line 218: verify something about the deletion time as well?

wchamber (commented):
I'd kind of like to see the tests pass before approving/reviewing.

yonghaoy:
@wchamber Test shows passed. And RBS test takes ~40 mintes is normal now. Each resource creation flight takes 2~5 minutes.
And totally we have 10+ resources to create in test

src/test/java/bio/terra/rbs/db/RbsDaoTest.java
yonghaoy: Timestamp `now` is calculated in Dao, and we can not get the exact time when we call this method. Unless I can use some 3rd party library which can mock static method..
So I think it is not worth to check that, and just skip it

wchamber: You can verify that a value is set when you retrieve the resource. You should probably at least check it's close to now, or after Instant.now() right before the test.
The fact that it's hard to test is a bit of a signal about the api design. You could make updating the time a separate method, or make it an argument to this method.

wchamber:
> @wchamber Test shows passed. And RBS test takes ~40 mintes is about right.
> Each resource creation flight takes 2 to 5 minutes, and we have 10+ resources to create in total. That is really annoying, I am trying various ways to reduce the time now, not none of them can reduce them significationly.

Yeah, definitely worth finding a speedup. Can we run the tests in parallel? Can we change some of the tests so that they skip the slow part of project creation & just re-use a project to set up networks? Worth exploring more, maybe ticket worthy.

wchamber (approved):

PR #37, opened 2020-10-26
[PF-163] Customize log config

yonghaoy:
https://github.com/broadinstitute/gcp-dm-templates/blob/master/templates/subnetwork.py#L48

This includes some changes in: https://github.com/DataBiosphere/terra-rbs/pull/35


yonghaoy:
@wchamber FYI, I don't think there is anything you would be interested in.

gjuggler (approved):

PR #38, opened 2020-10-29
[PF-202] Create DNS and record set

yonghaoy:


yonghaoy:
Hi @gjuggler I was trying to follow https://cloud.google.com/vpc-service-controls/docs/set-up-private-connectivity#configuring-dns
and https://github.com/broadinstitute/gcp-dm-templates/blob/eeee90fa4619b273d07206a867b01914cdeb0a30/templates/private_google_access_dns_zone.py

Not sure if those variables are correct. PTAL, thanks!

@wchamber FYI

gjuggler (approved):
All looks good to me – I read through and compared to the gcp-dm code, all looks in order.

PR #39, opened 2020-10-29
[PF-143] delete resource flight

yonghaoy:
Hi Will, sorry for another large PR. But I think the majority line diffs are moving common things into `IntegratinUtils` then Creation/Deletion integration test can both use.  Let me know if I should split this into small PRs

### DIFF
--- a/src/main/java/bio/terra/rbs/service/resource/flight/DeleteProjectStep.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/DeleteProjectStep.java
@@ -39,8 +39,7 @@ public class DeleteProjectStep implements Step {
           || project.get().getLifecycleState().equals("DELETE_REQUESTED")
           || project.get().getLifecycleState().equals("DELETE_IN_PROGRESS")) {
         // Skip is project does not exist, is deleted or being deleted. We know that the project is
-        // created by RBS hence
-        // RBS should have owner permission. So we assume 403 in this case means that the project
+        // created by RBS hence RBS should have owner permission. So we assume 403 in this case means that the project
         // does not exist.
         logger.info("Project id: {} is deleted or being deleted", projectId);
         return StepResult.getStepResultSuccess();
--- a/src/main/java/bio/terra/rbs/service/resource/flight/DeleteProjectStep.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/DeleteProjectStep.java
@@ -36,9 +36,8 @@ public class DeleteProjectStep implements Step {
     try {
       Optional<Project> project = retrieveProject(rmCow, projectId);
       if (!project.isPresent()
-          || project.get().getLifecycleState().equals("DELETE_REQUESTED")
-          || project.get().getLifecycleState().equals("DELETE_IN_PROGRESS")) {
-        // Skip is project does not exist, is deleted or being deleted. We know that the project is
+          || isProjectDeleting(project.get())) {
+        // Skip if project does not exist, or is being deleted. We know that the project is
         // created by RBS hence RBS should have owner permission. So we assume 403 in this case
         // means that the project does not exist.
         logger.info("Project id: {} is deleted or being deleted", projectId);
@@ -54,7 +53,7 @@ public class DeleteProjectStep implements Step {
 
   @Override
   public StepResult undoStep(FlightContext flightContext) {
-    // Noting to revert.
-    return StepResult.getStepResultSuccess();
+    // We can't undo deleting a project
+    return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL);
   }
 }
--- a/src/main/java/bio/terra/rbs/service/resource/flight/DeleteProjectStep.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/DeleteProjectStep.java
@@ -35,8 +35,7 @@ public class DeleteProjectStep implements Step {
             .getProjectId();
     try {
       Optional<Project> project = retrieveProject(rmCow, projectId);
-      if (!project.isPresent()
-          || isProjectDeleting(project.get())) {
+      if (!project.isPresent() || isProjectDeleting(project.get())) {
         // Skip if project does not exist, or is being deleted. We know that the project is
         // created by RBS hence RBS should have owner permission. So we assume 403 in this case
         // means that the project does not exist.
--- a/src/main/java/bio/terra/rbs/service/resource/flight/DeleteProjectStep.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/DeleteProjectStep.java
@@ -1,0 +1,61 @@
+package bio.terra.rbs.service.resource.flight;
+
+import static bio.terra.rbs.service.resource.FlightMapKeys.CLOUD_RESOURCE_UID;
+import static bio.terra.rbs.service.resource.flight.GoogleUtils.*;
+
+import bio.terra.cloudres.google.cloudresourcemanager.CloudResourceManagerCow;
+import bio.terra.rbs.generated.model.CloudResourceUid;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import bio.terra.stairway.exception.RetryException;
+import com.google.api.services.cloudresourcemanager.model.Project;
+import java.io.IOException;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Deletes the GCP project. */
+public class DeleteProjectStep implements Step {
+  private final Logger logger = LoggerFactory.getLogger(DeleteProjectStep.class);
+  private final CloudResourceManagerCow rmCow;
+
+  public DeleteProjectStep(CloudResourceManagerCow rmCow) {
+    this.rmCow = rmCow;
+  }
+
+  @Override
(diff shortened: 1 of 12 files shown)


src/main/java/bio/terra/rbs/service/resource/flight/InitialResourceDeletionStep.java
```diff
@@ -0,0 +1,47 @@
...
+  public StepResult undoStep(FlightContext flightContext) {
+    FlightMap inputMap = flightContext.getInputParameters();
+    // Revert resource state back to READY to let RBS scheduling service pick it up again.
+    // Currently we don't have FATAL state to prevent a bad state resource get into here again and
+    // again. We rely on alerting + scripts to prevent this. At this moment, the deletion logic is
+    // simple and it should work for most cases unless GCP outrage happens in which case there is
+    // nothing we can do.
+    if (rbsDao.updateResourceAsReady(
+        ResourceId.retrieve(inputMap),
+        inputMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class))) {
```
wchamber, line 41: Updating the resource state in the same rbsDao function is awkward here. Consider splitting modify the resource state from other changes to resources. We never want this to change the cloud resource id, but our rbsDao API forces that to be possible.

src/test/java/bio/terra/rbs/integration/DeleteProjectFlightIntegrationTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Test
+  public void testDeleteGoogleProject_success() throws Exception {
+    FlightManager manager = new FlightManager(flightSubmissionFactoryImpl, stairwayComponent);
+    Pool pool =
+        preparePool(
+            rbsDao,
+            newBasicGcpConfig()
+                .network(
+                    new bio.terra.rbs.generated.model.Network().enableNetworkMonitoring(true)));
```
wchamber, line 39: why does this test need to mess with the network config?

src/test/java/bio/terra/rbs/integration/DeleteProjectFlightIntegrationTest.java
```diff
@@ -0,0 +1,110 @@
...
+    assertEquals(ResourceState.READY, rbsDao.retrieveResource(resource.id()).get().state());
+
+    // Now try to submit the flight again, expect flight success and resource state is DELETED.
+    deleteFlightId = manager.submitDeletionFlight(resource, ResourceType.GOOGLE_PROJECT).get();
+    blockUntilFlightComplete(stairwayComponent, deleteFlightId);
+    assertEquals(ResourceState.DELETED, rbsDao.retrieveResource(resource.id()).get().state());
+  }
+
+  private Project assertProjectExists(Pool pool) throws Exception {
+    Resource resource = rbsDao.retrieveResources(pool.id(), ResourceState.READY, 1).get(0);
```
wchamber, line 86: this is kind of funny, we do the same retrieve resources after this call at the call site. Why not have this take the resource as the argument instead of the pool?

I think we can also get around the awkward retrieving exactly one ready resource from the pool id by looking at the Stairway FlightState#getResultMap. The result map should have the resource id that was stored during flight execution.
Maybe change that for this test and the CreateProjectFlightIntegrationTest.

wchamber (changes requested):

src/main/java/bio/terra/rbs/service/resource/flight/DeleteProjectStep.java
```diff
@@ -0,0 +1,60 @@
...
+      logger.info("Error when deleting GCP project", e);
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    // Noting to revert.
+    return StepResult.getStepResultSuccess();
```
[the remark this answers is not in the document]
yonghaoy: Changed to fatal. 
But I feel that is not just _we can not undo deleting_, it is also we _don't need to undo deleting_.

src/main/java/bio/terra/rbs/service/resource/flight/InitialResourceDeletionStep.java
```diff
@@ -0,0 +1,47 @@
...
+
+  public InitialResourceDeletionStep(RbsDao rbsDao) {
+    this.rbsDao = rbsDao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) {
+    ResourceId resourceId = ResourceId.retrieve(flightContext.getInputParameters());
+    Optional<Resource> resource = rbsDao.retrieveResource(resourceId);
+    if (resource.isPresent() && resource.get().state().equals(ResourceState.READY)) {
```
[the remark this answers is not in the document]
yonghaoy: Changed to:  return fatal if resource is not found or state is not READY or DELETING. 
And for the question about different flight put it in deleting. Not sure how stariway works here. If there are two flights tried to cleanup the same resource. If we fail the slower one in the 1st step, seems fine to me. 
If we allow the slower one pass this steps, step2(DeleteProjectStep) should also work because we check if Project  is deleted or being deleted there. Step3(update db entity to DELETED) will have a small problem that we update the deletion timestamp again. But I don't think that a big deal. WDTY?

src/main/java/bio/terra/rbs/service/resource/flight/InitialResourceDeletionStep.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    FlightMap inputMap = flightContext.getInputParameters();
+    // Revert resource state back to READY to let RBS scheduling service pick it up again.
+    // Currently we don't have FATAL state to prevent a bad state resource get into here again and
+    // again. We rely on alerting + scripts to prevent this. At this moment, the deletion logic is
+    // simple and it should work for most cases unless GCP outrage happens in which case there is
+    // nothing we can do.
```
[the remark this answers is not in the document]
yonghaoy: Changed to Fatal now.

src/main/java/bio/terra/rbs/service/resource/flight/InitialResourceDeletionStep.java
yonghaoy: no longer needed

src/test/java/bio/terra/rbs/integration/DeleteProjectFlightIntegrationTest.java
yonghaoy: Done

src/test/java/bio/terra/rbs/integration/DeleteProjectFlightIntegrationTest.java
wchamber: Hmm, that's not a bad idea, but I don't think this test will stay up to date. If we add more features to what makes a "full" project, we won't remember to update this test. I don't know if project deletion of  a project that has custom networks and logging is more interesting. If it really is, make the intention that this is a "full' project more obvious by a) using a private static final variable or function named "full' with a comment explaining the intention. As-is, the code doesn't explain why it's doing what it's doing.

src/main/java/bio/terra/rbs/service/resource/flight/InitialResourceDeletionStep.java
```diff
@@ -24,24 +22,17 @@ public StepResult doStep(FlightContext flightContext) {
     if (resource.isPresent() && resource.get().state().equals(ResourceState.READY)) {
       // Only update READY state resource.
       rbsDao.updateResourceAsDeleting(resourceId);
+    } else if (!resource.isPresent() || !resource.get().state().equals(ResourceState.DELETING)) {
+      // Fail the flight if resource is not found or resource state is not DELETING(DELETED,
+      // CREATING, HANDED_OUT).
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL);
```
wchamber, line 28: extra states in the comment?

src/main/java/bio/terra/rbs/service/resource/flight/InitialResourceDeletionStep.java
```diff
@@ -24,24 +22,17 @@ public StepResult doStep(FlightContext flightContext) {
...
-    // simple and it should work for most cases unless GCP outrage happens in which case there is
-    // nothing we can do.
-    if (rbsDao.updateResourceAsReady(
-        ResourceId.retrieve(inputMap),
-        inputMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class))) {
-      return StepResult.getStepResultSuccess();
-    } else {
-      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY);
-    }
+    // We can not undo a deletion
```
wchamber, line 35: but we could undo changing the RBS state - we're just not going to bother to try.
If we do then undo just this step, there's nothing wrong with putting the resource back in the state it was. We just have problems if we put it back as READY when it wasn't ready to begin with. So this is more us not wanting to work too hard for a case we think is rare, I think.

wchamber (approved):

src/test/java/bio/terra/rbs/integration/DeleteProjectFlightIntegrationTest.java
yonghaoy: Created a `newFullGcpConfig`

PR #40, opened 2020-10-29
[PF-216] Initial get pool info endpoint

yonghaoy:
I think it might be helpful if client want to know if pool is ready to use. 
Will add more if we need in the future.

### DIFF
--- a/src/main/resources/api/service_openapi.yaml
+++ b/src/main/resources/api/service_openapi.yaml
@@ -48,6 +48,31 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorReport'
+
+  '/api/pool/v1/{poolId}':
+    get:
+      summary: Get the pool info, including resource count in each state, and pool config.
+      operationId: getPoolInfo
+      parameters:
+        - $ref: '#/components/parameters/PoolId'
+      tags:
+        - rbs
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PoolInfo'
+        '404':
+          description: No resource available in this pool.
+        '500':
+          description: Get request error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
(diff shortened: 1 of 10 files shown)


src/main/resources/api/service_openapi.yaml
```diff
@@ -154,6 +179,42 @@ components:
...
+            Number of resources already handed out
+          type: integer
+        readyCount:
+          description: |
+            Number of resource ready for handing out.
+          type: integer
+        deletedCount:
+          description: |
+            Number of resource ready already deleted.
+          type: integer
```
wchamber, line 208: consider a resourceCount map of resource state to integer value instead of fields for each. not sure if that'd be better, but I think this API is probably ultimately only used by people.
(If I'm a client, and I want to know if I can get a resource, I might as well try to get one rather than check the size and then try to get a resource)

src/test/java/bio/terra/rbs/app/controller/RbsApiControllerTest.java
```diff
@@ -104,4 +106,67 @@ public void handoutResource_invalidPoolId() throws Exception {
...
+        resourceId1,
+        new CloudResourceUid().googleProjectUid(new GoogleProjectUid().projectId("projectId")));
+
+    String response =
+        this.mvc
+            .perform(get("/api/pool/v1/" + poolId.id()))
+            .andExpect(status().isOk())
+            .andReturn()
+            .getResponse()
+            .getContentAsString();
```
wchamber, line 150: not necessarily this PR, but we were talking after standup today about using the client library for this instead of the spring http mvc library. I think we've all been copying this from DR, but DR folks are considering using their client library instead.

wchamber (approved):

src/test/java/bio/terra/rbs/app/controller/RbsApiControllerTest.java
yonghaoy: ah, okay! I will do it in another PR

PR #45, opened 2020-11-03
[PF-133] Add initial metrics for resource count by state and ready resource ratio

yonghaoy:


src/main/java/bio/terra/rbs/common/MetricsHelper.java
```diff
@@ -0,0 +1,112 @@
...
+          .newMeasureMap()
+          .put(
+              READY_RESOURCE_RADIO,
+              (double)
+                      Math.round(
+                          resourceStates.count(ResourceState.READY)
+                              * 1.0
+                              / poolAndResourceStates.pool().size()
+                              * 100)
+                  / 100)
```
wchamber, line 108: I don't think we need to `Math.round` or `* 100) / 100`. I think its ok to have a precise fraction - this is being recorded as a double.

src/main/java/bio/terra/rbs/common/MetricsHelper.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  /**
+   * Records the latest count of {@link PoolAndResourceStates} and ready resource count to pool size
+   * ratio for ACTIVATE pools.
+   */
+  public static void recordResourceStateCount(PoolAndResourceStates poolAndResourceStates) {
+    Multiset<ResourceState> resourceStates = poolAndResourceStates.resourceStates();
+    resourceStates.entrySet();
+    for (Multiset.Entry<ResourceState> resourceStateEntry : resourceStates.entrySet()) {
```
wchamber, line 74: One thing I ran into with janitor is that recording the metrics this way won't update values that go to 0. ResourceStates with 0 entries won't be in the multiset, so the metric for them won't be updated and will state out of date. We can get around this by iterating over *all* ResourceStates, and then `resourceStates.count(state)`.
e.g. https://github.com/DataBiosphere/terra-resource-janitor/blob/2b683c49bb1b21d0e618df8056236c65d538578c/src/main/java/bio/terra/janitor/service/cleanup/FlightScheduler.java#L136-L141
consider adding a test for this case with sequential calls to recordResourceStateCount.

src/main/java/bio/terra/rbs/common/MetricsHelper.java
```diff
@@ -0,0 +1,112 @@
...
+                  TagValue.create(poolAndResourceStates.pool().status().toString()))
+              .build();
+
+      STATS_RECORDER
+          .newMeasureMap()
+          .put(RESOURCE_STATE_COUNT, resourceStateEntry.getCount())
+          .record(tctx);
+    }
+
+    if (poolAndResourceStates.pool().status().equals(PoolStatus.ACTIVE)) {
```
wchamber, line 92: similarly, when a pool is deactivated, it's last known ratio will persist. Not sure what the right thing to do here is.

1. leave it.
2. set 1.
3. set a bogus value, like 2. or -1.
4. ?

wchamber (changes requested):

src/main/java/bio/terra/rbs/common/MetricsHelper.java
yonghaoy: ah, okay changed to precise fraction, I thought we don't want 0.3333333333333333....333, and we can. only keep 2 decimal?

src/main/java/bio/terra/rbs/common/MetricsHelper.java
yonghaoy: Done

src/main/java/bio/terra/rbs/common/MetricsHelper.java
yonghaoy: Set 1 sg and easier for us to setup alerts by the ratio

src/main/java/bio/terra/rbs/common/MetricsHelper.java
```diff
@@ -85,28 +84,24 @@ public static void recordResourceStateCount(PoolAndResourceStates poolAndResourc
...
+            .build();
+    STATS_RECORDER
+        .newMeasureMap()
+        .put(
+            READY_RESOURCE_RADIO,
+            (poolAndResourceStates.pool().status().equals(PoolStatus.ACTIVE)
+                ? resourceStates.count(ResourceState.READY)
+                    * 1.0
+                    / poolAndResourceStates.pool().size()
+                : 1))
```
wchamber, line 104: pull the calculation out to its own line and add a comment on it.

wchamber (approved):

PR #47, opened 2020-11-04
[PF-217] Few improvements in test to reduce test time

yonghaoy:
1. BEFORE_EACH_TEST_METHOD should be AFTER, as we want to mark the ApplicationContext dirty after test.
2. Reduce polling period and increase polling number
3. `Duration.plus` returns a copy of the duraion, so we need to use the new copy: 
`waited = waited.plus(Duration.ofMillis(poll.toMillis())); `
4. Reduce flight-submission-period in test

### DIFF
--- a/src/test/resources/application-integration-enable-scheduler.yml
+++ b/src/test/resources/application-integration-enable-scheduler.yml
@@ -2,3 +2,4 @@ rbs:
   primary:
     # Enable scheduler service which will submit flight by scanning db.
     scheduler-enabled: true
+    flight-submission-period: 5s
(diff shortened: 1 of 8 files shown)


src/test/resources/application-integration-enable-scheduler.yml
```diff
@@ -2,3 +2,4 @@ rbs:
   primary:
     # Enable scheduler service which will submit flight by scanning db.
     scheduler-enabled: true
+    flight-submission-period: 5s
```
wchamber, line 5: do you want to update the default value instead of overriding it here?

wchamber (approved):
How much does this help?

yonghaoy:
> How much does this help?

Reduced ~3 minutes. Not too much, but it does fixed some issues

src/test/resources/application-integration-enable-scheduler.yml
yonghaoy: I'd like our main app scans db every 1 minutes, and only increase to 5s for test. 5s sounds too aggressive for production code?

PR #49, opened 2020-11-05
[PF-218]: Creates storage logs

yonghaoy:
This has 180 days life cycle and an WRITER ACL. 
See: https://github.com/broadinstitute/gcp-dm-templates/blob/eeee90fa4619b273d07206a867b01914cdeb0a30/templates/project.py#L165

### DIFF
--- a/src/main/java/bio/terra/rbs/service/pool/GcpResourceConfigValidator.java
+++ b/src/main/java/bio/terra/rbs/service/pool/GcpResourceConfigValidator.java
@@ -17,7 +17,8 @@ import java.util.List;
  */
 public class GcpResourceConfigValidator implements ResourceConfigValidator {
   /** List of services required to be enabled. */
-  private static final List<String> REQUIRED_SERVICES = ImmutableList.of("compute.googleapis.com");
+  private static final List<String> REQUIRED_SERVICES =
+      ImmutableList.of("compute.googleapis.com", "storage-component.googleapis.com");
 
   @Override
   public void validate(ResourceConfig config) {
--- a/src/main/java/bio/terra/rbs/service/resource/flight/CreateBucketStep.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/CreateBucketStep.java
@@ -1,17 +1,14 @@
 package bio.terra.rbs.service.resource.flight;
 
+import static bio.terra.rbs.service.resource.FlightMapKeys.GOOGLE_PROJECT_ID;
+import static bio.terra.rbs.service.resource.flight.GoogleUtils.*;
+
 import bio.terra.cloudres.common.ClientConfig;
-import bio.terra.cloudres.google.api.services.common.OperationCow;
-import bio.terra.cloudres.google.cloudresourcemanager.CloudResourceManagerCow;
 import bio.terra.cloudres.google.storage.StorageCow;
 import bio.terra.rbs.generated.model.GcpProjectConfig;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
-import bio.terra.stairway.StepStatus;
-import bio.terra.stairway.exception.RetryException;
-import com.google.api.services.cloudresourcemanager.model.Project;
-import com.google.api.services.cloudresourcemanager.model.ResourceId;
 import com.google.cloud.storage.Acl;
 import com.google.cloud.storage.BucketInfo;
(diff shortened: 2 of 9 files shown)


src/main/java/bio/terra/rbs/service/resource/flight/CreateBucketStep.java
```diff
@@ -0,0 +1,63 @@
...
+import bio.terra.stairway.StepResult;
+import com.google.cloud.storage.Acl;
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.StorageOptions;
+import com.google.common.collect.ImmutableList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Creates the basic GCP project. */
+public class CreateBucketStep implements Step {
```
wchamber, line 20: Fix the comment. Improve the class name. This is not a step to create a generic bucket, it's specifically for storage logs.
Make sure the javadoc description for this class includes why we want to do such a thing.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -382,6 +408,17 @@ private void assertIamBindingsContains(Project project, List<IamBinding> iamBind
     }
   }
 
+  private void assertBucketExists(Project project) throws Exception {
```
wchamber, line 411: assertLogStorageBucketExists?

wchamber (changes requested):

src/main/java/bio/terra/rbs/service/resource/flight/CreateBucketStep.java
yonghaoy: Done. I was thinking let this steps handles all buckets creation. But we can have make each step creates one bucket if we have more..

src/main/java/bio/terra/rbs/service/resource/flight/CreateStorageLogBucketStep.java
```diff
@@ -16,9 +15,13 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** Creates the basic GCP project. */
-public class CreateBucketStep implements Step {
-  private final Logger logger = LoggerFactory.getLogger(CreateBucketStep.class);
+/**
+ * Creates the Storage log bucket for the project.
+ *
+ * <p>See <a href="https://cloud.google.com/storage/docs/access-logs">Storage logs introduction</a>
```
wchamber, line 21: per the link, don't we have to do step 3 of enabling logging on the bucket? From my understanding of this link, this is how to allow access logs for the bucket to be written. But the comments on Deployment Manager are about logging GCE stuff to this bucket. What is this bucket really for?

yonghaoy: Just checked with Doug, the DM comment is copt/paste error from export bucket:
https://github.com/broadinstitute/gcp-dm-templates/blob/eeee90fa4619b273d07206a867b01914cdeb0a30/templates/project.py#L117
we enable logging on the workspace bucket when it is created, not this storage log bucket. The bucket here is where are logs are written. We do step 1 and 2 here, step 3 is in the workspace bucket. 

Changed the comment little bit

yonghaoy: added comment

src/main/java/bio/terra/rbs/service/pool/GcpResourceConfigValidator.java
```diff
@@ -17,7 +17,8 @@
  */
 public class GcpResourceConfigValidator implements ResourceConfigValidator {
   /** List of services required to be enabled. */
-  private static final List<String> REQUIRED_SERVICES = ImmutableList.of("compute.googleapis.com");
+  private static final List<String> REQUIRED_SERVICES =
```
wchamber, line 20: not related to this PR - but if we update validation, we need to make sure old deactivated pools don't need to be modified. Just thought of that, I don't know if that's already the case.

wchamber (approved):

src/main/java/bio/terra/rbs/service/pool/GcpResourceConfigValidator.java
yonghaoy: this validate configs loaded from file. And the file only has activate pool. Should be fine

PR #50, opened 2020-11-05
[PF-217] Boost test step2: Use one test to verifies we can insert the same step twice

yonghaoy:
Also make IAM part of the `FullGcpProject`
This saves ~30 minutes

### DIFF
--- a/src/main/java/bio/terra/rbs/service/resource/flight/StepUtils.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/StepUtils.java
@@ -3,7 +3,6 @@ package bio.terra.rbs.service.resource.flight;
 import static bio.terra.rbs.service.resource.FlightMapKeys.RESOURCE_READY;
 
 import bio.terra.rbs.common.ResourceId;
-import bio.terra.rbs.common.ResourceState;
 import bio.terra.rbs.db.RbsDao;
 import bio.terra.rbs.generated.model.CloudResourceUid;
 import bio.terra.rbs.service.resource.FlightMapKeys;
@@ -16,13 +15,9 @@ public class StepUtils {
   /** Update resource state to READY and update working map's RESOURCE_READY boolean value. */
   public static void markResourceReady(RbsDao rbsDao, FlightContext flightContext) {
     FlightMap workingMap = flightContext.getWorkingMap();
-    ResourceId resourceId = ResourceId.retrieve(workingMap);
-    if (rbsDao.retrieveResource(resourceId).get().state().equals(ResourceState.CREATING)) {
-      // Just in case the step retries even after DB is updated to READY. That is a waste if we fail the flight.
-      rbsDao.updateResourceAsReady(
-          ResourceId.retrieve(flightContext.getWorkingMap()),
-          workingMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class));
-    }
+    rbsDao.updateResourceAsReady(
+        ResourceId.retrieve(flightContext.getWorkingMap()),
+        workingMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class));
     workingMap.put(RESOURCE_READY, true);
   }
   /** Check resource is already marked as READY. This can prevent a READY resource got rollback. */
--- a/src/main/java/bio/terra/rbs/service/resource/flight/StepUtils.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/StepUtils.java
@@ -18,7 +18,7 @@ public class StepUtils {
     FlightMap workingMap = flightContext.getWorkingMap();
     ResourceId resourceId = ResourceId.retrieve(workingMap);
     if (rbsDao.retrieveResource(resourceId).get().state().equals(ResourceState.CREATING)) {
-      // Just in case the step retries even after DB is updated to READY.
+      // Just in case the step retries even after DB is updated to READY. That is a waste if we fail the flight.
       rbsDao.updateResourceAsReady(
           ResourceId.retrieve(flightContext.getWorkingMap()),
           workingMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class));
--- a/src/main/java/bio/terra/rbs/service/resource/flight/StepUtils.java
+++ b/src/main/java/bio/terra/rbs/service/resource/flight/StepUtils.java
@@ -3,6 +3,7 @@ package bio.terra.rbs.service.resource.flight;
 import static bio.terra.rbs.service.resource.FlightMapKeys.RESOURCE_READY;
 
 import bio.terra.rbs.common.ResourceId;
+import bio.terra.rbs.common.ResourceState;
 import bio.terra.rbs.db.RbsDao;
 import bio.terra.rbs.generated.model.CloudResourceUid;
 import bio.terra.rbs.service.resource.FlightMapKeys;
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/rbs/service/resource/flight/StepUtils.java
```diff
@@ -15,9 +16,13 @@
   /** Update resource state to READY and update working map's RESOURCE_READY boolean value. */
   public static void markResourceReady(RbsDao rbsDao, FlightContext flightContext) {
     FlightMap workingMap = flightContext.getWorkingMap();
-    rbsDao.updateResourceAsReady(
-        ResourceId.retrieve(flightContext.getWorkingMap()),
-        workingMap.get(FlightMapKeys.CLOUD_RESOURCE_UID, CloudResourceUid.class));
+    ResourceId resourceId = ResourceId.retrieve(workingMap);
+    if (rbsDao.retrieveResource(resourceId).get().state().equals(ResourceState.CREATING)) {
+      // Just in case the step retries even after DB is updated to READY. That is a waste if we fail the flight.
```
wchamber, line 21: I don't understand what you mean by `That is a waste if we fail the flight`. I think you should say what would happen if we didn't have this check.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -302,20 +200,23 @@ public ErrorCreateProjectFlight(FlightMap inputParameters, Object applicationCon
   }
 
   public static class MultiInstanceStepFlight extends GoogleProjectCreationFlight {
```
wchamber, line 202: there should be javadoc on this class.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -302,20 +200,23 @@ public ErrorCreateProjectFlight(FlightMap inputParameters, Object applicationCon
...
-    private static Class<? extends Step> multiStepClazz;
+    /**
+     * Duplication check is to avoid the scenario that "Step is retried while previously one succeeded". If that's
+     * not the scenario, not need to have the multiple step check.
```
wchamber, line 205: Sorry, I'm not understanding this comment.

wchamber (changes requested):

src/main/java/bio/terra/rbs/service/resource/flight/StepUtils.java
yonghaoy: ah, sorry, actually we don't need to have the state check. Removed

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: move most of the explaining to class Javadoc.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -199,10 +198,18 @@ public ErrorCreateProjectFlight(FlightMap inputParameters, Object applicationCon
...
+   * e.g., when creating Network, polling operation result timeout, but Network is created when the
+   * step is retried.
+   */
   public static class MultiInstanceStepFlight extends GoogleProjectCreationFlight {
     /**
-     * Duplication check is to avoid the scenario that "Step is retried while previously one succeeded". If that's
-     * not the scenario, not need to have the multiple step check.
+     * Steps that doesn't need to handle "retry after succeed" scenario.
+     * CreateResourceDbEntityStep doesn't have long wait operations; CreateProjectStep is safer and better to just
+     * fail the flight if project id is already in use.
```
wchamber, line 212: I'm not sure I agree on special casing these. We can fail db operations for transaction locking.
CreateProjectStep doesn't seem safe, it seems likely to fail and we should be able to handle that. Looking at the impl, it looks wrong to me. If we can retrieve the randomly generated project id, it's probably because we just created it. What's the probability of project id collision?
If we're worried that much about collision, our GenerateProjectId step could verify that a project with the id being generated doesn't exist before adding it to the working map, but this seems pretty paranoid to me.

wchamber (changes requested):

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: Yes, we can make all steps can handle duplicate steps, but I don't think we need to add those checks for every steps with some extra dp read or gcp read. I still doubt the necessary of having the extra network/route/dns query for every flight just to avoid the very low chance having the "retry but step success" chance here. Like I said here: https://github.com/DataBiosphere/terra-rbs/pull/31#discussion_r510494771.

I skip CreateResourceDbEntityStep because if we have such case, the flight will fail not success. (I don't think that will happen unless some issues in Stairway, and I don't want to have those extra db query to avoid rare case).
For CreateProjectStep, probability project id collision depends on the word dictionary we choose. Yes, it is possible we just created it, and we can have some checks in GenerateProjectId by querying gcp. But comparing with the extra query on every flights, I would prefer fail the flight.
I think my javadoc is misleading here, "CreateProjectStep is safer and better to **fail**", not "CreateProjectStep is safer and better".

wchamber: "When we just have network, we added the duplicated check, but as we have more and more stuff, I feel it is better to rollback the flight comparing with added checks in every single steps. Like you said in that comment, "It'd be a shame to throw away the project because the network api call fails." I think at lease we should have a clear definition about failing which steps is shameful, which are not."
The more steps we have, the higher probability that at least one of them will fail. This happens nonlinearly. Stairway enforces that any failing step causes the whole flight to be redone, so I think any step failing is shameful.
I think the question is - "is it worth the code complexity to make project creation flights more robust?" That depends on how complex allowing failed retries is vs how often it happens. If the extra gets seems too complex, or too paranoid, we can strip them out for now and add them back later. I'm really not worried about the performance hit of a 'get' to gcp - enabling apis takes minutes compared to milliseconds.

So we can exclude some of the tests here, but only with good comments about how we don't think it's worth enforcing that the step can happen multiple times.

yonghaoy: what is the actional item for this PR? I think we already have comments to explain why we are making a get call first and expect for 404 error in those steps class.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -300,21 +198,30 @@ public ErrorCreateProjectFlight(FlightMap inputParameters, Object applicationCon
...
+   * step is retried.
+   */
   public static class MultiInstanceStepFlight extends GoogleProjectCreationFlight {
-    private static Class<? extends Step> multiStepClazz;
+    /**
+     * Steps that doesn't need to handle "retry after succeed" scenario, if duplicates happens, the flight will fail
+     * instead of success.
+     */
+    private static final List<Class<? extends Step>> SKIP_DUP_CHECK_STEP_CLAZZ =
+        ImmutableList.of(CreateResourceDbEntityStep.class, CreateProjectStep.class);
```
wchamber, line 214: add `<li>` comments for why each of these steps are not ones we care to support multiple instance behavior. i.e. why it's safer for the createprojectstep to fail and why it's unlikely for the db entity step to fail.

wchamber (approved):

PR #55, opened 2020-11-12
[PF-196] Add cleanup_record table to track the resources published to Janitor

yonghaoy:
In my next PR, there will be a scheduler:
```
List<Resource> resources = rbsDao.retrieveResourceToCleanup();
for(Resource resource : resources) {
   publish(resource);
   insertCleanupRecord(resource.id());
}
```

### DIFF
--- a/src/main/java/bio/terra/rbs/db/RbsDao.java
+++ b/src/main/java/bio/terra/rbs/db/RbsDao.java
@@ -295,6 +295,46 @@ public class RbsDao {
     return jdbcTemplate.update(sql, params) == 1;
   }
 
+  /** Inserts a record into cleanup_record table. */
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public void insertCleanupRecord(ResourceId resourceId) {
+    String sql = "INSERT INTO cleanup_record (resource_id) values (:resource_id)";
+    MapSqlParameterSource params =
+        new MapSqlParameterSource().addValue("resource_id", resourceId.id());
+
+    jdbcTemplate.update(sql, params);
+  }
+
+  /**
+   * Retrieves resources that need to cleanup by Janitor. Those resources should be:
+   *
+   * <ul>
+   *   <li>State is HANDED_OUT in resource table
+   *   <li>Handed out before given {@code handedOutBefore}
+   *   <li>Not already in cleanup_record table
+   * </ul>
+   */
+  @Transactional(propagation = Propagation.SUPPORTS)
+  public List<Resource> retrieveResourceToCleanup(Instant handedOutBefore, int limit) {
+    String sql =
+        "select r.id, r.cloud_resource_uid, r.pool_id, r.state, r.request_handout_id, "
+            + "r.creation, r.deletion, r.handout_time "
+            + "FROM resource r "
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -295,6 +295,46 @@ public boolean deleteResource(ResourceId id) {
...
+
+    jdbcTemplate.update(sql, params);
+  }
+
+  /**
+   * Retrieves resources that need to cleanup by Janitor. Those resources should be:
+   *
+   * <ul>
+   *   <li>State is HANDED_OUT in resource table
+   *   <li>Handed out before given {@code handedOutBefore}
```
wchamber, line 313: why do we need the instant argument?

src/main/java/bio/terra/rbs/db/RbsDao.java
```diff
@@ -295,6 +295,46 @@ public boolean deleteResource(ResourceId id) {
     return jdbcTemplate.update(sql, params) == 1;
   }
 
+  /** Inserts a record into cleanup_record table. */
```
wchamber, line 298: Let's explain the semantics of the cleanup_record here. Why would I want to insert something into the cleanup record? When wouldn't I?

src/main/resources/db/changesets/20200925_initial_schema.yaml
```diff
@@ -106,3 +106,15 @@ databaseChangeLog:
                 constraints:
                   nullable: true
                   unique: false
+      - createTable:
+          tableName: cleanup_record
```
wchamber, line 110: add some comments about what this is for, especially since we're not planning on using it in prod.

wchamber (changes requested):

src/main/resources/db/changesets/20200925_initial_schema.yaml
yonghaoy: added

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: done

src/main/java/bio/terra/rbs/db/RbsDao.java
yonghaoy: I was thinking about not to publish those resources just handed out and leave some 'cache time' in RBS. But I realize that is overkill, that is overlap with Janitor's TTL

wchamber (changes requested):
I think you forgot to push your commits. I don't see any changes.

src/main/resources/db/changesets/20200925_initial_schema.yaml
```diff
@@ -106,3 +106,16 @@ databaseChangeLog:
...
+          columns:
+            - column:
+                name: resource_id
+                type: uuid
+                constraints:
+                  primaryKey: true
+                  references: resource(id)
+                  foreignKeyName: fk_resource_id
+                  unique: false
```
wchamber, line 120: shouldn't these be unique?

wchamber (approved):

wchamber (approved):
thanks!

PR #56, opened 2020-11-16
[PF-168] Generate Project Id from resource config

yonghaoy:
For now, we only support prefix + random char.
The Random char scheme is from AoU's code: 
https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java#L288


src/main/resources/config/resource_schema.yaml
```diff
@@ -63,6 +60,28 @@ components:
...
+        projectIDPrefix:
+          description: |-
+            The created project name prefix.
+          type: string
+        projectIDScheme:
+          type: string
+          # swagger-codegen always looks for and removes a common prefix for
+          # enum values, which breaks enums that only have a single value.
+          # PLACEHOLDER_VALUE should be removed when we have an actual second value
+          # to use.
```
wchamber, line 77: that's cool. thanks swagger!

src/main/resources/config/resource_schema.yaml
```diff
@@ -63,6 +60,28 @@ components:
...
+      description: |-
+        How to generate project ID, including prefix and the nameing schema. For example, projectIDPrefix is
+        aou_rw, projectIDScheme is random_char, the generated project ID might be aou_rw_a1bc23
+      properties:
+        projectIDPrefix:
+          description: |-
+            The created project name prefix.
+          type: string
+        projectIDScheme:
+          type: string
```
wchamber, line 73: do we want to start without any scheme if we only have one value, or are you very confident we'll have another value soon?

src/main/resources/config/resource_schema.yaml
```diff
@@ -63,6 +60,28 @@ components:
           $ref: '#/components/schemas/Network'
       type: object
 
+    ProjectIDGenerator:
```
wchamber, line 63: nit: maybe `ProjectIdSchema` instead? 'ProjectIdPattern`? Generator sounds like something that makes a thing, but this is the blueprint for how to make things.
(make sure to lowercase the 'd' in Id.)

src/main/resources/config/resource_schema.yaml
```diff
@@ -63,6 +60,28 @@ components:
...
+      description: |-
+        How to generate project ID, including prefix and the nameing schema. For example, projectIDPrefix is
```
wchamber, line 65: typo: s/nameing/naming

src/main/resources/config/resource_schema.yaml
```diff
@@ -63,6 +60,28 @@ components:
...
+        aou_rw, projectIDScheme is random_char, the generated project ID might be aou_rw_a1bc23
+      properties:
+        projectIDPrefix:
```
wchamber, line 68: I think `prefix` would be fine here. Don't think we need to repeat "projectId". ditto on the Scheme.

src/main/java/bio/terra/rbs/service/resource/projectid/GcpProjectIDGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.hash.Hashing;
+import java.util.UUID;
+import org.springframework.stereotype.Component;
+
+/** Generates Project ID for GCP project from config. */
+@Component
+public class GcpProjectIDGenerator {
+  /** The size of project when generating random characters. */
+  @VisibleForTesting static final int RANDOM_ID_SIZE = 8;
```
wchamber, line 13: that's not super big. Why 8?

src/test/java/bio/terra/rbs/integration/IntegrationUtils.java
```diff
@@ -105,7 +106,10 @@ public static Pool preparePool(RbsDao rbsDao, GcpProjectConfig gcpProjectConfig)
   /** Create a Basic {@link ResourceConfig}. */
   public static GcpProjectConfig newBasicGcpConfig() {
     return new GcpProjectConfig()
-        .projectIDPrefix("prefix")
+        .projectIDGenerator(
+            new ProjectIDGenerator()
+                .projectIDPrefix("prefix")
+                .projectIDScheme(ProjectIDGenerator.ProjectIDSchemeEnum.RANDOM_CHAR))
```
wchamber, line 112: this test specifies the scheme, but others don't. let's be consistent on whether it's required or not.

wchamber (changes requested):
few nits

src/main/resources/config/resource_schema.yaml
yonghaoy: I spent sometime searching around the solution for enum in Java's swagger codegen, and found this: https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/resources/api/service_openapi.yaml#L438

Thanks WM!

src/main/java/bio/terra/rbs/service/resource/projectid/GcpProjectIDGenerator.java
```diff
@@ -0,0 +1,41 @@
+package bio.terra.rbs.service.resource.projectid;
```
gjuggler, line 1: Same nitpick as Will, in this class name and a couple methods below: be sure "d" in "Id" is always lowercase. c.f. https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

gjuggler (commented):
+1 to most of Will's nits, looks fine otherwise. I'm especially persnickity about the "Id" not "ID" thing, since it's super annoying when as a reader / user of a system you have to remember where it's one way vs. another.

src/main/resources/config/resource_schema.yaml
yonghaoy: I think we will have another scheme in few weeks. Just watiing for the GCP feedbacou about using their word dictionary...

src/test/java/bio/terra/rbs/integration/IntegrationUtils.java
yonghaoy: yes, this is required, others don't have it because this is the only integration test which will use this field. Others are unit test which just need some general GcpProjectConfig. 
Make other tests also set this field even it is not used.

src/main/java/bio/terra/rbs/service/resource/projectid/GcpProjectIdGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.hash.Hashing;
+import java.util.UUID;
+import org.springframework.stereotype.Component;
+
+/** Generates Project Id for GCP project from config. */
+@Component
+public class GcpProjectIdGenerator {
+  /** The size of project when generating random characters. */
+  @VisibleForTesting static final int RANDOM_ID_SIZE = 8;
```
yonghaoy, line 13: @wchamber @gjuggler 
re Will's comment about the size. 
I followed AoU's pattern: https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/billing/BillingProjectBufferService.java#L46
And also want to check with Greg: Does 8 enough? And what would be a better length? If we increase the size, once AoU starts using RBS, will that be a breaking change since AoU is 8?

wchamber: At the least, I think we should comment on why it is 8 - historical reasons is a good comment.
I think in general we should very much avoid clients relying on the particular format of the project id.

wchamber (approved):

src/main/java/bio/terra/rbs/service/resource/projectid/GcpProjectIdGenerator.java
yonghaoy: ah, okay, added some comments. And I think 8 should be big enough for a long period... We also have prefix aou-, 
and for a pool, 8^36 is a lot.

PR #59, opened 2020-11-17
[PF-184][PF-136] Setup tools/dev/toolsalpha RBS resource config

yonghaoy:
Tools: For Rawls integration test
Toolsalpha: For local RBS server and personal environment
Dev: For AoU test and Rawls Dev

I am having hard time to comeup those names for config and pool...


src/main/resources/config/dev/resource-config/aou_ws_0_0_1.yml
```diff
@@ -2,9 +2,11 @@
 ---
 configName: "aou_ws_resource_v1"
 gcpProjectConfig:
-  projectIDPrefix: "aou-rw-test"
+  projectIdSchema:
+    prefix: "aou-rw-test"
+    scheme: "RANDOM_CHAR"
   # TODO: Fill this
-  parentFolderId: null
+  parentFolderId: "637867149294"
```
gjuggler, line 9: I'd add a comment here clarifying what folder this points to, e.g.

```# test.firecloud.org/terra_dev_aou_test```

File: src/main/resources/config/dev/resource-config/cwb_ws_0_0_1.yml [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
+    - "dns.googleapis.com"
+  iamBindings:
+    - members: ["group:firecloud-project-owners@test.firecloud.org"]
+      role: "roles/owner"
+    - members: ["user:yyu@broadinstitute.org", "serviceAccount:806222273987-gffklo3qfd1gedvlgr55i84cocjh8efa@developer.gserviceaccount.com", "serviceAccount:806222273987-ksinuueghug8u81i36lq8aof266q19hl@developer.gserviceaccount.com"]
+      role: "roles/editor"
+  network:
+    enableNetworkMonitoring: "true"
```
gjuggler, line 35: Unless you've already done so, best to double-check w/ @dvoet whether network monitoring should be turned on for projects in the CWB pool. The main practical impacts would be

(1) private Google access, which works fine for AoU but would require some testing / validation with CWB apps, and which also incurs some small additional cost due to the use of Cloud DNS.
(2) VPC flow logs, which result in additional Cloud Logging storage cost. Though default retention is 30 days, so it's unlikely to cause a meaningful uptick in cost to users.

src/main/resources/config/dev/resource-config/aou_ws_0_0_1.yml
yonghaoy: done.

src/main/resources/config/dev/resource-config/cwb_ws_0_0_1.yml
yonghaoy: Thanks, the version you reviewed was my local testing version. Now the PR is ready for review. Thanks!

src/main/resources/config/dev/pool_schema.yml
```diff
@@ -1,6 +1,9 @@
-# RBS Pools Schema in dev encironment
+# RBS Pools Schema in dev environment
 ---
 poolConfigs:
-  - poolId: "aou_ws_dev_v1"
+  - poolId: "aou_ws_test_v1"
```
gjuggler, line 4: nit / suggestion: I'd add a comment above this line to give space for a more long-form description of what this pool is and who it serves. (Do the same for the CWB pool too.)

Possible comment text for this pool:

"Pool for GCP projects to be used by workspaces in the AoU test environment. Rawls is the proximal owner of these projects; AoU sits as a layer atop Rawls."

src/main/resources/config/dev/resource-config/aou_ws_0_0_1.yml
```diff
@@ -1,13 +1,12 @@
...
+configName: "aou_ws_resource_test_v1"
 gcpProjectConfig:
   projectIdSchema:
     prefix: "aou-rw-test"
     scheme: "RANDOM_CHAR"
-  # TODO: Fill this
-  parentFolderId: null
-  # TODO: Fill this
+  # test.firecloud.org/terra_dev_aou_test
+  parentFolderId: "737976594150"
```
gjuggler, line 9: Hmm -- just thinking aloud, it seems likely that we're going to end up with many near-duplicate resource configs where the only diff between environments is the parentFolderId. Maybe it's not worth doing immediately, but a scheme where the pool config could override some subset of the resource config might provide a way to reduce that duplication.

src/main/resources/config/dev/resource-config/aou_ws_0_0_1.yml
```diff
@@ -26,3 +25,5 @@ gcpProjectConfig:
     - "storage-api.googleapis.com"
     - "storage-component.googleapis.com"
     - "dns.googleapis.com"
```
gjuggler, line 27: nit: alphabetize all API entries (here and in all other resource configs)

src/main/resources/config/tools/pool_schema.yml
```diff
@@ -0,0 +1,6 @@
+# RBS Pools Schema in tools environment
+---
+poolConfigs:
+  - poolId: "cwb_ws_test_v1"
```
gjuggler, line 4: broken record, but what is "test"? That's not a CWB environment or a Terra environment name. Maybe this would be clearer with a comment describing the pool's use case in words... but I also think a different short-token would be clearer. Maybe "cwb_ws_fiab_v1" if this is meant to provide a pool to be used by FIAB CWB tests?

src/main/resources/config/dev/resource-config/aou_ws_0_0_1.yml
```diff
@@ -1,13 +1,12 @@
 # All Of Us buffered workspace template
```
gjuggler, line 1: General note: what is the purpose of the pseudo-semantic-versioning used in the filename? Given that the configName only has a "v1" suffix, is there any benefit of having a different versioning format for the filename? My instinct would be to align them to keep things simple & consistent.

src/main/resources/config/toolsalpha/pool_schema.yml
```diff
@@ -0,0 +1,6 @@
+# RBS Pools Schema for local testing and on personal environment
+---
+poolConfigs:
+  - poolId: "ws_resource_toolsalpha_v1"
```
gjuggler, line 4: I think "ws" is meant to represent "workspace", but arguably this pool for RBS integration testing has nothing to do with workspaces – right? This comment applies to all of the above pool IDs as well. Is there an important reason why "ws" is part of all these names? If it's universally used, then it doesn't seem to provide much differentiating power.

src/main/resources/application.yml
```diff
@@ -10,7 +10,7 @@ rbs:
   pool:
     # The Folder to have all RBS configs.
     # TODO: Investigate where and how to put those files.
-    configPath: config/dev
+    config-path: config/toolsalpha
```
gjuggler, line 13: This is an interesting change. Can you add a comment clarifying why exactly we refer to toolsalpha from this file? I suspect the answer is "point to toolsalpha for local testing" but it's very helpful to make that explicit here since it's kind of non-obvious.

gjuggler (changes requested):

src/main/resources/config/dev/resource-config/aou_ws_0_0_1.yml
yonghaoy: yeah, I realized it as well.  The current config loading logic is simple, we pass in the entire configPath directory name and assume that the files are already well oranizaed in this simple way. 
Is there any exsiting exampels/well-known pattern about yaml override in the similar scenario?

src/main/resources/config/tools/pool_schema.yml
yonghaoy: Changed to fiab_.  Thanks!

src/main/resources/config/toolsalpha/pool_schema.yml
yonghaoy: removed `ws` Thanks

src/main/resources/application.yml
yonghaoy: Thanks for pointing out. Now I feel toolsalpha shouldn't be here. I am not setting any default values now. And will set it in run_local.sh and helmfile instead. 
https://github.com/broadinstitute/terra-helmfile/pull/613
But I will keep the configPath: record with the readme and comment here.

gjuggler (approved):
Thanks for the changes!

PR #61, opened 2020-11-18
[PF-184] Upgrade RBS to java 11

yonghaoy:
Need to keep rbs-client compatible with java8


build.gradle
```diff
@@ -136,7 +136,7 @@ task extractProfilerAgent(dependsOn: downloadProfilerAgent, type: Copy) {
 
 jib {
     from {
-        image = "openjdk:8-jdk-alpine"
+        image = "openjdk:8"
```
wchamber, line 139: why aren't we using jdk:11? How about "adoptopenjdk:11-jre-hotspot" to match WM and Janitor?

yonghaoy: oh, right, I should use Java11. I had instinction that Rawls want Java8, so I use that at the time when I setup RBS. Changed to 11 now.

wchamber (approved):

PR #62, opened 2020-11-19
[PF-137] Add security schema for RBS endpoint

yonghaoy:
Apache proxy needs this for authN/authZ


marikomedlock (approved):
LGTM.
I compared this to the open api YAML for workspace manager, and changes follow the same pattern. (Note it's slightly different than for data repo, but deployment there is handled differently.)

PR #63, opened 2020-11-19
[PF-184]: Set Acl after bucket is created to let bucket has bucket gets default permissions plus new ones

yonghaoy:


src/main/java/bio/terra/rbs/service/resource/flight/CreateStorageLogBucketStep.java
```diff
@@ -54,8 +54,9 @@ public StepResult doStep(FlightContext flightContext) {
     storageCow.create(
         BucketInfo.newBuilder(bucketName)
             .setLifecycleRules(ImmutableList.of(STORAGE_LOGS_LIFECYCLE_RULE))
-            .setAcl(ImmutableList.of(STORAGE_LOGS_WRITE_ACL))
             .build());
+    // Set Acl as a separate call so bucket gets default permissions plus this one.
```
gjuggler, line 58: [Initial comment]

It's not really clear what exactly those "default permissions" are. I see that this comment traces back to https://github.com/broadinstitute/gcp-dm-templates/blame/master/templates/project.py#L213 which is a similarly cryptic comment. Is there any additional detail you can add to shine light on what default ACL entry we are trying to preserve?

[After reading a bit more on https://cloud.google.com/storage/docs/access-control/lists#default]

Yuck, I didn't realize just how messy GCS ACLs are. I wonder if we should opt to use best practices here, which per Google's documentation seems to be to use IAM rather than ACLs where possible.

[After reading even more at https://cloud.google.com/storage/docs/uniform-bucket-level-access#restrictions]

Welp, it sounds like for this use case (audit / storage logs written by Google-internal systems), we may not be able to use uniform bucket level access. It still may be possible to assign the permission using IAM rather than ACLs, but it looks like the upside is limited.

[Conclusion]
It's probably best to just link to https://cloud.google.com/storage/docs/access-control/lists#default in the comment here, so future readers can understand why we're doing this in two steps.

(Technically, it probably would be possible to replicate the default ACLs in the single setAcl call. But it feels safer to not replicate any of Google's defaults here.)

As an aside – I still think IAM will generally be our preferred way of managing access to buckets, e.g. as a controlled resource within a workspace. It will be nice to track the small task of adding that support to CRL in anticipation of preferring that model.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -354,8 +354,15 @@ private void assertLogStorageBucketExists(Project project) throws Exception {
         new StorageCow(clientConfig, StorageOptions.newBuilder().setProjectId(projectId).build());
     String bucketName = "storage-logs-" + projectId;
     BucketInfo bucketInfo = storageCow.get(bucketName).getBucketInfo();
-    assertEquals(bucketInfo.getAcl().get(0).getEntity(), STORAGE_LOGS_WRITE_ACL.getEntity());
-    assertEquals(bucketInfo.getAcl().get(0).getRole(), STORAGE_LOGS_WRITE_ACL.getRole());
+    // There might be multiple ACLs as we didn't remove the default ACLs. Only need to verify the
```
gjuggler, line 357: Do we want to actually verify that there are multiple ACLs? It seems important, since otherwise I could just revert the code change and the test would still pass.

gjuggler (approved):
See below for one long-winded comment on my journey of sorrow re: bucket ACLs, and one small request for add'l verification on the integration test. OK to merge after those are resolved.

src/test/java/bio/terra/rbs/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: I see it as CRL function and responsibility, so I just added the test in CRL code:
https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/google-storage/src/test/java/bio/terra/cloudres/google/storage/StorageCowTest.java#L123

Having the same test in RBS makes I feel it is testing CRL's code. But I can verify that the ACL is multiple here as well.

gjuggler: 👍 sounds fine, thanks for the pointer

src/main/java/bio/terra/rbs/service/resource/flight/CreateStorageLogBucketStep.java
yonghaoy: As I verified in Default ACLs that we dropped if we do it in the same line:
`memeber: Editors of project: prefix-e47c973d role: bucket owner`
And
`viewer: Editors of project: prefix-e47c973d role: bucket viewer`
Created a PF-231 to track that.

PR #65, opened 2020-11-20
[PF-134] Export metrics and tacking to stackdriver

yonghaoy:


src/main/java/bio/terra/buffer/app/configuration/StackdriverConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "buffer.stackdriver")
+public class StackdriverConfiguration {
+  /** Whether to enable stackdriver metrics & tracing collection. */
+  private boolean enabled = true;
+
+  /** The probability to record a trace. Should be between 0 and 1. */
+  private double traceSampleProbability = 1.;
```
wchamber, line 15: 1. is a big default. WDYT about 0.1, or 0.01?

src/main/resources/application.yml
```diff
@@ -14,6 +14,8 @@ buffer:
     # Whether to update pool when config changes during server startup.
     # Might be useful to avoid potential backward compatibility caused by config change and make tests easier.
     update-pool-on-start: true
+  stackdriver:
+    enabled: true
```
wchamber, line 18: if we're default enabled, I don't think we need this extra default enabling.

wchamber (approved):

PR #68, opened 2020-11-24
[no ticket ] Fix client library's security schema issue and local runner script config

yonghaoy:
Initially we mix use `googleoauth` with `authroization` in request header


marikomedlock (approved):
Confirmed that this fixed the missing bearer token error I was seeing when using the client library. Thank you!

PR #72, opened 2020-12-03
[PF-254] Add pool config for RBS performance test. 

yonghaoy:
Set pool size to 1500. Not sure if that is the right value for a perf test.

### DIFF
--- a/src/main/resources/config/perf/resource-config/resource_perf_v1.yml
+++ b/src/main/resources/config/perf/resource-config/resource_perf_v1.yml
@@ -1,4 +1,4 @@
-# Resource template for running performace test for RBS
+# Resource template for running buffer service performance test
 ---
 configName: "resource_perf_v1"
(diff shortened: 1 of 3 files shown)


src/main/resources/config/perf/resource-config/resource_perf_v1.yml
```diff
@@ -0,0 +1,34 @@
+# Resource template for running buffer service performance test
```
wchamber, line 1: I would have hoped that we can use the same configs across multiple environments. Why can't we? As is, we want this to probably stay mostly in sync between this, dev, tools, etc.
For perf, I think I could buy the argument that we might want to test different parts of creation. But not so much for dev/alpha/prod/tools

wchamber (approved):
Not going to block the review, but I think we should revisit sharing resource configs.

PR #73, opened 2020-12-04
[PF-254] Add performance test using test runner

yonghaoy:
Test steps:
Before: Poll until resource pool is full.
During: Handout X number of resources, 
After:  Poll until resource pool is full.

For now, I just let the code ready. May need more work to finalize the pool config(size, threads, etch.)


buffer-clienttests/src/main/java/scripts/utils/BufferServiceUtils.java
```diff
@@ -20,6 +25,15 @@
 public class BufferServiceUtils {
   private static final Logger logger = LoggerFactory.getLogger(BufferServiceUtils.class);
 
+  /**
+   * The pool id to get projects from. Pool config can be found at src/resources/config/perf folder
+   * under Buffer Service repo.
+   */
+  public static final String POOL_ID = "resource_toolsalpha_v1";
```
marikomedlock, line 32: Is the pool id an environment-specific string? (i.e. Will this test script still work pointed at say, a perf environment?).

buffer-clienttests/src/main/resources/configs/BasicAuthenticated.json
```diff
@@ -6,11 +6,11 @@
...
-      "name": "GetPoolInfo",
-      "numberOfUserJourneyThreadsToRun": 10,
-      "userJourneyThreadPoolSize": 2,
-      "expectedTimeForEach": 5,
-      "expectedTimeForEachUnit": "SECONDS"
+      "name": "HandoutResource",
+      "numberOfUserJourneyThreadsToRun": 3,
+      "userJourneyThreadPoolSize": 3,
+      "expectedTimeForEach": 2,
+      "expectedTimeForEachUnit": "HOURS"
```
marikomedlock, line 13: I think you should leave this BasicAuthenticated.json file the same and create a new one (e.g. EmptyPoolAndWaitForRefill.json) that exercises your new test script.

buffer-clienttests/src/main/java/scripts/testscripts/HandoutResource.java
```diff
@@ -0,0 +1,72 @@
...
+      logger.info(
+          "Caught exception requesting resource, handoutRequestId: {}", handoutRequestId, apiEx);
+    }
+  }
+
+  @Override
+  public void cleanup(List<TestUserSpecification> testUsers) throws Exception {
+    // For now we verifies all RESOURCE_COUNT calls successfully. Not sure that is too ideal or we
+    // want to set some threshold like we allow 0.1% failure rate is allowable in this burst case.
+    PoolInfo poolInfo = pollUntilPoolFull(bufferApi, Duration.ofHours(2), poolSize);
```
marikomedlock, line 68: You may want to get a new `bufferApi` in this method so that the service account credential is refreshed before you use it here.

marikomedlock (changes requested):
Some comments inline. Also, did the render-config.sh script change at all or get deleted?

buffer-clienttests/src/main/java/scripts/utils/BufferServiceUtils.java
yonghaoy: that is environment specific thing. The javadoc above explained how to get the pool id. 
Changed to perf on by default

yonghaoy:
> Some comments inline. Also, did the render-config.sh script change at all or get deleted?

The script reminds there with no change. I just make that script executable.

buffer-clienttests/src/main/java/scripts/utils/BufferServiceUtils.java
marikomedlock: Got it, okay then I think we should derive this property from the ServerSpecification object instead of hard-coding it here. Given the Buffer Service URL and client SA, can we lookup this property somehow, or the only way to find it for a particular deployment is to look in the Helm charts or application.properties for that deployment?

marikomedlock (commented):
Sorry, did you make any changes since my last review -- maybe forgot to push your branch?

buffer-clienttests/src/main/java/scripts/utils/BufferServiceUtils.java
yonghaoy: I think we may want to put it in ServerSpecification. Is the correct order is:
Change code in test-runner and added it into ServerSpecification, then publish a new package? 
Can we make some improvement on ServerSpecification and make this not depend on test-runer code&publishing? 
I don't think it will scale well once we have many apps uses the test runner. I think there are some 'spring configuration library' available we can use. Then clients can have their own configs.

buffer-clienttests/src/main/resources/servers/perf.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "yyu-dev",
```
gjuggler, line 2: nit: should this server config be named "perf" to match the filename?

buffer-clienttests/src/main/java/scripts/utils/BufferServiceUtils.java
gjuggler: My $.02, we should use the server config name to key this in a static lookup map.

As I understand it, the pool ID isn't possible to strictly determine from a deployment's application config, since RBS has potentially multiple pools within an environment.

What we're saying here is something like: "depending on which target environment I'm pointed at, I should use X, Y or Z pool ID to request resources, since that pool is appropriately set up for handling this test".

It may well be that we don't have an appropriate RBS pool to support this test within an arbitrary RBS deployment. That's fine, and I think that is correctly encoded by having a set of mapped values that we store here.

marikomedlock: It sounds like POOL_ID is part of the description of a Terra environment, which is what a ServerSpecification is intended to represent.

I agree we need an improvement here so we're not always updating Test Runner library whenever we need a new property, but I think this applies to nearly every property of ServerSpecification. We need a way to lookup information about a Terra environment dynamically. Putting it in a K8S config map that we can query is one solution. Until we have some solution, we are manually defining the required information for each Terra environment in these ServerSpecification JSON files.

For this particular property POOL_ID, is it something another service would also need when talking to Buffer Service? For example, say I want to write a WSM test script that checks the size of the Buffer Service pool first. If I would need the POOL_ID there also, then I think it should be added to ServerSpecification directly and not hard-coded just in buffer-service code. Otherwise, we may need to duplicate this hard-coding in other repositories. Longer-term, having the K8S map or other solution would obviate the need for this at all.

yonghaoy: 1. POOL_ID is not part of Terra env. It is RBS perf test only thing. If WSM, or TDR need RBS in their perf test, pool id is their environment variable just like how they use it in production instance. 
2. I am little bit confuse about the `configs for a Terra environment`
My guess what you want to have: ServerSpecification list all common Terra env configs like SAM url, WSM url, RBS url, RBS pool id. Then all test runner can share. 
My understanding: That will not work because all those values are actually part of their own configuration. e.g., in WSM integration test, they overide what ever they want, and e.g. WSM: https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/resources/application.yml#L19
So I am confulese about " a Terra environment" you mentioned. I think test runner belongs to each individual servers, and they can config what ever SAM/RBS/TDR they want to talk to in their test.

gjuggler: Maybe let's chat after standup about this one?

POOL_ID is especially tricky because each RBS serves multiple clients (e.g. AoU and CWB each pass in a different pool ID to the same prod RBS instance). So it's technically a client-specific config, even though the set of possible values is bounded by the particular RBS environment being targeted.

PR #74, opened 2020-12-04
[PF-254]: Move handout request id into requestBody for put operation

yonghaoy:
Similar to what TDR did: https://github.com/DataBiosphere/jade-data-repo/pull/631
Body can not be null for jersey2 client library

### DIFF
--- a/src/main/resources/api/service_openapi.yaml
+++ b/src/main/resources/api/service_openapi.yaml
@@ -36,6 +36,9 @@ paths:
       parameters:
         - $ref: '#/components/parameters/PoolId'
         - $ref: '#/components/parameters/HandoutRequestId'
+      requestBody:
+        required: false
+        content: {}
       tags:
         - buffer
       responses:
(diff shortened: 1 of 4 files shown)


src/main/resources/api/service_openapi.yaml
```diff
@@ -29,13 +29,18 @@ paths:
...
 
-  '/api/pool/v1/{poolId}/resource/{handoutRequestId}':
+  '/api/pool/v1/{poolId}/resource':
     put:
       summary: Get a resource from pool and update the resource state, each handoutRequestId is assigned to the resource no more than once. Using the same handoutRequestId will get the same ResourceInfo
       operationId: handoutResource
       parameters:
         - $ref: '#/components/parameters/PoolId'
-        - $ref: '#/components/parameters/HandoutRequestId'
+      requestBody:
```
marikomedlock, line 38: In the Data Repo PR, I see that each of the parameters has a "in:" line that indicates if it is in the query or body. Do you not need that here?

marikomedlock (approved):
One question inline from comparing to Data Repo related PR.

src/main/resources/api/service_openapi.yaml
yonghaoy: `in` is defined in parameters compoment, in line 90

PR #75, opened 2020-12-07
[no ticket] Catch conflict error when creating cloud resouces 

yonghaoy:
Instead of get then create, use create directly and catch 409 error code.


yonghaoy:
> Just curious, anything in particular cause you to make this change?

There are so many 404 error logs in gcp concole, and I am worried that would bring us more noise when we dubug some real issues.

PR #76, opened 2020-12-09
[PF-220] Add metrics to record handout request by pool id

yonghaoy:


src/main/java/bio/terra/buffer/common/MetricsHelper.java
```diff
@@ -36,6 +36,10 @@ private MetricsHelper() {}
           "ready resource count to pool size ratio",
           RESOURCE_TO_POOL_SIZE_RATIO);
 
+  private static final Measure.MeasureLong HANDOUT_RESOURCE_COUNT =
```
wchamber, line 39: Why do we need this? Don't we already have this information from the RESOURCE_STATE_COUNT?
As implementdd, this actually counts the number of handout *requests*; if I use the same handout id in two requests, they will both increment this count.

wchamber (changes requested):

src/main/java/bio/terra/buffer/common/MetricsHelper.java
yonghaoy: This is what Greg suggested from a 1 on 1: to have a metrics to track the handout request event, that would be a good metrics to measure pool qps stuff..

gjuggler: My thought was that having an event-style metric might be a useful complement to the existing gauge-style metrics. This PR seems to follow the same pattern as the existing counts. Is there no underlying difference in how OpenCensus tracks gauges vs. events?

wchamber: I believe the difference between gauges and events is in the View, `Aggregation.LastValue` vs `Aggregation.Count`. https://opencensus.io/stats/view/#aggregations

Greg, to be clear, do you want an event of the handout requests, handout events, or something else? Are we explicitly duplicating the information in another form for convenience, or measuring something slightly different?

yonghaoy:
@wchamber @gjuggler Moved metrics to controller. PTAL. Thanks

src/main/java/bio/terra/buffer/common/MetricsHelper.java
wchamber, line 39: I think this, and it's metric description, and the method name, comments, etc. should be updated to make it clear this is counting HANDOUT_RESOURCE_REQUEST_COUNT, - requests & not handouts. worth a comment on the distinction.

wchamber (changes requested):

src/main/java/bio/terra/buffer/common/MetricsHelper.java
yonghaoy: Changed view name ,counter name and method name.

wchamber (approved):

PR #78, opened 2020-12-09
[PF-220] Set spotbugs and jacoco for static analysis

yonghaoy:
Copied from Data Repo build.gradle
And those are parts of McTerra tech stack


build.gradle
```diff
@@ -235,6 +236,51 @@ swaggerSources {
...
+            destination = file("$buildDir/reports/spotbugs/test.html")
+        }
+    }
+}
+
+tasks.withType(JacocoReport) {
+    reports {
+        xml.enabled false
+        csv.enabled false
+        html.destination file("${buildDir}/jacocoHtml")
```
gjuggler, line 267: Nice! I made an attempt some time back to get jacoco running, but kept running into issues where it wouldn't actually output the report. Did you have to tinker much here, or is this following the basic instructions from jacoco docs?

Side Q: is there a fast path to getting the coverage reports uploaded to GitHub as an artifact for further analysis? No worries if not – we can easily follow up afterwards if it's worthwhile.

gjuggler (approved):

build.gradle
yonghaoy: The basic instructions doesn't actually generated the report, and it only does the `jacocoTestCoverageVerification` 
We see the report, we need  `./gradlew jacocoTestReport` I have it in our github action now, and also upload that report.

yonghaoy: @gjuggler

gjuggler: 👍

PR #83, opened 2020-12-16
Add pools for WM to RBS in dev and tools

tnadler:


src/main/resources/config/dev/resource-config/workspace_manager_v1.yml
```diff
@@ -0,0 +1,28 @@
...
+  - "dataflow.googleapis.com"
+  - "dataproc.googleapis.com"
+  - "dns.googleapis.com"
+  - "genomics.googleapis.com"
+  - "logging.googleapis.com"
+  - "monitoring.googleapis.com"
+  - "storage-api.googleapis.com"
+  - "storage-component.googleapis.com"
+  network:
+    enableNetworkMonitoring: "true"
```
yonghaoy, line 28: nit: add a new line at the end of file

src/main/resources/config/tools/resource-config/workspace_manager_v1.yml
```diff
@@ -0,0 +1,28 @@
...
+  - "dataflow.googleapis.com"
+  - "dataproc.googleapis.com"
+  - "dns.googleapis.com"
+  - "genomics.googleapis.com"
+  - "logging.googleapis.com"
+  - "monitoring.googleapis.com"
+  - "storage-api.googleapis.com"
+  - "storage-component.googleapis.com"
+  network:
+    enableNetworkMonitoring: "true"
```
yonghaoy, line 28: ditto

src/main/resources/config/dev/resource-config/workspace_manager_v1.yml
```diff
@@ -0,0 +1,28 @@
+# Workspace Manager buffered workspace template
+---
+configName: "workspace_manager_v1"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-dev"
```
yonghaoy, line 6: as we chatted yesterday, I am not sure what is project id perfix would be. Have you check with @gjuggler ? If we use terra-dev, it would be the same as the projects in `cwb_ws_resource_dev_v1`. I am not sure if we want to have different prefix for `classic terra` vs `McTerra`.

yonghaoy (approved):
LGTM! Just have a minor question about project id prefix.

src/main/resources/config/dev/resource-config/workspace_manager_v1.yml
yonghaoy: Found the doc: https://docs.google.com/spreadsheets/d/1pVECSx12nNirhF_Z5m9YWJj5BBCCXKk95WXmqQIyA2Y/edit?ts=5f7f670d#gid=1410828808
I guess then this should be `terra-dev` then?

tnadler: Yes, this was based on a conversation with @gjuggler ... we agreed terra-dev and terra-test. Look OK?

gjuggler: Ah, I hadn't realized that CWB already uses the "terra-dev" prefix. It is pretty awkward to have two different pools with the same prefix in the same RBS (in prod we won't have this issue, since enterprise Terra will be in totally distinct cloud space).

Maybe terra-wsm-dev? That would make these pretty hard to mistake.

tnadler: Ah sorry I missed this, will send a follow on.

PR #84, opened 2020-12-17
[PF-287] Add retry rules for all steps

yonghaoy:
Those three are all simple changes.


src/main/java/bio/terra/buffer/service/resource/flight/StepUtils.java
```diff
@@ -8,9 +8,13 @@
 import bio.terra.buffer.service.resource.FlightMapKeys;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.RetryRuleFixedInterval;
 
 /** Utilities used in Stairway steps. */
 public class StepUtils {
+  /** The stairway retry rule used in all resource creation/deletion steps. */
+  public static final RetryRuleFixedInterval RETRY_RULE =
+      new RetryRuleFixedInterval(/* intervalSeconds =*/ 180, /* maxCount =*/ 5);
```
wchamber, line 17: maybe `FIXED_RETRY_RULE` or `DEFAULT_RETRY_RULE`?
I kind of disagree that the same retry rule works for every step. We don't really need to wait 3 minutes to retry a database transaction error. Let's be more granular and think about what the right retries are for different steps.

wchamber (changes requested):

src/main/java/bio/terra/buffer/service/resource/flight/StepUtils.java
```diff
@@ -8,9 +8,17 @@
...
 import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.RetryRuleFixedInterval;
 
 /** Utilities used in Stairway steps. */
 public class StepUtils {
+  /** The stairway retry rule for GCP API calls(retry every 3 minutes). */
+  public static final RetryRuleFixedInterval CLOUD_API_DEFAULT_RETRY_RULE =
+      new RetryRuleFixedInterval(/* intervalSeconds =*/ 180, /* maxCount =*/ 5);
+
+  /** The stairway retry rule for Buffer service internal operations(retry every 5 seconds). */
```
wchamber, line 19: rather than have the javadoc describe how often the retry happens, consider explaining why we might want to wait a short or long amount of time.
I think there might even be variations for different GCP API calls - those with quotas that refill by the minute make sense to wait more than a minute. Other gcp api calls, like setting permissions or creating buckets are probably fine to retry in seconds.

src/main/java/bio/terra/buffer/service/resource/flight/StepUtils.java
```diff
@@ -8,9 +8,17 @@
...
+  public static final RetryRuleFixedInterval INTERNAL_DEFAULT_RETRY_RULE =
```
wchamber, line 20: optional nit: i think it's fine to just name these `INTERNAL_DEFAULT_RETRY`. don't think `_RULE` adds much clarity for its extra 5 chars.

wchamber (approved):

src/main/java/bio/terra/buffer/service/resource/flight/StepUtils.java
yonghaoy: 1: Done
2: Making the cloud call usually longer is safter, and not just for quota issue, also some unknown slow/outrage in cloud side. So I use 3 minutes and think that won't hurt. Actually 3min is little bit too long. Changed to 90 seconds.

PR #85, opened 2020-12-17
[PF-287] Improve Stairway setup in Buffer service

yonghaoy:
Background: Each project creation is a flight and it block a thread for 10~15 minutes. We need more threads to be able to fill the pool sooner.
Things in the PR:
1.  Set stairway name - because without the name every-time we restart Buffer server it will create a new stairway name, and we can not recover the previous one. 
2. Add more threads for stairway - We need more threads to make is faster to fill a pool. 
3. Don't swipe flight log when flight finished - Flightlog is useful for debug purpose.

### DIFF
--- a/src/main/java/bio/terra/buffer/service/stairway/StairwayComponent.java
+++ b/src/main/java/bio/terra/buffer/service/stairway/StairwayComponent.java
@@ -48,6 +48,7 @@ public class StairwayComponent {
         Stairway.newBuilder()
             .maxParallelFlights(stairwayConfiguration.getMaxParallelFlights())
             .applicationContext(applicationContext)
+            .keepFlightLog(true)
             .stairwayName(stairwayConfiguration.getName())
             .stairwayClusterName(stairwayConfiguration.getClusterName());
     try {
--- a/src/main/java/bio/terra/buffer/service/stairway/StairwayComponent.java
+++ b/src/main/java/bio/terra/buffer/service/stairway/StairwayComponent.java
@@ -44,6 +44,7 @@ public class StairwayComponent {
         stairwayConfiguration.getClusterName(),
         stairwayConfiguration.getClusterName());
     // TODO(PF-161): Configure the workqueue pubsub subscription and topic for multi-instance.
+    // TODO(PF-314): Cleanup old flightlogs.
     Stairway.Builder builder =
         Stairway.newBuilder()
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/buffer/service/stairway/StairwayComponent.java
```diff
@@ -48,6 +48,7 @@ public StairwayComponent(
         Stairway.newBuilder()
             .maxParallelFlights(stairwayConfiguration.getMaxParallelFlights())
             .applicationContext(applicationContext)
+            .keepFlightLog(true)
```
wchamber, line 51: we're eventually going to want some cleanup of the flight log & flight history. We probably don't want millions of entries in our stairway db. Does stairway have a feature for that today? If not, can we file & TODO a ticket to put in the backlog?

src/main/resources/application.yml
```diff
@@ -20,8 +20,11 @@ buffer:
       uri: jdbc:postgresql://127.0.0.1:5432/${BUFFER_STAIRWAY_DATABASE_NAME}
       username:
     force-clean-start: false
-    max-parallel-flights: 80
+    # This decides how soon we can fill a pool. Example: if total pools size is 3000 and each flight takes 10 minues.
+    # having 500 hears means it takes 3000 resources / 500 threads * 10minute = 60 minutes to create this pool
+    max-parallel-flights: 500
     migrate-upgrade: true
+    name: buffer-stairway
```
wchamber, line 27: I think it'd be better to make this the default value on the config var definition, that way it's really hard to not have a name. Also leave a comment there on why it's important to name the stairway instance, and when you would want to have multiple names (multiple instances).
Would you mind checking Janitor & WM and filing tickets for them if they're also using a null stairway name?

wchamber (approved):
Please include in the PR description why it's important to set the stairway name and what happens when you don't.

src/main/resources/application.yml
wchamber: Oops, see you're already doing this for Janitor. thanks!

src/main/java/bio/terra/buffer/service/stairway/StairwayComponent.java
yonghaoy: Add a todo and ticket.

PR #87, opened 2020-12-18
[PF-289] Pull resource state update logic out from flight 

yonghaoy:
Pervasively we update resource state(CREATING/DELETING) inside flight. And there is no way to tell scheduler how many flights already in the queue. So we end up schedule way more than it should be. So moving it out from flight can prevent that.


src/main/java/bio/terra/buffer/db/BufferDao.java
```diff
@@ -261,14 +261,15 @@ public boolean updateResourceAsHandedOut(ResourceId id, RequestHandoutId request
...
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
-  public boolean updateResourceAsDeleting(ResourceId id) {
-    String sql = "UPDATE resource SET state = :state WHERE id = :id";
+  public boolean updateReadyResourceToDeleting(ResourceId id) {
+    String sql = "UPDATE resource SET state = :state WHERE id = :id AND state = :currentState";
 
     MapSqlParameterSource params =
         new MapSqlParameterSource()
             .addValue("state", ResourceState.DELETING.toString())
+            .addValue("currentState", ResourceState.READY.toString())
```
wchamber, line 272: just inline STATE and currentState?

src/main/java/bio/terra/buffer/db/BufferDao.java
```diff
@@ -261,14 +261,15 @@ public boolean updateResourceAsHandedOut(ResourceId id, RequestHandoutId request
     return jdbcTemplate.update(sql, params) == 1;
   }
 
-  /** Updates resource state to DELETING. */
+  /** Updates resource in READY state to DELETING. */
```
wchamber, line 264: describe what the boolean return value represents

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -1,42 +1,86 @@
...
-        flightSubmissionFactory.getDeletionFlightSubmission(resource, resourceType));
+    return transactionTemplate.execute(
+        status -> updateResourceAsDeletingAndSubmitFlight(resource, resourceType, status));
+  }
+
+  /**
+   * Create entity in resource table with CREATING and submit creation flight.
+   *
+   * <p>This should be done as a part of a transaction. The TransactionStatus is unused, but a part
+   * of the signature as a reminder.
```
wchamber, line 56: describe *why* it's important to do this as a transaction.

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -1,42 +1,86 @@
...
+            .state(ResourceState.CREATING)
+            .build());
+    return submitToStairway(flightSubmissionFactory.getCreationFlightSubmission(pool, resourceId));
+  }
+
+  /**
+   * Update a READY resource state to DELETING and submit deletion flight.
+   *
+   * <p>This should be done as a part of a transaction. The TransactionStatus is unused, but a part
+   * of the signature as a reminder.
```
wchamber, line 75: describe why

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -1,42 +1,86 @@
...
+   */
+  private Optional<String> updateResourceAsDeletingAndSubmitFlight(
+      Resource resource, ResourceType resourceType, TransactionStatus unused) {
+    if (bufferDao.updateReadyResourceToDeleting(resource.id())) {
```
wchamber, line 79: do we want to log something if the resource wasn't READY? Are there some states that are more interesting about the Resource than others?

src/test/java/bio/terra/buffer/integration/IntegrationUtils.java
```diff
@@ -66,14 +67,23 @@
...
-  public static Optional<FlightMap> blockUntilFlightComplete(
+  /**
+   * Poll until flight finished, and return ResourceId back from input map.
+   *
+   * <p>Stairway convert all input parameter type to String when using FlightState. So we need extra
+   * step of converting from String to UUID to ResourceId. See <a
+   * href="https://broadworkbench.atlassian.net/browse/PF-316">PF-316</a>
+   */
+  public static Optional<ResourceId> blockUntilFlightComplete(
       StairwayComponent stairwayComponent, String flightId)
```
wchamber, line 78: this is kind of awkward. We're taking a very general flight id, and assuming it has the resource id. The resource id also doesn't have a lot to do with waiting for the result.
Let's return the entire FlightState from this function. If we want, we can add a helper for getting the ResourceId from the flight state/input parameters.

wchamber (changes requested):
few nits, thanks!

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
yonghaoy: added some log. If we don't have other bug, only state it may change to is HANDED_OUT. It is possible that a resource is just handed out after scheduler scans before the code here.

wchamber: "if we don't have other bug." If we could verify that in this code, it would help us catch such bugs earlier. e.g. return the previous status from updateReadyResourceToDeleting.

yonghaoy:
I would there are two files I didn't push last time after I run integration. I added two initial steps with undo method.
Creating Resource: Delete the entity from DB
Deleting: Return FATAL flight

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
yonghaoy: just saw this. Working on addressing this comment. Will ping you once I address this

yonghaoy: Tried few ways. Choose to log inside the BufferDao. Return previous state is more complicated than that because we will have to add few more if else blocks to see when to return what. And there are some wired case makes it more complicated, e.g., resource is READY, but jddb.update fail, or resource is null, return `optional.empty`

Anyway, the current logging in BufferDao should be enough and I like this approach.

src/main/java/bio/terra/buffer/db/BufferDao.java
```diff
@@ -261,17 +266,29 @@ public boolean updateResourceAsHandedOut(ResourceId id, RequestHandoutId request
...
-            .addValue("state", ResourceState.DELETING.toString())
-            .addValue("currentState", ResourceState.READY.toString())
-            .addValue("id", id.id());
-    return jdbcTemplate.update(sql, params) == 1;
+    String readSql =
+        "select id, pool_id, creation, handout_time, state, request_handout_id, cloud_resource_uid, deletion "
+            + "FROM resource WHERE id = :id ";
+    MapSqlParameterSource readParams = new MapSqlParameterSource().addValue("id", id.id());
+    Resource resource =
+        DataAccessUtils.singleResult(jdbcTemplate.query(readSql, readParams, RESOURCE_ROW_MAPPER));
```
wchamber, line 280: Why can't we use `retrieveReosurce`? transactions?

src/main/java/bio/terra/buffer/service/resource/flight/InitialResourceCreationStep.java
```diff
@@ -0,0 +1,37 @@
...
+import bio.terra.buffer.db.BufferDao;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+
+/**
+ * Initial resource creation step. Now we create CREATING resource before the flight, but we need to
+ * delete it if flight fails.
+ */
+public class InitialResourceCreationStep implements Step {
```
wchamber: Refering to 'now' vs how thing were done previously in code isn't usually very helpful for future maintainers. Instead, prefer to describe what's going on now & why (with 'now' being implicit').
Consider renaming this as something other than 'InitialResourceCreationStep' which just says it should go first but doesn't describe what this step actually does. Naming is hard :(

src/main/java/bio/terra/buffer/service/resource/flight/InitialResourceCreationStep.java
```diff
@@ -0,0 +1,37 @@
...
+  private final BufferDao bufferDao;
+
+  public InitialResourceCreationStep(BufferDao bufferDao) {
+    this.bufferDao = bufferDao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) {
+    return StepResult.getStepResultSuccess();
```
wchamber, line 24: add a tiny comment noting that this does nothing on purpose

src/main/java/bio/terra/buffer/service/resource/flight/InitialResourceDeletionStep.java
```diff
@@ -0,0 +1,26 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+
+/**
+ * Initial resource deletion step. Now we mark resource as DELETING before the flight, but we need
+ * to add extra step to handle the rollback, i.e. return STEP_RESULT_FAILURE_FATAL when rollback.
+ */
+public class InitialResourceDeletionStep implements Step {
```
wchamber: dittos

src/main/java/bio/terra/buffer/service/resource/flight/InitialResourceDeletionStep.java
```diff
@@ -0,0 +1,26 @@
...
+  @Override
+  public StepResult doStep(FlightContext flightContext) {
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) {
+    // Rolling back resource state READY requires more complicated workflow but we think this is
+    // rare and unnecessary. We would just fail the flight.
+    return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL);
```
wchamber: Do we really need this step? If we go straight to project deletion step, and try to undo that, that step should fatally fail. So what does this step get us?

wchamber (approved):

src/main/java/bio/terra/buffer/service/resource/flight/InitialResourceDeletionStep.java
yonghaoy: ah, right, I thought this is some `extra level protection`. But I guess we don't need it

src/main/java/bio/terra/buffer/service/resource/flight/InitialResourceCreationStep.java
yonghaoy: add more things in javadoc. Not sure if I understand your now vs how clearly....
I thought my previous comment(Now we create CREATING resource before the flight, but we need to
 * delete it if flight fails.) is trying to 'describe what's going on now & why (with 'now' being implicit')' 
 Which part I am missing? 

Also, rename the file, it is really really hard to come up a short&meaningfull name and I can not get a CamelCase badge..

PR #89, opened 2020-12-21
[no ticket]Add more tips in readme

yonghaoy:
I found this is helpful and saves my time


gjuggler (approved):
LGTM. For reference, [here](https://github.com/all-of-us/workbench/blob/master/api/libproject/devstart.rb#L2126) is AoU's script-wrapping approach to bootstrapping the SQL proxy connection, with or without read-only mode. Not necessarily a near-term need, but it does help make it easier for the whole team to diagnose issues in the dev environment.

PR #90, opened 2020-12-22
[PF-319] Pick random resource for handout request. And make this DB query transactional with retry

yonghaoy:
Previously we always pick the top element from resource table, and get/update are in not handled in the same transcation. 
It causes issue that `800 of 1000` requests return empty in perf test. 
I tried the new query and get this result: 
For pool size of 5000:
Results:
Making 1000 concurrent requests, all success
Making 3000 concurrent requests, 5 failed
Making 5000 concurrent requests, 11 failed

After I add retry in test runner side and Buffer server code, seems all can success even for 5000/5000 cases. 

This seems to be a working solution and meet production requirement because in production world, the actual rate is **< 1000/day,** and **< 150/hour** 

More details can be found at: https://docs.google.com/document/d/1FpfjGrjS3t_x2gfPe2CfQsN1vpMT_GiQ0OJrgmZpiDI/edit?resourcekey=0-uDux8J3Jg12lzkbQ2UELWQ#


src/main/java/bio/terra/buffer/db/BufferDao.java
```diff
@@ -248,22 +249,55 @@ public boolean updateResourceAsReady(ResourceId id, CloudResourceUid resourceUid
...
+        MapSqlParameterSource params =
+            new MapSqlParameterSource()
+                .addValue("state", ResourceState.HANDED_OUT.toString())
+                .addValue("request_handout_id", requestHandoutId.id())
+                .addValue("handout_time", OffsetDateTime.now(ZoneOffset.UTC))
+                .addValue("id", selectedResource.id().id());
+
+        return jdbcTemplate.update(sql, params) == 1
+            ? Optional.of(selectedResource)
+            : Optional.empty();
```
marikomedlock, line 298: What does it mean when this update doesn't return 1? I think it means that there were 2 requests that picked the same ready resource. One of the requests successfully updated the state and request_handout_id, and the other request's update here fails. Is that right? Regardless, a comment here explaining the situation would be very helpful.

src/main/java/bio/terra/buffer/db/BufferDao.java
```diff
@@ -248,22 +249,55 @@ public boolean updateResourceAsReady(ResourceId id, CloudResourceUid resourceUid
...
+      List<Resource> resources = retrieveResourcesRandomly(poolId, ResourceState.READY, 1);
+      if (resources.size() == 0) {
+        logger.warn("No resource is ready to use at this moment for pool: {}.", poolId);
+        return Optional.empty();
+      } else {
+        Resource selectedResource = resources.get(0);
+        String sql =
+            "UPDATE resource "
+                + "SET state = :state, request_handout_id = :request_handout_id, handout_time = :handout_time"
+                + " WHERE id = :id AND request_handout_id IS null";
```
marikomedlock, line 287: Do you also want to have state=READY in the where clause here? I'm not sure if it's necessary or not -- I think only if it's possible for a READY resource to have its state changed without having its request_handout_id set.

src/main/java/bio/terra/buffer/service/pool/PoolService.java
```diff
@@ -90,37 +92,28 @@ public PoolInfo getPoolInfo(PoolId poolId) {
...
     }
-    if (existingResource.isPresent()) {
-      if (existingResource.get().state().equals(ResourceState.HANDED_OUT)) {
-        return existingResource.get();
+    try {
+      Optional<Resource> resource =
+          executeAndRetry(
+              () -> bufferDao.updateOneReadyResourceToHandedOut(poolId, requestHandoutId),
+              Duration.ofSeconds(2),
+              10);
```
marikomedlock, line 104: This means to do 10 retries with a timeout of 2 seconds each? Please add a comment here.

src/main/java/bio/terra/buffer/service/pool/PoolService.java
```diff
@@ -90,37 +92,28 @@ public PoolInfo getPoolInfo(PoolId poolId) {
 
   private Resource handoutResourceTransactionally(
       PoolId poolId, RequestHandoutId requestHandoutId, TransactionStatus unused) {
```
marikomedlock, line 94: Nit: add method comment. In particular, explaining what you mean by "transactionally" I think would be helpful.

src/test/java/bio/terra/buffer/db/BufferDaoTest.java
```diff
@@ -116,7 +117,7 @@ public void retrievePoolWithResourceState() {
     bufferDao.createResource(newResource(pool1.id(), ResourceState.READY));
     bufferDao.createResource(newResource(pool1.id(), ResourceState.READY));
     bufferDao.createResource(newResource(pool2.id(), ResourceState.READY));
-    bufferDao.createResource(newResource(pool2.id(), ResourceState.HANDED_OUT));
+    bufferDao.createResource(newResource(pool2.id(), HANDED_OUT));
```
marikomedlock, line 120: Nit: Don't import the static here, so it matches the above lines.

src/test/java/bio/terra/buffer/service/resource/FlightSchedulerTest.java
```diff
@@ -178,17 +185,20 @@ public void scheduleDeactivationFlights_smallerLimitPerExecute() throws Exceptio
...
     Thread.sleep(4000);
 
-    resources.forEach(
-        resource ->
-            verify(flightManager).submitDeletionFlight(resource, ResourceType.GOOGLE_PROJECT));
+    verify(flightManager, times(3))
+        .submitDeletionFlight(resourceArgumentCaptor.capture(), eq(ResourceType.GOOGLE_PROJECT));
+
+    assertThat(
+        resources, (Matcher) Matchers.hasItems(resourceArgumentCaptor.getAllValues().toArray()));
```
marikomedlock, line 201: I don't follow what you are testing here. You start with a pool with 5 resources, 4 of which are ready. Then you retrieve all 4 ready resources. Then you are checking that ...?

src/main/java/bio/terra/buffer/db/BufferDao.java
```diff
@@ -248,22 +249,55 @@ public boolean updateResourceAsReady(ResourceId id, CloudResourceUid resourceUid
...
-            .addValue("handout_time", OffsetDateTime.now(ZoneOffset.UTC))
-            .addValue("id", id.id());
-    return jdbcTemplate.update(sql, params) == 1;
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public Optional<Resource> updateOneReadyResourceToHandedOut(
+      PoolId poolId, RequestHandoutId requestHandoutId) {
+    Optional<Resource> existingResource = retrieveResource(poolId, requestHandoutId);
+    if (existingResource.isPresent()) {
+      if (existingResource.get().state().equals(ResourceState.HANDED_OUT)) {
+        return existingResource;
```
marikomedlock, line 268: Nit: Add a log statement to this block indicating that we already handed this resource out.

Also, under what conditions would the same request try to update a single resource to ready more than once -- is this to support Stairway retries maybe?

src/test/java/bio/terra/buffer/db/BufferDaoTest.java
```diff
@@ -218,12 +236,12 @@ public void updateReadyResourceAsDeleting_success() {
   @Test
   public void updateReadyResourceAsDeleting_currentStateIsNotReady() {
     Pool pool = newPool(PoolId.create("poolId"));
-    Resource resource = newResource(pool.id(), ResourceState.HANDED_OUT);
+    Resource resource = newResource(pool.id(), HANDED_OUT);
```
marikomedlock, line 239: Nit: Don't import the static to match elsewhere in this file.

src/main/java/bio/terra/buffer/db/BufferDao.java
```diff
@@ -248,22 +249,55 @@ public boolean updateResourceAsReady(ResourceId id, CloudResourceUid resourceUid
...
-   * client.
+   * Pick one READY resource and handed it out to client. The steps are:
+   *
+   * <ul>
+   *   <li>Step 1: Checks if any resource uses this {@link RequestHandoutId}, if yes, return the
+   *       resource.
+   *   <li>Step 2: Randomly pick a READY entity from resource table
+   *   <li>Step 3: Update this resource state to HANDED_OUT
+   *   <li>Step 4: Return step 2's resource.
+   * </ul>
```
marikomedlock, line 260: Great summary here thanks, also I like the new method name better.

marikomedlock (changes requested):
Comments/questions inline.

src/main/java/bio/terra/buffer/db/BufferDao.java
yonghaoy: The sql above has `where id==.` make sure there will be only 1 get updated as most. 
The value here is more 1 or 0(if anything cause update failed somehow). I just use Optional instead of null to control the workflow. That is common practice in Buffer&Janitor to use 
`jdbcTemplate.update(sql, params) == 1` in every update/delete single row statement. Added a comment.

src/main/java/bio/terra/buffer/db/BufferDao.java
yonghaoy: Added. Just to confirm: The resource is retrieved from `retrieveResourcesRandomly(READY)` above.  The whole method is wrapped 
`@Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)`
Does that means the to be updated resource is locked in READY state?

src/main/java/bio/terra/buffer/service/pool/PoolService.java
yonghaoy: added

src/test/java/bio/terra/buffer/service/resource/FlightSchedulerTest.java
yonghaoy: the test behaviors doesn't change(4 are ready but scheduler should only delete), but we need to change the code to fix test breakage. 
The scheduler scans db and pick the X number ready resources it should delete. 
Previously, it always pick the top X resources(limit X in sql). So the previous test just pick the top 3 and verify it matches. 
Now it uses select randomResource(and it should). So I need to retrieve all 4 ready now, and verify 3 of them are deleted.

src/main/java/bio/terra/buffer/db/BufferDao.java
yonghaoy: Done. 
Retries happens in client side to make sure traffic is not dropped. Client caches the requestHandoutId and use that to retry if they didn't get resource back.

marikomedlock: Got it, thanks.

src/main/java/bio/terra/buffer/db/BufferDao.java
marikomedlock: Yes, I think that's right. But just seems safer since it's in two separate methods, in case we add new callers in the future.

marikomedlock (approved):
LGTM

PR #93, opened 2021-01-08
[no ticket] Increase pool size for tools and dev

yonghaoy:
The pool size is estimated from the qps of how many test are triggerd. 
AoU has a higher number because it seems to have more tests triggered


wchamber (approved):
Worth a description on *why* we're bumping these numbers in the PR description. And why they're not even bigger. Even if you don't have a hard reason for 100 vs 200, good to point that out for future maintainers.

PR #94, opened 2021-01-11
Enable Google LifeSciences API in addition to the Google Genomics API [BT-60]

grsterin:


mcovarr (approved):

kshakir (approved):

yonghaoy:
Hi @grsterin This is a breaking change as directly update pool config is not allowed. 
Can I understand the scenario of this? Also, do we need to backfill existing projects after this change? 
Thanks

grsterin:
> Hi @grsterin This is a breaking change as directly update pool config is not allowed.
> Can I understand the scenario of this? Also, do we need to backfill existing projects after this change?
> Thanks

Hi @yonghaoy. We need to change default GCP backend used by Cromwell in Terra from Genomics API v2alpha (aka PAPI v2alpha) to LifeSciences API v2beta  (aka PAPIv2beta). In order to do so we need to have LifeSciences API to be enabled in Google projects corresponding to workspaces.
We were going to do retrofitting of existing projects using [this script](https://github.com/broadinstitute/firecloud-develop/blob/7908df43e0c45aa97d201db0893231520d684e64/run-context/live/scripts/migrations/enable_lifesciences.sh) from another PR.

Here is the underlying ticket: https://broadworkbench.atlassian.net/browse/BT-60

Please let me know if there are any questions. Also if this PR is not the proper way to do what we want, what would you suggest to do instead?

yonghaoy:
> > Hi @grsterin This is a breaking change as directly update pool config is not allowed.
> > Can I understand the scenario of this? Also, do we need to backfill existing projects after this change?
> > Thanks
> 
> Hi @yonghaoy. We need to change default GCP backend used by Cromwell in Terra from Genomics API v2alpha (aka PAPI v2alpha) to LifeSciences API v2beta (aka PAPIv2beta). In order to do so we need to have LifeSciences API to be enabled in Google projects corresponding to workspaces.
> We were going to do retrofitting of existing projects using [this script](https://github.com/broadinstitute/firecloud-develop/blob/7908df43e0c45aa97d201db0893231520d684e64/run-context/live/scripts/migrations/enable_lifesciences.sh) from another PR.
> 
> Here is the underlying ticket: https://broadworkbench.atlassian.net/browse/BT-60
> 
> Please let me know if there are any questions. Also if this PR is not the proper way to do what we want, what would you suggest to do instead?

Hi @grsterin 
We don't allow directly modify the existing configuration because of backfill & backward compatibility concerns. 
https://github.com/DataBiosphere/terra-resource-buffer#upgrade-pool-configuration is the link of how to update pool config.  
Basically  we need to create a _v2 resourceconfig in each envs
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/config/dev/resource-config/aou_ws_resource_test_v1.yml#L3
Also also a _v2 pool that uses _v2 config 
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/config/dev/pool_schema.yml#L6

After that release, whenever you feel comfortable, just let clients(Rawls in our case) to switch to use `_v2 pool` Then you don't need to backfill projects in  _v1's pool

src/main/resources/config/dev/resource-config/aou_ws_resource_test_v1.yml
```diff
@@ -21,6 +21,7 @@ gcpProjectConfig:
     - "dataproc.googleapis.com"
     - "dns.googleapis.com"
     - "genomics.googleapis.com"
+    - "lifesciences.googleapis.com"
```
yonghaoy, line 24: Create v2 resource config in this file. ditto for others

yonghaoy (changes requested):

src/main/resources/config/dev/resource-config/aou_ws_resource_test_v1.yml
grsterin: Shouldn't it be in the new file like `aou_ws_resource_test_v2.yml`? Seems like the file name contains version in it too.

yonghaoy (approved):

mcovarr:
Hi @yonghaoy, I was curious how and exactly when we should do this part: 

> After that release, whenever you feel comfortable, just let clients(Rawls in our case) to switch to use _v2 pool

mcovarr (approved):

kshakir (commented):
You have two 👍 so I think you're good to go. FYI I'm abstaining from a review at the moment as I would like more context if I'm going to make similar changes to this repo in the future.

yonghaoy:
> Hi @yonghaoy, I was curious how and exactly when we should do this part:
> 
> > After that release, whenever you feel comfortable, just let clients(Rawls in our case) to switch to use _v2 pool

I can help you on that. (Basically I would just ping those two teams in their slack channel #dsp-cloud-integration and #platform-foundations)

PR #97, opened 2021-01-15
[no ticket]: TranscationTemplate should swallow exceptions

yonghaoy:
Context:  I saw there is a CREATING state resource stays in DB but there is no such record in stairway Flight. After digging into logs. I suspect TranscationTemplate doesn't propagate exceptions, and ignore exceptions being catched. 
So I made the PR to verify my thought.

If this looks reasonable, I will all delete _withTryCatch() methods then merge this PR.

Source code that might be useful: 
https://github.com/spring-projects/spring-framework/blob/master/spring-tx/src/main/java/org/springframework/transaction/support/TransactionTemplate.java#L140

### DIFF
--- a/src/main/java/bio/terra/buffer/service/resource/FlightManager.java
+++ b/src/main/java/bio/terra/buffer/service/resource/FlightManager.java
@@ -52,10 +52,9 @@ public class FlightManager {
   /**
    * Create entity in resource table with CREATING and submit creation flight.
    *
-   * <p>This should be done as a part of a transaction because we don't want resource state update
-   * without submitting a flight. It will still work when Stairway submission succeeds but DB update
-   * failed. Because in Stairway flight, we check DB state and will abort the flight if DB entity in
-   * bad state.
+   * <p>If the Stairway submission fails, the transaction will be rolled back. If the Stairway
+   * submission succeeds but the DB update transaction fails, the flight checks the DB state and
+   * aborts if the state is bad.
    */
   private Optional<String> createResourceEntityAndSubmitFlight(
       Pool pool, TransactionStatus status) {
@@ -74,10 +73,9 @@ public class FlightManager {
   /**
    * Update a READY resource state to DELETING and submit deletion flight.
    *
-   * <p>This should be done as a part of a transaction because we don't want resource state update
-   * without submitting a flight. It will still work when Stairway submission succeeds but DB update
-   * failed. Because in Stairway flight, we check DB state and will abort the flight if DB entity in
-   * bad state.
+   * <p>If the Stairway submission fails, the transaction will be rolled back. If the Stairway
+   * submission succeeds but the DB update transaction fails, the flight checks the DB state and
+   * aborts if the state is bad.
    */
   private Optional<String> updateResourceAsDeletingAndSubmitFlight(
       Resource resource, ResourceType resourceType, TransactionStatus status) {
@@ -98,9 +96,7 @@ public class FlightManager {
       return Optional.of(flightId);
     } catch (DatabaseOperationException | StairwayExecutionException | InterruptedException e) {
       logger.error("Error submitting flight id: {}", flightId, e);
-      // Let TranscationTemplate rollback the previous DB commit if failed to submit a flight. Here
-      // we try-catch
-      // checked exception and we need to manually set it up.
+      // If the flight submission fails, set the transaction to be rolled back.
       status.setRollbackOnly();
       return Optional.empty();
     }
--- a/src/main/java/bio/terra/buffer/service/resource/FlightManager.java
+++ b/src/main/java/bio/terra/buffer/service/resource/FlightManager.java
@@ -96,6 +96,8 @@ public class FlightManager {
       return Optional.of(flightId);
     } catch (DatabaseOperationException | StairwayExecutionException | InterruptedException e) {
       logger.error("Error submitting flight id: {}", flightId, e);
+      // Let TranscationTemplate rollback the previous DB commit if failed to submit a flight. Here we try-catch
+      // checked exception and we need to manually set it up.
       status.setRollbackOnly();
       return Optional.empty();
     }
--- a/src/main/java/bio/terra/buffer/service/resource/FlightManager.java
+++ b/src/main/java/bio/terra/buffer/service/resource/FlightManager.java
@@ -40,16 +40,10 @@ public class FlightManager {
 
   /** Submit Stairway Flight to create resource. */
   public Optional<String> submitCreationFlight(Pool pool) {
-    try {
-      return transactionTemplate.execute(
-          status -> createResourceEntityAndSubmitFlight(pool, status));
-    } catch (RuntimeException e) {
-      logger.error("Failed to submit creation flight for pool {}", pool.id(), e);
-      return Optional.empty();
-    }
+    return transactionTemplate.execute(status -> createResourceEntityAndSubmitFlight(pool, status));
   }
 
-  /** Submit Stairway Flight to create resource.  TODO: Remove before submit. */
+  /** Submit Stairway Flight to create resource. TODO: Remove before submit. */
   public Optional<String> submitCreationFlightWithTryCatch(Pool pool) {
     return transactionTemplate.execute(
         status -> createResourceEntityAndSubmitFlightWithTryCatch(pool, status));
@@ -57,13 +51,8 @@ public class FlightManager {
 
   /** Submit Stairway Flight to delete resource. */
   public Optional<String> submitDeletionFlight(Resource resource, ResourceType resourceType) {
-    try {
-      return transactionTemplate.execute(
-          status -> updateResourceAsDeletingAndSubmitFlight(resource, resourceType, status));
-    } catch (RuntimeException e) {
-      logger.error("Failed to submit deletion flight for resource {}", resource.id());
-      return Optional.empty();
-    }
+    return transactionTemplate.execute(
+        status -> updateResourceAsDeletingAndSubmitFlight(resource, resourceType, status));
   }
 
   /**
@@ -74,7 +63,7 @@ public class FlightManager {
    * reminder.
    */
   private Optional<String> createResourceEntityAndSubmitFlight(
-      Pool pool, TransactionStatus unused) {
+      Pool pool, TransactionStatus status) {
     ResourceId resourceId = ResourceId.create(UUID.randomUUID());
     bufferDao.createResource(
         Resource.builder()
@@ -83,7 +72,8 @@ public class FlightManager {
             .creation(Instant.now())
             .state(ResourceState.CREATING)
             .build());
-    return submitToStairway(flightSubmissionFactory.getCreationFlightSubmission(pool, resourceId));
+    return submitToStairway(
+        flightSubmissionFactory.getCreationFlightSubmission(pool, resourceId), status);
   }
 
   /**
@@ -93,10 +83,10 @@ public class FlightManager {
    * without submitting a flight. The TransactionStatus is unused, but a part of the signature as a
    * reminder.
    *
-   * TODO: Remove before submit.
+   * <p>TODO: Remove before submit.
    */
   private Optional<String> createResourceEntityAndSubmitFlightWithTryCatch(
-      Pool pool, TransactionStatus unused) {
+      Pool pool, TransactionStatus status) {
     ResourceId resourceId = ResourceId.create(UUID.randomUUID());
     bufferDao.createResource(
         Resource.builder()
@@ -106,7 +96,7 @@ public class FlightManager {
             .state(ResourceState.CREATING)
             .build());
     return submitToStairwayWithTryCatch(
-        flightSubmissionFactory.getCreationFlightSubmission(pool, resourceId));
+        flightSubmissionFactory.getCreationFlightSubmission(pool, resourceId), status);
   }
 
   /**
@@ -117,17 +107,17 @@ public class FlightManager {
    * reminder.
    */
   private Optional<String> updateResourceAsDeletingAndSubmitFlight(
-      Resource resource, ResourceType resourceType, TransactionStatus unused) {
+      Resource resource, ResourceType resourceType, TransactionStatus status) {
     if (bufferDao.updateReadyResourceToDeleting(resource.id())) {
       return submitToStairway(
-          flightSubmissionFactory.getDeletionFlightSubmission(resource, resourceType));
+          flightSubmissionFactory.getDeletionFlightSubmission(resource, resourceType), status);
     }
     logger.info("Failed to submit resource deletion flight for resource{}", resource.id());
     return Optional.empty();
   }
 
   private Optional<String> submitToStairway(
-      FlightSubmissionFactory.FlightSubmission flightSubmission) {
+      FlightSubmissionFactory.FlightSubmission flightSubmission, TransactionStatus status) {
     String flightId = stairway.createFlightId();
     try {
       stairway.submitToQueue(
@@ -135,13 +125,14 @@ public class FlightManager {
       return Optional.of(flightId);
     } catch (DatabaseOperationException | StairwayExecutionException | InterruptedException e) {
       logger.error("Error submitting flight id: {}", flightId, e);
-      throw new RuntimeException(String.format("Error submitting flight id: %s", flightId), e);
+      status.setRollbackOnly();
+      return Optional.empty();
     }
   }
 
   /** TODO: Remove before submit. */
   private Optional<String> submitToStairwayWithTryCatch(
-      FlightSubmissionFactory.FlightSubmission flightSubmission) {
+      FlightSubmissionFactory.FlightSubmission flightSubmission, TransactionStatus status) {
     String flightId = stairway.createFlightId();
     try {
       stairway.submitToQueue(
--- a/src/main/java/bio/terra/buffer/service/resource/FlightManager.java
+++ b/src/main/java/bio/terra/buffer/service/resource/FlightManager.java
@@ -44,7 +44,7 @@ public class FlightManager {
       return transactionTemplate.execute(
           status -> createResourceEntityAndSubmitFlight(pool, status));
     } catch (RuntimeException e) {
-      logger.error("Failed to submit creation flight for pool {}", pool.id());
+      logger.error("Failed to submit creation flight for pool {}", pool.id(), e);
       return Optional.empty();
     }
   }
--- a/src/main/java/bio/terra/buffer/service/resource/FlightManager.java
+++ b/src/main/java/bio/terra/buffer/service/resource/FlightManager.java
@@ -43,12 +43,6 @@ public class FlightManager {
     return transactionTemplate.execute(status -> createResourceEntityAndSubmitFlight(pool, status));
   }
 
-  /** Submit Stairway Flight to create resource. TODO: Remove before submit. */
-  public Optional<String> submitCreationFlightWithTryCatch(Pool pool) {
-    return transactionTemplate.execute(
-        status -> createResourceEntityAndSubmitFlightWithTryCatch(pool, status));
-  }
-
   /** Submit Stairway Flight to delete resource. */
   public Optional<String> submitDeletionFlight(Resource resource, ResourceType resourceType) {
     return transactionTemplate.execute(
@@ -73,29 +67,6 @@ public class FlightManager {
             .state(ResourceState.CREATING)
             .build());
     return submitToStairway(
-        flightSubmissionFactory.getCreationFlightSubmission(pool, resourceId), status);
-  }
-
-  /**
-   * Create entity in resource table with CREATING and submit creation flight.
-   *
-   * <p>This should be done as a part of a transaction because we don't want resource state update
-   * without submitting a flight. The TransactionStatus is unused, but a part of the signature as a
-   * reminder.
-   *
-   * <p>TODO: Remove before submit.
-   */
-  private Optional<String> createResourceEntityAndSubmitFlightWithTryCatch(
-      Pool pool, TransactionStatus status) {
-    ResourceId resourceId = ResourceId.create(UUID.randomUUID());
-    bufferDao.createResource(
-        Resource.builder()
-            .id(resourceId)
-            .poolId(pool.id())
-            .creation(Instant.now())
-            .state(ResourceState.CREATING)
-            .build());
-    return submitToStairwayWithTryCatch(
         flightSubmissionFactory.getCreationFlightSubmission(pool, resourceId), status);
   }
 
@@ -126,20 +97,6 @@ public class FlightManager {
     } catch (DatabaseOperationException | StairwayExecutionException | InterruptedException e) {
       logger.error("Error submitting flight id: {}", flightId, e);
       status.setRollbackOnly();
-      return Optional.empty();
-    }
-  }
-
-  /** TODO: Remove before submit. */
-  private Optional<String> submitToStairwayWithTryCatch(
-      FlightSubmissionFactory.FlightSubmission flightSubmission, TransactionStatus status) {
-    String flightId = stairway.createFlightId();
-    try {
-      stairway.submitToQueue(
-          flightId, flightSubmission.clazz(), flightSubmission.inputParameters());
-      return Optional.of(flightId);
-    } catch (DatabaseOperationException | StairwayExecutionException | InterruptedException e) {
-      logger.error("Error submitting flight id: {}", flightId, e);
       return Optional.empty();
     }
   }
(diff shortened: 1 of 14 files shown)


yonghaoy:
I don't think the codacy complains are reasonable..

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -89,6 +129,20 @@ public FlightManager(
   private Optional<String> submitToStairway(
       FlightSubmissionFactory.FlightSubmission flightSubmission) {
     String flightId = stairway.createFlightId();
+    try {
```
gjuggler, line 132: What is the purpose of this try-catch and re-throw? Is there any obvious benefit compared to just allowing the exception to bubble up directly to the transaction manager?

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -40,13 +40,30 @@ public FlightManager(
 
   /** Submit Stairway Flight to create resource. */
   public Optional<String> submitCreationFlight(Pool pool) {
-    return transactionTemplate.execute(status -> createResourceEntityAndSubmitFlight(pool, status));
+    try {
+      return transactionTemplate.execute(
+          status -> createResourceEntityAndSubmitFlight(pool, status));
+    } catch (RuntimeException e) {
```
gjuggler, line 46: Is it important to catch all exceptions here and return an empty value? At the very least, if we're catching all RuntimeExceptions we should log out the exception message before returning, so the stack traces show up in our logs.

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
yonghaoy: This method is one part of the scheduler service scheduling flight. We don't want scheduler to continue even one flight submit failed.

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
yonghaoy: This method is one part of the scheduler service scheduling flight. We don't want scheduler to continue even one flight submit failed.
Added logs

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
gjuggler: OK. To confirm my understanding then, the conversion of checked exceptions into a runtime exception is a key part of causing the txn to be appropriately rolled back? It's worth adding a comment to the re-throw line to make its role in this process explicit, to help future readers of this code know what's important and why.

gjuggler (approved):

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
wchamber: [TransactionTemplate#execute](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/support/TransactionTemplate.html#execute-org.springframework.transaction.support.TransactionCallback-)'s [TransactionCallback](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/support/TransactionCallback.html) cannot throw checked exceptions by its signature. So if we throw a checked exception in our code executed as the TransactionCallback, we must rethrow it as a runtime exception to compile. If we don't rethrow, then the exception is lost and the TransactionTemplate#execute never sees the exception.
Alternatively, we could set the [TransactionStatus#setRollback](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/TransactionExecution.html#setRollbackOnly--) when we catch the checked exception, which also notifies the TransactionTemplate that we must roll back.

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
wchamber: Please add a comment to that effect.

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -40,13 +40,30 @@ public FlightManager(
...
+      logger.error("Failed to submit creation flight for pool {}", pool.id(), e);
+      return Optional.empty();
+    }
+  }
+
+  /** Submit Stairway Flight to create resource.  TODO: Remove before submit. */
```
wchamber, line 52: TODO remove before submit?

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -40,13 +40,30 @@ public FlightManager(
...
   }
 
   /** Submit Stairway Flight to delete resource. */
   public Optional<String> submitDeletionFlight(Resource resource, ResourceType resourceType) {
-    return transactionTemplate.execute(
-        status -> updateResourceAsDeletingAndSubmitFlight(resource, resourceType, status));
+    try {
+      return transactionTemplate.execute(
+          status -> updateResourceAsDeletingAndSubmitFlight(resource, resourceType, status));
+    } catch (RuntimeException e) {
```
wchamber, line 63: ditto comment

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -89,6 +129,20 @@ public FlightManager(
...
+      stairway.submitToQueue(
+          flightId, flightSubmission.clazz(), flightSubmission.inputParameters());
+      return Optional.of(flightId);
+    } catch (DatabaseOperationException | StairwayExecutionException | InterruptedException e) {
+      logger.error("Error submitting flight id: {}", flightId, e);
+      throw new RuntimeException(String.format("Error submitting flight id: %s", flightId), e);
+    }
+  }
+
+  /** TODO: Remove before submit. */
```
wchamber, line 142: TODO remove before submit

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -69,6 +86,29 @@ public FlightManager(
...
+    ResourceId resourceId = ResourceId.create(UUID.randomUUID());
+    bufferDao.createResource(
+        Resource.builder()
+            .id(resourceId)
+            .poolId(pool.id())
+            .creation(Instant.now())
+            .state(ResourceState.CREATING)
+            .build());
+    return submitToStairwayWithTryCatch(
+        flightSubmissionFactory.getCreationFlightSubmission(pool, resourceId));
```
wchamber, line 109: If the flight submission has a try-catch, then we won't roll back the transaction if the flight submission fails. In that case, we'll have a CREATING resource sitting around without a Flight - effectively leaking the resource I think?

Oh, it is leaking, and the test verifies that... Am I supposed to be seeing with the two ways, that one doesn't work and one does? I'm ready to review the code that we're trying to submit.

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -69,6 +86,29 @@ public FlightManager(
...
   }
 
+  /**
+   * Create entity in resource table with CREATING and submit creation flight.
+   *
+   * <p>This should be done as a part of a transaction because we don't want resource state update
+   * without submitting a flight. The TransactionStatus is unused, but a part of the signature as a
+   * reminder.
+   *
+   * TODO: Remove before submit.
```
wchamber, line 96: these TODOs are confusing me.

src/main/java/bio/terra/buffer/service/resource/flight/StepUtils.java
```diff
@@ -17,14 +17,14 @@
    * outrage, and other unknown issues.
    */
   public static final RetryRuleFixedInterval CLOUD_API_DEFAULT_RETRY =
-      new RetryRuleFixedInterval(/* intervalSeconds =*/ 90, /* maxCount =*/ 5);
+      new RetryRuleFixedInterval(/* intervalSeconds =*/ 60, /* maxCount =*/ 10);
```
wchamber, line 20: why are we messing with the cloud api  & internal retry in this PR?

wchamber (changes requested):

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -69,6 +86,29 @@ public FlightManager(
     return submitToStairway(flightSubmissionFactory.getCreationFlightSubmission(pool, resourceId));
   }
 
+  /**
```
wchamber, line 79: (This should be on lines 76-87. I can't comment on code not in the PR - thanks github.)

I don't think it's correct to create the resource & submit to stairway within a single transaction. Can the transaction commit fail after a successful submission to Stairway? If it does, we'll rollback the created resouce in the dao, but we have no way of rolling back the Stairway submission.

wchamber: @gjuggler I think this is why we had the Janitor pattern of not including calls to Stairway within the database transaction. I've changed my mind that I think this is ok.

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
yonghaoy: All the TODO _withTryCatch is to demonstrate my thought(see my PR description).
After you and Greg reviewed the old leaking version with test, now I am removing those

src/main/java/bio/terra/buffer/service/resource/flight/StepUtils.java
yonghaoy: sorry, that was messed up from another PR

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
wchamber: Oops. Sorry. From the description "After digging into logs. I suspect TranscationTemplate doesn't propagate exceptions, and ignore exceptions being catched." I don't know what you mean by this. It doesn't magically propagate exceptions that don't bubble up to it? If the exception is caught within the TransactionCallback, then yes, it will not cause a rollback.
Also, we can't rollback stairway submissions with a TransactionTemplate - TransactionTemplate doesn't know anything about Stairway.

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
yonghaoy: Yup, I just saw the 2nd approach from an article. And I think that would be better, Switch to the new way

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
yonghaoy: removed

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
yonghaoy: I think submittToQueue is another postgres db commit call eventually, and I think it should able to rollback. 
And the flight will get rollback if resource in a wrong state.
Currently the checks happens as the last step of the flight (when we update state from DB). 
I can add a step at beginning of the flight to check the state, and let Flight fail earlier.

wchamber: Ah, I forgot we were trying to handle this in the flight itself.
What happens if the flight fails fatally? Then we don't have a DB entry, but we may have leaked a project. I guess we have to go and manually debug all fatal flights?

It's very much worth some comments about what's going on here.

wchamber: I'm trying to remember all the discussion we had here with Dan. I think this pattern of submitting to stairway in the transaction & having the flight rollback if the resource didn't get created came out of a doc you wrote Yonghao. The alternative was the complicated Janitor logic of monitoring for resources that were not successfully submitted or finished. Do you know where all that discussion lived?
Do we have tests for the case where a creation/deletion flight runs when the transaction kicking it off failed?

yonghaoy: Just added two steps at the begining of the flight to assert resource state correct. 
There is not too much we can do if flight fails fatally. The current rollback was two steps : 
1. Delete project
2. Delete DB
 The only thing we can do is to investigate which steps it fails, and manually cleanup things(it hasn't happen to me so far).

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
yonghaoy: (1) I was just trying to say I saw transcation was leaked
(2) Feels that is a general problems of using TransactionTemplate, we don't really know how the sub-steam libraries we use implemented. But we use different libraries all the time. 
And to me using TransactionTemplate was more like to make sure our own db get rollbacked if something went wrong. 
And actually, I suspect Stairway can rollback as the code to TransactionTemplate is just another DB(in stairway DB) commit eventually

wchamber: 2) In general, you should not do anything other than query your db during a transaction. No API calls, no calls to other services, including Stairway. You can make exceptions, but carefully.
Today, Stairway cannot roll back as a part of the Spring TransactionTemplate. Stairway is connecting to a different database and isn't using any of the Spring magic. I'ts not obvious to me we'd want to support allowing its modifications to be undone as a part of a transaction.
We can verify this is the case:
```
String flightId = "foo"
transactionTemplate.execute(status -> { stairway.submitFlight(...); throw new RuntimeException(); });
stairway.getFlight(flightId);  // Does the flight submission exist even though the transaction was rolled back?
```

src/main/java/bio/terra/buffer/service/resource/flight/AssertResourceCreatingInDbStep.java
```diff
@@ -0,0 +1,44 @@
...
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Checks if resource exists and in DELETING state before the actual deletion. It may happen when
+ * submitting flight success but updating DB fails.
```
wchamber, line 17: there's a race here, which may cause us to leave a resource in CREATING. Nothing forces the transaction around submission to finish before this step executes.
1. Transaction begins
2. Create resource in CREATING (pending transaction completeness).
3. Submit Flight successfully.
4. Flight begins executing.
5. This step executes, causes the flight to fail fatally as the resource does not exist yet.
6. Transaction completes, creating the resource with state CREATING.

Now there's a race between 5 & 6. If 4 executes first, the resource is stuck in CREATING with a fatal flight. Even worse, the race is likely to cause one of these transactions to fail since they're trying to read the same row. In general, it's not safe to use transactions with Stairway. Pragmatically, we can retry this step a few times. So that step execution always happens at least once after transaction completes.
This is also the case for DELETING. Working with transactions & Stairway is very tricky.

src/main/java/bio/terra/buffer/service/resource/flight/AssertResourceCreatingInDbStep.java
```diff
@@ -0,0 +1,44 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Checks if resource exists and in DELETING state before the actual deletion. It may happen when
```
wchamber, line 16: s/DELETING/CREATING ?

src/main/java/bio/terra/buffer/service/resource/flight/AssertResourceCreatingInDbStep.java
```diff
@@ -0,0 +1,44 @@
...
+ * submitting flight success but updating DB fails.
+ */
+public class AssertResourceCreatingInDbStep implements Step {
```
wchamber, line 19: optional nit: I'd be fine removing `InDb` from both these class names.

src/main/java/bio/terra/buffer/service/resource/flight/AssertResourceDeletingInDbStep.java
```diff
@@ -0,0 +1,44 @@
...
+
+  private final BufferDao bufferDao;
+
+  public AssertResourceDeletingInDbStep(BufferDao bufferDao) {
+    this.bufferDao = bufferDao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext) {
+    // Do nothing. We just use this step's undo method.
```
wchamber, line 30: ditto

src/main/java/bio/terra/buffer/service/resource/flight/AssertResourceCreatingInDbStep.java
```diff
@@ -0,0 +1,44 @@
+package bio.terra.buffer.service.resource.flight;
```
wchamber, line 1: I think there should be a test exercising these assertion steps for each of the flights.

wchamber (changes requested):

src/main/java/bio/terra/buffer/service/resource/flight/AssertResourceCreatingInDbStep.java
yonghaoy: Oh, there is the race. Thanks! I will let the step return error and retry on it.

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
yonghaoy: Yeah, we need to have the extra check step like I the one I added

src/main/java/bio/terra/buffer/service/resource/flight/AssertResourceCreatingInDbStep.java
yonghaoy: > I think there should be a test exercising these assertion steps for each of the flights.

Currently we have Flight integration tests that touches all steps of the flight. Feel that is enough for a success case. 
For failure, do we want to exercise failure test, I feel this logic is straight forward so I didn't add this. Adding test for a failure requires requires 200> lines change, and I don't think that worth the effort.

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -96,7 +98,8 @@ public FlightManager(
       return Optional.of(flightId);
     } catch (DatabaseOperationException | StairwayExecutionException | InterruptedException e) {
       logger.error("Error submitting flight id: {}", flightId, e);
-      // Let TranscationTemplate rollback the previous DB commit if failed to submit a flight. Here we try-catch
+      // Let TranscationTemplate rollback the previous DB commit if failed to submit a flight. Here
+      // we try-catch
       // checked exception and we need to manually set it up.
```
wchamber: nit: `// If the flight submission fails, set the transaction to be rolled back.`

src/main/java/bio/terra/buffer/service/resource/flight/AssertResourceCreatingStep.java
```diff
@@ -13,28 +13,30 @@
...
     Optional<Resource> resource =
         bufferDao.retrieveResource(ResourceId.retrieve(flightContext.getInputParameters()));
     if (resource.isPresent() && resource.get().state().equals(ResourceState.CREATING)) {
       return StepResult.getStepResultSuccess();
     }
     logger.warn("Resource {} does not exist or not in CREATING state", resource);
-    return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL);
+    // Retry this steps to avoid the potential race that "submitting Flight and update DB"
+    // transaction does not complete
+    // but the flight begins and it reaches here.
```
wchamber, line 38: nit: fix comment whitespace

src/main/java/bio/terra/buffer/service/resource/flight/AssertResourceDeletingStep.java
```diff
@@ -16,24 +16,26 @@
...
-    // Do nothing. We just use this step's undo method.
     Optional<Resource> resource =
         bufferDao.retrieveResource(ResourceId.retrieve(flightContext.getInputParameters()));
     if (resource.isPresent() && resource.get().state().equals(ResourceState.DELETING)) {
       return StepResult.getStepResultSuccess();
     }
     logger.warn("Resource {} does not exist or not in DELETING state", resource);
+    // Retry this steps to avoid the potential race that "submitting Flight and update DB"
+    // transaction does not complete
+    // but the flight begins and it reaches here.
```
wchamber, line 38: nit: ditto wwhiteespace

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -74,8 +75,9 @@ public FlightManager(
    * Update a READY resource state to DELETING and submit deletion flight.
    *
    * <p>This should be done as a part of a transaction because we don't want resource state update
-   * without submitting a flight. The TransactionStatus is unused, but a part of the signature as a
-   * reminder.
+   * without submitting a flight. It will still work when Stairway submission succeeds but DB update
+   * failed. Because in Stairway flight, we check DB state and will abort the flight if DB entity in
+   * bad state.
```
wchamber, line 80: nit: ditto wordsmithing

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -53,8 +53,9 @@ public FlightManager(
    * Create entity in resource table with CREATING and submit creation flight.
    *
    * <p>This should be done as a part of a transaction because we don't want resource state update
-   * without submitting a flight. The TransactionStatus is unused, but a part of the signature as a
-   * reminder.
+   * without submitting a flight. It will still work when Stairway submission succeeds but DB update
+   * failed. Because in Stairway flight, we check DB state and will abort the flight if DB entity in
+   * bad state.
```
wchamber, line 58: nit: some wordsmithing
`// If the Stairway submission fails, the transaction will be rolled back. If the Stairway submission succeeds and t the DB update transaction fails, the flight checks the DB state and aborts if the state is bad.`

src/main/java/bio/terra/buffer/service/resource/flight/AssertResourceCreatingInDbStep.java
wchamber: It's not so much that I'm worried that the code we have now is wrong for the failure case, it's that I'd like to have the test so that we don't accidentally break it by changing the code later. This is a subtle race bug, which is the best kind of thing to have a test on.
I'd imagine the test looks like

1. Create the resource in the DAO.
2. Start flight with scheduler. This flight starts with a latch step before any other steps.
3. Delete the resource in the DAO.
4. Resume the latch step.
5. Verify the flight is aborted and no project was created.

We should be in a spot where it's not hard to write these kinds of tests, but if that's not the case, fine.

src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -260,6 +288,21 @@ public ErrorCreateProjectFlight(FlightMap inputParameters, Object applicationCon
...
+      super(inputParameters, applicationContext);
+    }
+
+    @Override
+    protected void addStep(Step step, RetryRule retryRule) {
+      if (step instanceof AssertResourceCreatingStep) {
+        addStep(new LatchStep());
+      }
+      super.addStep(step);
+    }
```
wchamber, line 303: no work needed: I wish there was a cleaner way to do this, but I think we have to do it this way.

src/test/java/bio/terra/buffer/service/resource/flight/LatchStep.java
```diff
@@ -30,6 +30,7 @@ public static void releaseLatch() {
   @Override
   public StepResult doStep(FlightContext flightContext) throws InterruptedException {
     latch.await();
+    System.out.println("~~~~~~~LatchStep");
```
wchamber, line 33: nit: remove this?

src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -241,6 +241,34 @@ public void testCreateValidLabel() {
                 + "1234567890"));
   }
 
+  @Test
+  public void testCreateGoogleProject_errorWhenResourceStateChange() throws Exception {
```
wchamber, line 245: is there a related test for resource deletion? the resource deletion flight gets submitted, but the transaction to mark the resource as deleting fails. subsequently, the same resource gets handed out.

wchamber (approved):
sorry for the slow turnaround.

PR #103, opened 2021-01-26
[PF-368] Add workspace_manager_v3 pool to dev and perf

tnadler:


yonghaoy (approved):
What's your plan for using perf?

tnadler:
Decision was that for now we can point perf->perf as no WSM tests are running regular performance tests. In the long run we may deprecate perf overall.

PR #104, opened 2021-01-27
[no ticket]: Add rate limiting on project creation. 

yonghaoy:
Backgrond: We are seeing many 429 errors(mainly are serviceusage.batchEnabled).
The default quota limit is 0.2/s. So I add this rate limit based on the match I put in math. 

See [GCP console metrics](https://console.cloud.google.com/monitoring/metrics-explorer?authuser=0&organizationId=548622027621&project=terra-kernel-k8s&pageState=%7B%22xyChart%22:%7B%22dataSets%22:%5B%7B%22timeSeriesFilter%22:%7B%22filter%22:%22metric.type%3D%5C%22custom.googleapis.com%2Fopencensus%2Fterra%2Fbuffer%2Fresource%2Fresource_state_count%5C%22%20resource.type%3D%5C%22k8s_container%5C%22%20resource.label.%5C%22namespace_name%5C%22%3D%5C%22terra-yyu%5C%22%20metric.label.%5C%22pool_id%5C%22%3D%5C%22yyu_resource_toolsalpha_v2%5C%22%22,%22minAlignmentPeriod%22:%2260s%22,%22aggregations%22:%5B%7B%22perSeriesAligner%22:%22ALIGN_MEAN%22,%22crossSeriesReducer%22:%22REDUCE_SUM%22,%22groupByFields%22:%5B%22metric.label.%5C%22resource_state%5C%22%22%5D%7D,%7B%22crossSeriesReducer%22:%22REDUCE_NONE%22%7D%5D%7D,%22targetAxis%22:%22Y1%22,%22plotType%22:%22LINE%22%7D%5D,%22options%22:%7B%22mode%22:%22COLOR%22%7D,%22constantLines%22:%5B%5D,%22timeshiftDuration%22:%220s%22,%22y1Axis%22:%7B%22label%22:%22y1Axis%22,%22scale%22:%22LINEAR%22%7D%7D,%22isAutoRefresh%22:true,%22timeSelection%22:%7B%22timeRange%22:%226h%22%7D%7D)


wchamber (commented):
I'm a little nervous that these numbers operate per-pool but the quota is across all pools. If we add a new pool, or remove one, we have to remember to tweak all of these numbers. Additionally, different environments may have different numbers of pools.
I think it'd be better to define our limits for how often to query for flights to submit/how many to submit per query at a global level, then taking turns going to each pool for a batch of submissions.
Is that much more work?

(Another idea is to look at how many flights are CREATING to decide how many to submit instead of relying on how long each flight takes to complete, but I think that's a *bad* idea right now because we leak flights into CREATING when we're over quota, so we could get stuck.)

yonghaoy:
> I'm a little nervous that these numbers operate per-pool but the quota is across all pools. If we add a new pool, or remove one, we have to remember to tweak all of these numbers. Additionally, different environments may have different numbers of pools.
> I think it'd be better to define our limits for how often to query for flights to submit/how many to submit per query at a global level, then taking turns going to each pool for a batch of submissions.
> Is that much more work?
> 
> (Another idea is to look at how many flights are CREATING to decide how many to submit instead of relying on how long each flight takes to complete, but I think that's a _bad_ idea right now because we leak flights into CREATING when we're over quota, so we could get stuck. Though if we change our flight creation to put things in a STUCK state on failure/rollback, maybe that's better. Just brainstorming...)

Those numbers are estimated that assume we have 4 pools, but I think this numbers should work well with 1~10 pools. 
We don't need to tweek the numbers that often. 
Even currently, without the change, we do see many 429 errors(27k), but resources are created successfully with correct state. 

This number can reduce the 429 error numbers, but I think 0 is too aggressive. Dropped from 27k to 500 is also a good improvement. And even the pool is 10, I think RBS will still works fine with the existing retries.

wchamber:
sg. Add comments talking about the range of the acceptable number of pools please.

src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
```diff
@@ -6,6 +6,17 @@
...
 
+/**
+ * This configuration is to rate limit on project creation/deletion.
+ *
+ * <p>Rate limiting: The bottleneck for GCP quota comes from ServiceUsage.batchEnable API. By
+ * default, it is 20/100s. Here is how we estimate the time: A Creation flight takes 600~900
+ * seconds. If scheduler runs every 10 seconds, in 900 seconds it schedulers runs 70 times. If we
+ * have 4 pools, it schedules ({@code resourceCreationPerPoolLimit} * 4) = 4 flights. In total 70
+ * times run, 280 flights is scheduled.
+ * We estimate this number should works when pool number is 1~10. If we see more errors, we will comeback and rewise
```
wchamber, line 17: nit: s/`rewise`/`revise`. and I think run spotless Apply.

src/main/resources/application.yml
```diff
@@ -20,7 +20,7 @@ buffer:
       uri: jdbc:postgresql://127.0.0.1:5432/${BUFFER_STAIRWAY_DATABASE_NAME}
       username:
     force-clean-start: false
-    # This decides how soon we can fill a pool. Example: if total pools size is 3000 and each flight takes 10 minues.
+    # This decides how soon we can fill a pool. Example: if total pools size is 3000 and each flight takes 10 minuets.
```
wchamber, line 23: minutes

wchamber (approved):

PR #105, opened 2021-02-01
[PF-275] Switch to use bio.terraform.common.exception package

yonghaoy:
waiting on https://github.com/DataBiosphere/terra-common-lib/pull/12


src/main/java/bio/terra/buffer/service/pool/PoolService.java
```diff
@@ -66,8 +78,8 @@ public ResourceInfo handoutResource(PoolId poolId, RequestHandoutId requestHando
 
   /** Gets pool information by given {@link PoolId}. */
   public PoolInfo getPoolInfo(PoolId poolId) {
-    Optional<PoolAndResourceStates> poolAndResourceStates =
```
gjuggler, line 69: Curious, why this switch in syntax?

gjuggler (approved):

PR #107, opened 2021-02-03
[PF-247] Github action to run perf test nightly

yonghaoy:
1. Github action to trigger the nightly run at 1am est
2. Change pool size to 500, 1500 is too aggressive and it takes few hours. 
3. I just use `upload-artifact` to upload report, and seems that meets my need. If I missed anything reason why WSM uploads this to GCP bucket, please let me know.
4. The next step would be getting QA help to port test result to some slack channel.


gjuggler:
> 1. Github action to trigger the nightly run at 1am est
> 2. Change pool size to 500, 1500 is too aggressive and it takes few hours.

I'm a little unclear on what the threadPoolSize does — is this controlling how many parallel scripts we'll kick off?

And are you saying that it takes a few hours at 1500, or that it takes a few hours at 500?

Not sure I have any opinion on this point, I'm just trying to learn :)

> 3. I just use `upload-artifact` to upload report, and seems that meets my need. If I missed anything reason why WSM uploads this to GCP bucket, please let me know.

The ultimate goal is for all perf results to be exported to BQ for central dashboarding & analysis by QA. It's also easier to interpret perf test results if they're uploaded to a central place — kind of like inspecting tracing spans in the cloud console vs. having a zip file of traces sitting somewhere in GitHub.

So maybe this PR doesn't need to add the BQ upload, but I think it's an important goal to have a consistent flow for all services.

> 4. The next step would be getting QA help to port test result to some slack channel.

Yup, this is a good conversation to start having with QA team. My hope is that they'll be able to set up some automation, based on analysis of the BQ tables, to automatically determine when a given nightly run is worth flagging.

.github/workflows/test-runner-nightly-perf.yml
```diff
@@ -0,0 +1,51 @@
...
+            -e "VAULT_ADDR=${VAULT_ADDR}" \
+            vault:1.1.0 \
+            vault write -field token \
+              auth/approle/login role_id=${{ secrets.VAULT_APPROLE_ROLE_ID }} \
+              secret_id=${{ secrets.VAULT_APPROLE_SECRET_ID }})
+          echo ::set-output name=vault-token::$VAULT_TOKEN
+          echo ::add-mask::$VAULT_TOKEN
+      - name: Render configs for Test Runner
+        run: |
+          buffer-clienttests/tools/render-config.sh perf
```
gjuggler, line 33: For now this seems fine, as long as we don't expect much contention in the perf cluster. But should we look to eventually create a rbstest environment in the integration cluster and execute nightly tests there, to follow the pattern from workspace manager? We could at least file a ticket or two to track.

(Thinking further on this... is there any concern that the perf buffer service won't have projects cleaned up by Janitor? Or is there some explicit Janitor coverage in that environment?)

gjuggler (approved):

.github/workflows/test-runner-nightly-perf.yml
yonghaoy: 1. I think I should create a env called 'buffer-perf', and use that for buffer service test. 
2. And the existing buffer integration tests is using toolsalpha. I should create a buffertest for that. But that should just be a 'rename' thing, because toolsalpha was just created for buffer test. 
3: We can control whether to use Janitor by turning on this flag: https://github.com/broadinstitute/terra-helmfile/blob/master/terra/values/buffer/live/perf.yaml#L25

yonghaoy:
> I'm a little unclear on what the threadPoolSize does — is this controlling how many parallel scripts we'll kick off?
numberOfUserJourneyThreadsToRun: Is how many total threads we want to run. Each thread does this one time:
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/buffer-clienttests/src/main/java/scripts/testscripts/HandoutResource.java

So in total we expect 500 resources are handout, and pool should refill those 500. Now we deduce QPS because of quota issue. And we agree that 1500 is too aggressive, that takes ~5 hours to finish. 500 is a better reasonable chose. 
> 
> The ultimate goal is for all perf results to be exported to BQ for central dashboarding & analysis by QA. It's also easier to interpret perf test results if they're uploaded to a central place — kind of like inspecting tracing spans in the cloud console vs. having a zip file of traces sitting somewhere in GitHub.
> 
Thanks for explaining that, that is the new thing to me. Buffer service perf test is special, and the time it takes various, 500 pool can be 1hr ~ 1.5hr. And I don't think that is a time sensitive task. Should be fine with artifact?   
> 
> 
> Yup, this is a good conversation to start having with QA team. My hope is that they'll be able to set up some automation, based on analysis of the BQ tables, to automatically determine when a given nightly run is worth flagging.
I do remember Ambica mentioned that to me one time. I will check with her.

PR #109, opened 2021-02-05
[PF-242]: Add Stairway tracing hook

yonghaoy:


build.gradle
```diff
@@ -67,8 +67,8 @@ repositories {
 
 dependencies {
     // Terra deps
-    implementation group: 'bio.terra', name: 'stairway', version: '0.0.20'
-    implementation group: 'bio.terra', name: 'terra-common-lib', version: '0.0.11-SNAPSHOT'
+    implementation group: 'bio.terra', name: 'stairway', version: '0.0.36-SNAPSHOT'
+    implementation group: "bio.terra", name: 'terra-common-lib', version: "0.0.13-SNAPSHOT"
```
wchamber, line 71: let's use consistent quoting.

wchamber (approved):
Update the gradle lockfile as well?

src/main/java/bio/terra/buffer/service/resource/FlightManager.java
```diff
@@ -94,7 +99,10 @@ public FlightManager(
       stairway.submitToQueue(
           flightId, flightSubmission.clazz(), flightSubmission.inputParameters());
       return Optional.of(flightId);
-    } catch (DatabaseOperationException | StairwayExecutionException | InterruptedException e) {
+    } catch (DatabaseOperationException
```
wchamber, line 102: optional: Catching `StairwayException | InterruptedException`  would handle all the use cases and be future-proof to more exceptions being added I think.

PR #117, opened 2021-02-19
[PF-161]: Use the common library's LiquibaseMigrator

yonghaoy:


src/main/java/bio/terra/buffer/app/Main.java
```diff
@@ -6,7 +6,7 @@
 import org.springframework.context.annotation.ComponentScan;
 
 @SpringBootApplication(exclude = {DataSourceAutoConfiguration.class})
-@ComponentScan(basePackages = "bio.terra.buffer")
+@ComponentScan(basePackages = {"bio.terra.buffer", "bio.terra.common"})
```
ddietterich, line 9: This did not work for me. I think you need to scan the specific package. It would have to be "bio.terra.common.migrate".

ddietterich (commented):
I think you will need one change on the `@ComponentScan`. I commented there. I'd be interested if it turns out you _do not_ need that change.

Otherwise, it looks good to me.

src/main/java/bio/terra/buffer/app/Main.java
yonghaoy: it is working for me. At least test passs.

gjuggler: +1 to Dan's comment. Even if it works, I think we should align on a convention where each common-lib sub-package is scanned independently. It's a small # of lines of code to manage, and it makes it more explicit what exactly is being adopted and why.

See https://github.com/DataBiosphere/terra-workspace-manager/blob/d3cc1129b5a6f1c9a1eb5887276836d854c791ae/src/main/java/bio/terra/workspace/app/Main.java for my example in WSM.

yonghaoy: Done to switch to use sub-package. My understanding what's Dan's comment is `bio.terra.common` does not work for him in WSM, and he is interested to see if it works here. 
To Dan's this question: The parent package does work for Buffer but I don't know why...
Changed to sub-package to follow the common pattern. Good to know there is a pattern...

ddietterich (approved):

PR #118, opened 2021-02-19
[QA-1338] Notify qa channel for pert test result

yonghaoy:


.github/workflows/test-runner-nightly-perf.yml
```diff
@@ -1,51 +1,64 @@
...
+        cd buffer-clienttests
+        TEST_RUNNER_SERVER_SPECIFICATION_FILE="perf.json" ./gradlew  runTest --args="configs/EmptyPoolAndWaitForRefill.json build/reports"
+    - name: Upload Test Reports
+      if: always()
+      uses: actions/upload-artifact@v1
+      with:
+        name: Test Reports
+        path: buffer-clienttests/build/reports
+    - name: "Notify QA Slack"
+      if: ${{ github.event_name != 'workflow_dispatch' && always() }}
```
gjuggler, line 54: The "always()" call threw me off here, since it's not exactly obvious (to someone who hasn't just read the GitHub docs) what's happening. Can you add a comment above explaining why always() needs to be called? Or is this common enough in GHA land that all devs should just be expected to know what's going on?

Also, according to https://github.community/t/multiple-if-conditional-with-always-is-always-true/128840/3 you can leave out the `${{ }}` wrapping for conditionals.

gjuggler (approved):

.github/workflows/test-runner-nightly-perf.yml
yonghaoy: Done

PR #119, opened 2021-02-23
[PF-460] Add staging and alpha resource config

yonghaoy:


src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v1.yml
```diff
@@ -0,0 +1,30 @@
...
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-dev"
+    scheme: "RANDOM_CHAR"
+  # test.firecloud.org/alpha/CommunityWorkbench
+  parentFolderId: "102618195587"
+  billingAccount: "01A82E-CA8A14-367457"
+  enabledApis:
+  - "bigquery-json.googleapis.com"
+  - "compute.googleapis.com"
```
gjuggler, line 13: nit: maybe it's the same in other pre-existing configs, but L13 and L14 should be alphabetized, they should go before containerregistry.googleapis.com

gjuggler (approved):

src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v1.yml
yonghaoy: Done for alpha and staging. Leave other as it because re-order for existing pools requires a pool upgrade(delete and add)

PR #120, opened 2021-02-23
[CA-1084] project setup scripts

andy7i:
https://broadworkbench.atlassian.net/browse/CA-1084

These yamls were copied from the existing (v2) yamls and version bumped. The enabled APIs are the same as in DM. The groups (terra-billing@test.firecloud.org and firecloud-project-owners@test.firecloud.org) were added to follow what Rawls / DM currently does.

Arguably, instead of adding those groups to each test env project, we could add grant folder level permissions to both groups.

I made a spreadsheet that helped me track the permissions (it's a bit messy so I'd be happy to walk through it with folks): 
https://docs.google.com/spreadsheets/d/1n4BnfYIbEFSybKvh8WGbgbGNhYub8SdZexya8Pf5p4k/edit?pli=1#gid=1572590727

The PRs related to this set of changes are:
https://github.com/broadinstitute/rawls/pull/1369
https://github.com/broadinstitute/terraform-terra/pull/144
https://github.com/broadinstitute/firecloud-develop/pull/2445
https://github.com/DataBiosphere/terra-resource-buffer/pull/120

### DIFF
--- a/src/main/resources/config/dev/resource-config/aou_ws_resource_test_v3.yml
+++ b/src/main/resources/config/dev/resource-config/aou_ws_resource_test_v3.yml
@@ -1,0 +1,33 @@
+# All Of Us buffered workspace template
+---
+configName: "aou_ws_resource_test_v3"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "aou-rw-test"
+    scheme: "RANDOM_CHAR"
+  # test.firecloud.org/terra_dev_aou_test
+  parentFolderId: "737976594150"
+  billingAccount: "01A82E-CA8A14-367457"
+  enabledApis:
+    - "bigquery-json.googleapis.com"
+    - "compute.googleapis.com"
+    - "container.googleapis.com"
+    - "cloudbilling.googleapis.com"
+    - "clouderrorreporting.googleapis.com"
+    - "cloudkms.googleapis.com"
+    - "cloudtrace.googleapis.com"
+    - "containerregistry.googleapis.com"
+    - "dataflow.googleapis.com"
+    - "dataproc.googleapis.com"
+    - "dns.googleapis.com"
+    - "genomics.googleapis.com"
+    - "lifesciences.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
+  iamBindings:
+    - members: [ "group:terra-billing@test.firecloud.org", "group:firecloud-project-owners@test.firecloud.org" ]
+      role: "roles/owner"
+  network:
+    enableNetworkMonitoring: "true"
--- a/src/main/resources/config/perf/pool_schema.yml
+++ b/src/main/resources/config/perf/pool_schema.yml
@@ -16,3 +16,6 @@ poolConfigs:
   - poolId: "workspace_manager_v5"
     size: 20
     resourceConfigName: "workspace_manager_v5"
+  - poolId: "resource_perf_v3"
+    size: 1500
+    resourceConfigName: "resource_perf_v3"
--- a/src/main/resources/config/perf/resource-config/resource_perf_v3.yml
+++ b/src/main/resources/config/perf/resource-config/resource_perf_v3.yml
@@ -1,0 +1,37 @@
+# Resource template for running buffer service performance test
+---
+configName: "resource_perf_v3"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-perf"
+    scheme: "RANDOM_CHAR"
+  # RBS Testing
+  parentFolderId: "637867149294"
+  billingAccount: "01A82E-CA8A14-367457"
+  enabledApis:
+    - "bigquery-json.googleapis.com"
+    - "compute.googleapis.com"
+    - "container.googleapis.com"
+    - "cloudbilling.googleapis.com"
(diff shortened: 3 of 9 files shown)


src/main/resources/config/tools/resource-config/cwb_ws_resource_fiab_v3.yml
```diff
@@ -0,0 +1,34 @@
...
+    - "dns.googleapis.com"
+    - "genomics.googleapis.com"
+    - "lifesciences.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
+  iamBindings:
+    - members: [ "group:terra-billing@test.firecloud.org", "group:firecloud-project-owners@test.firecloud.org",
+                 "group:terra-billing@quality.firecloud.org", "group:firecloud-project-owners@quality.firecloud.org" ]
```
andy7i, line 31: since there are dev and qa fiabs, I added both quality.firecloud.org and test.firecloud.org

src/main/resources/config/dev/resource-config/aou_ws_resource_test_v3.yml
```diff
@@ -0,0 +1,33 @@
...
+    - "dataproc.googleapis.com"
+    - "dns.googleapis.com"
+    - "genomics.googleapis.com"
+    - "lifesciences.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
+  iamBindings:
+    - members: [ "group:terra-billing@test.firecloud.org", "group:firecloud-project-owners@test.firecloud.org" ]
```
andy7i, line 30: these yamls were copied from the existing (v2) yamls and version bumped. The enabled APIs are the same as in DM. The members here were added to follow the DM model.

src/main/resources/config/perf/resource-config/resource_perf_v3.yml
```diff
@@ -0,0 +1,37 @@
...
+    - "lifesciences.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
+  iamBindings:
+    - members: ["group:terra-rbs-test@broadinstitute.org"]
+      role: "roles/editor"
+    - members: ["group:terra-rbs-viewer-test@broadinstitute.org"]
+      role: "roles/viewer"
```
andy7i, line 33: these are the members that already existed in the perf and toolsalpha yamls so I left them in.

src/main/resources/config/perf/pool_schema.yml
```diff
@@ -16,3 +16,6 @@ poolConfigs:
   - poolId: "workspace_manager_v5"
     size: 20
     resourceConfigName: "workspace_manager_v5"
+  - poolId: "resource_perf_v3"
```
yonghaoy, line 19: The perf is RBS self test, it will not used by AoU or Rawls, so you don't need this.

src/main/resources/config/tools/resource-config/cwb_ws_resource_fiab_v3.yml
yonghaoy: is quality for perf test? That one we need another rawls-perf env for that. Perf traffic might be too much for tools RBS? I would want to know more details about that...
BTW, we don't have perf buffer setup yet.

yonghaoy (changes requested):
If you sync PR, you will see I just added staging and alpha...

yonghaoy:
And why not set it up in folder level? I that that is what we are doing today.

gpolumbo-broad (approved):
slight concern for the repetition needed in all of these.  Means when we add/remove one api we need to modify 4 different files.  But that's outside the scope of this PR.

src/main/resources/config/tools/resource-config/cwb_ws_resource_fiab_v3.yml
dvoet: @yonghaoy quality is also for fiab, not perf. Fiabs can run in 2 places: dev which uses test.firecloud.org and qa which uses quality.firecloud.org.

I don't think they should use the same pool though

dvoet:
why don't we add the groups to the folder?

yonghaoy:
> slight concern for the repetition needed in all of these. Means when we add/remove one api we need to modify 4 different files. But that's outside the scope of this PR.

We do realize that is a problem, having https://broadworkbench.atlassian.net/browse/PF-260 to track. But there is no plan to do it recently, but please let me know if you think it is important

src/main/resources/config/tools/resource-config/cwb_ws_resource_fiab_v3.yml
yonghaoy: I see, right now, there is no RBS&Janitor has permission on quality.firecloud.org, if we need such a pool, once we have that folder created, we also need to grant `tools` Janitor and RBS permission to that folder:
https://github.com/broadinstitute/terraform-ap-deployments/blob/master/terra-env/tfvars/tools.tfvars
(Add that folders in `buffer_external_folder_ids`, and `janitor_google_folder_ids`)

andy7i:
closing PR since the permissions are being set at the folder level now as part of https://github.com/broadinstitute/terraform-terra/pull/144

PR #122, opened 2021-03-03
[PF-231] Use storage IAM instead of legacy ACL for storage log bucket

yonghaoy:
I followed https://cloud.google.com/storage/docs/access-logs#json-api so choose to use legacy legacyBucketWriter role


yonghaoy:
@dionloy FYI, this should be the last step

gjuggler (approved):

PR #123, opened 2021-03-04
PF-496: Add flags to control deletion of default Compute Engine SA and VPC network.

marikomedlock:
- Added 2 flags to the resource config schema: `Network.keepDefaultNetwork` and `ComputeEngine.keepDefaultServiceAcct`. Default for both is false, which matches the current behavior.
  - Added utility methods in `GoogleUtils` to check whether these flags are set. Similar to the existing `isNetworkMonitoringEnabled`, these methods include null checks, to handle resource-config definitions that do not define these flags.
  - If the `keepDefaultServiceAcct` flag is true, then `DeleteDefaultServiceAccountStep` is a no-op.
  - If the `keepDefaultNetwork` flag is true, then `DeleteDefaultNetworkStep` is a no-op, and `CreateFirewallRuleStep` creates identical firewall rules for both the "high-security" network and the default network.
  - Added two tests, one for each flag.

- Created two new resource-config files for the `tools` and `dev` environments, and updated their pool_schema files accordingly.
  - These flags should be enabled for WSM dev and optionally for personal environment WSM.
  - There is a related [terra-helmfile PR](https://github.com/broadinstitute/terra-helmfile/pull/1064) to update WSM dev, test and mmedlock environments to point to this new RBS pool.
--------------------

The purpose of this change is to support running Nextflow against RBS-created projects. Nextflow requires the default Compute Engine SA and the default VPC network to both exist. Previously, RBS always deletes these things. This PR adds flags to control this behavior, so we can turn it off in select environments.

In the long-term, we want to keep the current behavior (i.e. delete default SA and network). But that will likely require some contributions to the Nextflow codebase, and we want to use Nextflow for the enterprise pilot in the meantime. So these flags are a near-term workaround.

### DIFF
--- a/src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
@@ -85,7 +85,8 @@ public class CreateFirewallRuleStep implements Step {
       addFirewallRule(LEONARDO_SSL.setNetwork(highSecurityNetwork.getSelfLink()));
 
       // If the default network was not deleted, then create identical firewall rules for it.
-      if (!gcpProjectConfig.getNetwork().isDeleteDefaultNetwork()) {
+      Boolean deleteDefaultNetwork = gcpProjectConfig.getNetwork().isDeleteDefaultNetwork();
+      if (deleteDefaultNetwork != null && !deleteDefaultNetwork) {
         Network defaultNetwork =
             getResource(
                     () -> computeCow.networks().get(projectId, DEFAULT_NETWORK_NAME).execute(), 404)
--- a/src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
@@ -24,8 +24,20 @@ import org.slf4j.LoggerFactory;
 
 /** Creates firewall rules for the GCP project. */
 public class CreateFirewallRuleStep implements Step {
-  private static final String ALLOW_INTERNAL_RULE_NAME = "allow-internal";
-  private static final String LEONARDO_SSL_RULE_NAME = "leonardo-ssl";
+  /** Names for firewall rules on the high-security network (called 'network'). */
+  @VisibleForTesting
+  public static final String ALLOW_INTERNAL_RULE_NAME_FOR_NETWORK = "allow-internal";
+
+  @VisibleForTesting public static final String LEONARDO_SSL_RULE_NAME_FOR_NETWORK = "leonardo-ssl";
+
+  /** Names for firewall rules on the default network (called 'default'). */
+  @VisibleForTesting
+  public static final String ALLOW_INTERNAL_RULE_NAME_FOR_DEFAULT =
+      DEFAULT_NETWORK_NAME + "-vpc-" + ALLOW_INTERNAL_RULE_NAME_FOR_NETWORK;
+
+  @VisibleForTesting
+  public static final String LEONARDO_SSL_RULE_NAME_FOR_DEFAULT =
+      DEFAULT_NETWORK_NAME + "-vpc-" + LEONARDO_SSL_RULE_NAME_FOR_NETWORK;
 
   /**
    * See <a
@@ -34,7 +46,6 @@ public class CreateFirewallRuleStep implements Step {
   @VisibleForTesting
   public static final Firewall ALLOW_INTERNAL =
       new Firewall()
-          .setName(ALLOW_INTERNAL_RULE_NAME)
           .setDescription("Allow internal traffic on the network.")
           .setDirection("INGRESS")
           .setSourceRanges(ImmutableList.of("10.128.0.0/9"))
@@ -51,7 +62,6 @@ public class CreateFirewallRuleStep implements Step {
   @VisibleForTesting
   public static final Firewall LEONARDO_SSL =
       new Firewall()
-          .setName(LEONARDO_SSL_RULE_NAME)
           .setDescription("Allow SSL traffic from Leonardo-managed VMs.")
           .setDirection("INGRESS")
           .setSourceRanges(ImmutableList.of("0.0.0.0/0"))
@@ -81,11 +91,17 @@ public class CreateFirewallRuleStep implements Step {
       Network highSecurityNetwork =
           getResource(() -> computeCow.networks().get(projectId, NETWORK_NAME).execute(), 404)
               .get();
-      addFirewallRule(ALLOW_INTERNAL.setNetwork(highSecurityNetwork.getSelfLink()));
-      addFirewallRule(LEONARDO_SSL.setNetwork(highSecurityNetwork.getSelfLink()));
+      addFirewallRule(
+          ALLOW_INTERNAL
+              .setNetwork(highSecurityNetwork.getSelfLink())
+              .setName(ALLOW_INTERNAL_RULE_NAME_FOR_NETWORK));
+      addFirewallRule(
+          LEONARDO_SSL
+              .setNetwork(highSecurityNetwork.getSelfLink())
+              .setName(LEONARDO_SSL_RULE_NAME_FOR_NETWORK));
 
       // If the default network was not deleted, then create identical firewall rules for it.
-      if (!isDeleteDefaultNetwork(gcpProjectConfig)) {
+      if (keepDefaultNetwork(gcpProjectConfig)) {
         Network defaultNetwork =
             getResource(
                     () -> computeCow.networks().get(projectId, DEFAULT_NETWORK_NAME).execute(), 404)
@@ -95,11 +111,11 @@ public class CreateFirewallRuleStep implements Step {
         addFirewallRule(
             ALLOW_INTERNAL
                 .setNetwork(defaultNetwork.getSelfLink())
-                .setName(DEFAULT_NETWORK_NAME + "-" + ALLOW_INTERNAL_RULE_NAME));
+                .setName(ALLOW_INTERNAL_RULE_NAME_FOR_DEFAULT));
         addFirewallRule(
             LEONARDO_SSL
                 .setNetwork(defaultNetwork.getSelfLink())
-                .setName(DEFAULT_NETWORK_NAME + "-" + LEONARDO_SSL_RULE_NAME));
+                .setName(LEONARDO_SSL_RULE_NAME_FOR_DEFAULT));
       }
 
       for (OperationCow<?> operation : operationsToPoll) {
--- a/src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
@@ -61,23 +61,34 @@ public class CreateFirewallRuleStep implements Step {
   private final Logger logger = LoggerFactory.getLogger(CreateFirewallRuleStep.class);
   private final CloudComputeCow computeCow;
 
-  /** Optional prefix for the firewall rule name. */
-  private final String ruleNamePrefix;
+  /** Name of the network to add the firewall rule for. */
+  private final String networkName;
+
+  /** Name of the "allow-internal" firewall rule. */
+  private final String allowInternalRuleName;
+
+  /** Name of the "leonardo-ssl" firewall rule. */
+  private final String leonardoSslRuleName;
 
   public CreateFirewallRuleStep(CloudComputeCow computeCow) {
-    this(computeCow, ""); // no prefix is the default
+    this.computeCow = computeCow;
+    this.networkName = NETWORK_NAME;
+    this.allowInternalRuleName = ALLOW_INTERNAL_RULE_NAME;
+    this.leonardoSslRuleName = LEONARDO_SSL_RULE_NAME;
   }
 
   /**
-   * Constructor that allows specifying a prefix for the firewall rule names. This is useful for
-   * creating identical firewall rules for different networks (e.g. prefix = network name).
+   * Constructor that allows specifying the network name. Firewall rule names will be prefixed with
+   * the network name. This is useful for creating identical firewall rules for different networks.
    *
    * @param computeCow cloud compute wrapper object
-   * @param ruleNamePrefix prefix for firewall rule names
+   * @param networkName name of the network
    */
-  public CreateFirewallRuleStep(CloudComputeCow computeCow, String ruleNamePrefix) {
+  public CreateFirewallRuleStep(CloudComputeCow computeCow, String networkName) {
     this.computeCow = computeCow;
-    this.ruleNamePrefix = ruleNamePrefix;
+    this.networkName = networkName;
+    this.allowInternalRuleName = networkName + "-" + ALLOW_INTERNAL_RULE_NAME;
+    this.leonardoSslRuleName = networkName + "-" + LEONARDO_SSL_RULE_NAME;
   }
 
   @Override
@@ -87,8 +98,7 @@ public class CreateFirewallRuleStep implements Step {
       // Network is already created and checked in previous step so here won't be empty.
       // If we got NPE, that means something went wrong with GCP, fine to just throw NPE here.
       Network network =
-          getResource(() -> computeCow.networks().get(projectId, NETWORK_NAME).execute(), 404)
-              .get();
+          getResource(() -> computeCow.networks().get(projectId, networkName).execute(), 404).get();
 
       List<OperationCow<?>> operationsToPoll = new ArrayList<>();
       createResourceAndIgnoreConflict(
@@ -99,7 +109,7 @@ public class CreateFirewallRuleStep implements Step {
                           projectId,
                           ALLOW_INTERNAL
                               .setNetwork(network.getSelfLink())
-                              .setName(ruleNamePrefix + ALLOW_INTERNAL_RULE_NAME))
+                              .setName(allowInternalRuleName))
                       .execute())
           .ifPresent(
               insertOperation ->
@@ -113,7 +123,7 @@ public class CreateFirewallRuleStep implements Step {
                           projectId,
                           LEONARDO_SSL
                               .setNetwork(network.getSelfLink())
-                              .setName(ruleNamePrefix + LEONARDO_SSL_RULE_NAME))
+                              .setName(leonardoSslRuleName))
                       .execute())
           .ifPresent(
               insertOperation ->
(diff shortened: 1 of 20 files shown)


src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
```diff
@@ -49,10 +49,29 @@ public GoogleProjectCreationFlight(FlightMap inputParameters, Object application
     addStep(new SetIamPolicyStep(rmCow, gcpProjectConfig), CLOUD_API_DEFAULT_RETRY);
     addStep(
         new CreateStorageLogBucketStep(clientConfig, gcpProjectConfig), CLOUD_API_DEFAULT_RETRY);
-    addStep(new DeleteDefaultServiceAccountStep(iamCow), CLOUD_API_DEFAULT_RETRY);
+
+    // project config determines whether we want to keep the Compute Engine default service account
+    if (gcpProjectConfig.getComputeEngine() != null
```
marikomedlock, line 54: I'm not sure this null check is needed. I think this is needed if I don't update all the .yml files under resources/config directory. But should I instead be updating all .yml files, so this is never null?

src/main/resources/config/toolsalpha/resource-config/resource_toolsalpha_v2.yml
```diff
@@ -33,3 +33,6 @@ gcpProjectConfig:
       role: "roles/viewer"
   network:
     enableNetworkMonitoring: "true"
+    keepDefaultNetwork: "false"
+  computeEngine:
+    deleteDefaultServiceAccount: "false"
```
marikomedlock, line 38: Should I be editing this file or creating a new _v3.yml instead?

yonghaoy: resource configs are immutable, we do need v3.
But why we changing this? Which RBS will use deleteDefaultServiceAccount: "false" pool?

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
```diff
@@ -49,10 +49,29 @@ public GoogleProjectCreationFlight(FlightMap inputParameters, Object application
...
+        && gcpProjectConfig.getComputeEngine().isDeleteDefaultServiceAccount()) {
```
yonghaoy, line 55: we usually always add those steps but add config check inside those steps. 
Like
```
 if(gcpProjectConfig.getComputeEngine().isDeleteDefaultServiceAccount()) {
 return Step.successStep()
}
```

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
yonghaoy: yeah, it does look wired, and if we keep adding things like that, it is also hard for us to track the exactly resource config, and know when to delete SA account, and when to not. 
But from your logic, seems all existings configs will stop deleting default SAs because their value is null.

I have been thinking about this recently, it is hard because the resource config is immutable, but as we keep adding new field in resource config schema, how to tell existing values which is the 'default value'? 
One solution, may not be the best, is to comment on the resource schema, saying what is the default value if not present. 
And when RBS populated those configs, we set some default values into DB. Logic is here:
https://github.com/DataBiosphere/terra-resource-buffer/blob/c319d27654cbd5910b978017c3e207dc57f221a8/src/main/java/bio/terra/buffer/service/pool/PoolService.java#L170
You can leave a null check, I can try to find a solution to fix that.

yonghaoy:
Actually, an easier solution, not to use ComputeEngine object, just add 'deleteDefaultComputeServiceAccount" in pool config directly, with a comment, default behaviour is false. 

And flight can check 
```
if deleteDefaultComputeServiceAccount != null && deleteDefaultComputeServiceAccount == true {
}
```

src/main/resources/config/toolsalpha/resource-config/resource_toolsalpha_v2.yml
marikomedlock: - I want WSM dev to have both delete flags set to "false". Would that be dev/workspace_manager_v6.yml? And then update terra-helmfile [here](https://github.com/broadinstitute/terra-helmfile/blob/902d123ec9e62403f1e1147543da616bc0f9ca3a/terra/values/workspacemanager/live/dev.yaml#L23)?

- I think it would be good to also have the option to set both delete flags to "false" in a personal environment. Would that be tools/workspace_manager_v5.yml? And then update terra-helmfile [here](https://github.com/broadinstitute/terra-helmfile/blob/902d123ec9e62403f1e1147543da616bc0f9ca3a/terra/values/workspacemanager/personal/mmedlock.yaml#L11)?

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
marikomedlock: Okay, I'll change ComputeEngine.deleteDefaultServiceAccount object.property to just a property. Seems like we still need a solution to resource schema upgrade but if this is the only use case now, agreed it doesn't seem like we need to figure this out now.

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
marikomedlock: I can do that for the default SA step, but what about the firewall rules step? Should I expand that one to create rules for both networks?

src/main/resources/config/toolsalpha/resource-config/resource_toolsalpha_v2.yml
yonghaoy: What you said about personal environment and WSM is correct. 
For the toolsalpha, you don't need one that unless you want your personal RBS having that pool.

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
```diff
@@ -49,10 +49,29 @@ public GoogleProjectCreationFlight(FlightMap inputParameters, Object application
...
+      // name (i.e. "default-").
+      addStep(
+          new CreateFirewallRuleStep(
+              cloudComputeCow, DeleteDefaultNetworkStep.DEFAULT_NETWORK_NAME),
+          CLOUD_API_DEFAULT_RETRY);
+    }
+
     addStep(new CreateNetworkStep(cloudComputeCow, gcpProjectConfig), CLOUD_API_DEFAULT_RETRY);
     addStep(new CreateRouteStep(cloudComputeCow, gcpProjectConfig), CLOUD_API_DEFAULT_RETRY);
     addStep(new CreateFirewallRuleStep(cloudComputeCow), CLOUD_API_DEFAULT_RETRY);
```
yonghaoy, line 77: There is a CreateFirewallRuleStep

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
```diff
@@ -49,10 +49,29 @@ public GoogleProjectCreationFlight(FlightMap inputParameters, Object application
...
+    if (gcpProjectConfig.getNetwork().isDeleteDefaultNetwork()) {
+      addStep(
+          new DeleteDefaultNetworkStep(cloudComputeCow, gcpProjectConfig), CLOUD_API_DEFAULT_RETRY);
+    } else {
+      // if we don't delete the default network, then create firewall rules for it that are
+      // identical to the ones for the second network below. prefix these rules with the network
+      // name (i.e. "default-").
+      addStep(
+          new CreateFirewallRuleStep(
+              cloudComputeCow, DeleteDefaultNetworkStep.DEFAULT_NETWORK_NAME),
```
yonghaoy, line 71: Might be good to move this file if it becomes a common variable:
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/service/resource/flight/GoogleUtils.java

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -63,8 +61,34 @@
...
+  /** Name of the "allow-internal" firewall rule. */
+  private final String allowInternalRuleName;
+
+  /** Name of the "leonardo-ssl" firewall rule. */
+  private final String leonardoSslRuleName;
+
   public CreateFirewallRuleStep(CloudComputeCow computeCow) {
     this.computeCow = computeCow;
+    this.networkName = NETWORK_NAME;
+    this.allowInternalRuleName = ALLOW_INTERNAL_RULE_NAME;
```
yonghaoy, line 76: This variable is not accurate now. Maybe ALLOW_INTERNAL_RULE_NAME_SUFFIX? or something else you think is better. Then add some comments to explain those two variables.

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
yonghaoy: I feel that is a better way. Take list of networks and create for each of them? Right now, two CreateFirewallRuleStep
 confuses me.

yonghaoy (changes requested):

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
marikomedlock: Moved config logic out of steps and into flight. PTAL.

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
marikomedlock: Modified that step to check the config, instead of calling it twice.

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
marikomedlock: done.

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
marikomedlock: Updated to have separate variables for each network's firewall rule names.

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -23,8 +24,23 @@
...
+
+  @VisibleForTesting public static final String LEONARDO_SSL_RULE_NAME_FOR_NETWORK = "leonardo-ssl";
+
+  /**
+   * Names for firewall rules on the default network (called 'default'). Rule names must be unique
+   * within a project, so prefix these rule names with 'default-vpc'.
+   */
+  @VisibleForTesting
+  public static final String ALLOW_INTERNAL_RULE_NAME_FOR_DEFAULT =
+      DEFAULT_NETWORK_NAME + "-vpc-" + ALLOW_INTERNAL_RULE_NAME_FOR_NETWORK;
```
yonghaoy, line 39: do we need 'vpc' in the middle? Not sure what is the usual practice for naming a firewall rule, but that does not seem to be a VPC.

yonghaoy (approved):

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -62,42 +76,48 @@
 
   private final Logger logger = LoggerFactory.getLogger(CreateFirewallRuleStep.class);
   private final CloudComputeCow computeCow;
+  private final GcpProjectConfig gcpProjectConfig;
+  private String projectId;
+  private List<OperationCow<?>> operationsToPoll = new ArrayList<>();
```
wchamber, line 81: Avoid using mutable member variables if possible. Using mutable member variables data flow makes it harder to see what's supposed to happen in what order.

Instead of making `operationsToPoll` a member variable, keep it as a local variable in doStep and return an Optional<OperationCow<?>> from addFirewallRule. That makes the data flow and ordering explicit, instead of the implicit ordering that addFirewallRule should be called before the operations are polled.

Making `projectId` a mutable member variable saves us from passing it to `addFirewallRule`, but I think it's simple enough to add an argument there and not require explaining/relying on projectId is set at the beginning of `doStep`.

wchamber (changes requested):
lgtm other than couple of things on the CreateFirewallRuleStep

src/main/java/bio/terra/buffer/service/resource/flight/DeleteDefaultNetworkStep.java
```diff
@@ -33,6 +32,11 @@ public DeleteDefaultNetworkStep(CloudComputeCow computeCow, GcpProjectConfig gcp
 
   @Override
   public StepResult doStep(FlightContext flightContext) throws RetryException {
+    if (keepDefaultNetwork(gcpProjectConfig)) {
+      logger.info("Skipping deletion of default network");
+      return StepResult.getStepResultSuccess();
```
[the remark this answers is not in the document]
marikomedlock: Yes, @yonghaoy requested this above. I originally had logic in the flight, not the steps. I don't have strong feelings though, prefer to be consistent with existing patterns in the codebase.

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -62,42 +76,48 @@
...
-                  computeCow
-                      .firewalls()
-                      .insert(projectId, LEONARDO_SSL.setNetwork(network.getSelfLink()))
-                      .execute())
-          .ifPresent(
-              insertOperation ->
-                  operationsToPoll.add(
-                      computeCow.globalOperations().operationCow(projectId, insertOperation)));
+      // If the default network was not deleted, then create identical firewall rules for it.
+      if (keepDefaultNetwork(gcpProjectConfig)) {
```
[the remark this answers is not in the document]
marikomedlock: Tickets already filed (PF-537, PF-538), will add TODOs in the code.

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -62,42 +76,48 @@
...
     try {
       // Network is already created and checked in previous step so here won't be empty.
       // If we got NPE, that means something went wrong with GCP, fine to just throw NPE here.
-      Network network =
+      Network highSecurityNetwork =
           getResource(() -> computeCow.networks().get(projectId, NETWORK_NAME).execute(), 404)
               .get();
+      addFirewallRule(
+          ALLOW_INTERNAL
+              .setNetwork(highSecurityNetwork.getSelfLink())
```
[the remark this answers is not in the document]
marikomedlock: Good catch. Will update.

src/main/java/bio/terra/buffer/service/resource/flight/GoogleUtils.java
```diff
@@ -69,6 +72,24 @@ public static boolean isNetworkMonitoringEnabled(GcpProjectConfig gcpProjectConf
         && gcpProjectConfig.getNetwork().isEnableNetworkMonitoring();
   }
 
+  /** Checks the config to see if we should keep the default network. */
+  public static boolean keepDefaultNetwork(GcpProjectConfig gcpProjectConfig) {
```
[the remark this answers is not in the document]
marikomedlock: will do. this was really confusing to me originally, and it seems easy to break for backwards compatibility. I think a wrapper around the proejct config would be a good addition.

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
wchamber: I think not having mutable member variables is of value in itself. Returning the optional is just one way to make that work. I
One thing to consider is that Steps are instantiated on Flight construction. What happens when a Step is retried? Does Stairway call `doStep` on the same Step instance multiple times? What does our code start to do when it polls operations from previous tries? You could get around this by reinstantiating `operationsToPoll` at the start of each doStep, but it's easier to reason about if it's just a local variable instead of a member variable.

It's not very verbose with `Optional#map` and `Optional#ifPresent`

```
private Optional<OperationCow<?>> addFirewallRule(Firewall rule) throws IOException {
    return createResourceAndIgnoreConflict(() -> computeCow.firewalls().insert(projectId, rule).execute())
            .map(insertOperation -> computeCow.globalOperations().operationCow(projectId, insertOperation));
  }

// in doStep
addFirewallRule(allowInternal).ifPresent(operationsToPool::add);
```

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -87,36 +55,37 @@ public CreateFirewallRuleStep(CloudComputeCow computeCow, GcpProjectConfig gcpPr
...
-              .setName(LEONARDO_SSL_RULE_NAME_FOR_NETWORK));
+      Firewall allowInternalRuleForNetwork =
+          buildAllowInternalFirewallRule(highSecurityNetwork, ALLOW_INTERNAL_RULE_NAME_FOR_NETWORK);
+      Firewall leonardoSslRuleForNetwork =
+          buildLeonardoSslFirewallRule(highSecurityNetwork, LEONARDO_SSL_RULE_NAME_FOR_NETWORK);
+      addFirewallRule(projectId, allowInternalRuleForNetwork).ifPresent(operationsToPoll::add);
+      addFirewallRule(projectId, leonardoSslRuleForNetwork).ifPresent(operationsToPoll::add);
 
+      // TODO: revisit whether we still need this flag after NF allows specifying a network
+      // https://broadworkbench.atlassian.net/browse/PF-538
```
wchamber, line 76: nit: We've been using a `// TODO(PF-538): ` todo style in comments in janitor, buffer, and workspace manager. worth being consistent?
ditto others

wchamber (approved):
thanks!

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
marikomedlock: ok

PR #124, opened 2021-03-08
[no ticket] Disable stackdriver when running RBS locally

yonghaoy:


wchamber (commented):
I was able to start buffer up at head, so I'm not sure what this is going to fix. If the SA is getting a 403, is it going to be able to make projects?
fwiw, I'm planning on making stackdriver export failures not prevent the server from coming up in the common package. If we wanted, we could change that here if there's something to unblock.

yonghaoy:
> I was able to start buffer up at head, so I'm not sure what this is going to fix. If the SA is getting a 403, is it going to be able to make projects?
> fwiw, I'm planning on making stackdriver export failures not prevent the server from coming up in the common package. If we wanted, we could change that here if there's something to unblock.

Buffer is able to start up, that one does not block buffer to start. 
Then do we still need this? Feels it is right to disable exporting metrics/tracing to stackdriver by default. Or say, it is necessary to export those things by default.

marikomedlock (commented):
I think RBS did start successfully locally, though I didn't do extensive testing. It just produced a lot of errors in the output.

I'm not blocked either way. I got confused because I thought I needed to stand up a server in order to run the `integrationTest` target. But that is my misunderstanding, and I was able to run those tests without starting RBS locally.

So from my side, it's okay to wait on changes to common lib. WDYT @yonghaoy ?

PR #125, opened 2021-03-11
PF-280 Migrate buffer to use terra common tracing package

wchamber:
Also make it a log error instaed of an exception if unable to export metrics to stackdriver.


jmczerk (approved):
LGTM

yonghaoy (approved):

yonghaoy (commented):
Actually I am not sure how that package will be used. Someone need to call
`StackdriverTraceExporter.createAndRegister`
And where it happens? 
And currently we have a flag to control both tracing and metrics. What happens after? Two flags? No flags for tracing?  If possible, I'd like this still be off in test:
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/test/resources/application-test.yml#L17

wchamber:
> Actually I am not sure how that package will be used. Someone need to call
> `StackdriverTraceExporter.createAndRegister`
> but where and who?
> And currently we have a flag to control both tracing and metrics. What happens after? Two flags? No flags for tracing? If possible, I'd like this be off in test:
> https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/test/resources/application-test.yml#L17

By adding `bio.terra.common.tracing` to the component scan, the [TracingConfig registers the tracing exporter](https://github.com/DataBiosphere/terra-common-lib/blob/887f36a798d8636d8b5b4f3482680d24cdd0f1cb/src/main/java/bio/terra/common/tracing/TracingConfig.java#L85).
Now we have 2 flags. We could keep having 2 flags when we moving stats to the common lib, or we could make a single flag to control all stackdriver exporter.
Added config so it's off in tests.

yonghaoy (approved):

PR #127, opened 2021-03-15
[PF-450] Switch to using the common logging package

gjuggler:
Enables the common logging package for resource buffer and turns on debug-level logs for CRL.

DO NOT MERGE until the related common-lib and CRL PRs are merged too.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,6 +48,8 @@ allprojects {
 }
 
 repositories {
+//  Uncomment to read from the local Maven cache.
+	mavenLocal()
     mavenCentral()
     maven {
         url 'https://broadinstitute.jfrog.io/broadinstitute/libs-release/'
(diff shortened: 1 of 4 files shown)


src/main/resources/application.yml
```diff
@@ -27,6 +27,9 @@ buffer:
     quiet-down-timeout: 20s
     terminate-timeout: 5s
 
+# Turn on debug-level logs for CRL, to include all CRL operation payloads.
+logging.level.bio.terra.cloudres: DEBUG
```
jaycarlton, line 31: confirm you want to leave DEBUG on

jaycarlton (approved):
Looks good

src/main/resources/application.yml
gjuggler: Thanks – yes, I think this is correct. I'd almost made the CRL per-operation logs output at `info` level, but I realized that this is legitimately debug-level logging, that we happen to want to turn on in the RBS context. Other services may make a different call, and it's useful to have these logs show up as debug-level in the Cloud Logging UI too.

@yonghaoy I think this was aligned with your instinct too, but let me know if you disagree.

We can always adjust this over time if the arrangement isn't quite right.

yonghaoy (approved):
Thanks!

build.gradle
```diff
@@ -48,6 +48,8 @@ allprojects {
 }
 
 repositories {
+//  Uncomment to read from the local Maven cache.
+//	mavenLocal()
```
yonghaoy, line 52: remove before merge?

gjuggler: Good point. Done.

PR #128, opened 2021-03-15
[PF-456] Use common lib for stairway init

jaycarlton:
goes with common lib PR https://github.com/DataBiosphere/terra-common-lib/pull/31


src/main/java/bio/terra/buffer/app/StartupInitializer.java
```diff
@@ -31,7 +30,7 @@ public static void initialize(ApplicationContext applicationContext) {
     } else if (bufferJdbcConfiguration.isUpdateDbOnStart()) {
       migrateService.upgrade(changelogPath, bufferJdbcConfiguration.getDataSource());
     }
-    applicationContext.getBean(StairwayComponent.class).initialize();
+    //    applicationContext.getBean(StairwayComponent.class).initialize();
```
yonghaoy, line 33: We should keep it here, but  common-lib's StairwayComponent instead.

src/main/java/bio/terra/buffer/app/configuration/BufferJdbcThing.java
```diff
@@ -12,7 +14,7 @@
 @EnableConfigurationProperties
 @EnableTransactionManagement
 @ConfigurationProperties(prefix = "buffer.db")
-public class BufferJdbcConfiguration extends JdbcConfiguration {
+public class BufferJdbcThing extends JdbcProperties {
```
jaycarlton, line 17: This is the bit that makes the least sense. It's not a `@Configuration` so I temporarily renamed it to thing.

build.gradle
```diff
@@ -48,6 +51,9 @@ allprojects {
 }
 
 repositories {
+    if (useMavenLocal) {
```
ddietterich, line 54: What is this for? And should it be part of this PR?
If it is a good thing, should we do it everywhere?

src/main/java/bio/terra/buffer/app/configuration/StairwayDatabaseConfiguration.java
```diff
@@ -5,8 +5,8 @@
 
 @Configuration
 @EnableConfigurationProperties(value = StairwayDatabaseProperties.class)
-public class StairwayDatabaseDatabaseConfiguration extends BaseDatabaseConfiguration {
-  public StairwayDatabaseDatabaseConfiguration(StairwayDatabaseProperties databaseProperties) {
+public class StairwayDatabaseConfiguration extends BaseDatabaseConfiguration {
```
ddietterich, line 8: Huh, I would have expected this to be in the common lib.

ddietterich (commented):
I'm curious about the `StairwayDatabaseConfiguration` class.
Otherwise, this looks good to me.

build.gradle
jaycarlton: Yes, it allows you to use locally deployed artifacts (like the common lib). I think it's a good thing. @gjuggler what do you think of this little paradigm?

I used the local repo extensively in testing this PR, so I didn't want to have to remove the maven local stuff.

src/main/java/bio/terra/buffer/app/configuration/StairwayDatabaseConfiguration.java
jaycarlton: Interesting. By what criteria? Sometimes it seemed arbitrary, and I went with the paths of least resistance.

build.gradle
yonghaoy: Or we can follow this instruction: https://github.com/DataBiosphere/terra-common-lib#local-testing
i.e, don't  add `mavenLocal` here. If engineer wants to test their change locally, then manually add temporary.

src/main/java/bio/terra/buffer/app/StartupInitializer.java
```diff
@@ -1,13 +1,16 @@
 package bio.terra.buffer.app;
 
-import bio.terra.buffer.app.configuration.BufferDatabaseDatabaseConfiguration;
+import bio.terra.buffer.app.configuration.BufferDatabaseConfiguration;
```
yonghaoy, line 3: thanks for doing this!

yonghaoy (commented):
We can not merge this PR before helm chart is updated with correct config: 
https://github.com/broadinstitute/terra-helm/blob/master/charts/buffer/templates/deployment.yaml#L137
The correct step should be: 
(1) Update helm chart to add the new configs,  e.g.  adding TERRA_COMMON_KUBERNETES_POD_NAME, 
TERRA_COMMON_BUFFER_KUBERNETES_POD_NAME_FILTER

(2) Wait until that PR merged, and deploied, merge this PR
(3) Delete those old configs from helm chart

src/main/resources/application.yml
```diff
@@ -14,18 +14,20 @@ buffer:
     # Whether to update pool when config changes during server startup.
     # Might be useful to avoid potential backward compatibility caused by config change and make tests easier.
     update-pool-on-start: true
-  stairway:
-    db:
-      password:
-      uri: jdbc:postgresql://127.0.0.1:5432/${BUFFER_STAIRWAY_DATABASE_NAME}
-      username:
```
yonghaoy, line 21: are you changing stairway DB configs?  I through we should still use this:
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/app/configuration/StairwayDatabaseProperties.java
That means the config should still be `buffer.stairway.db`

yonghaoy (changes requested):

src/main/java/bio/terra/buffer/app/configuration/StairwayDatabaseConfiguration.java
ddietterich: I would have thought Stairway database config would be processed in the common lib code, and not need to be present in the service code. I'm just trying to understand - not asking for any change.

src/main/resources/application.yml
ddietterich: Seems like a trade-off between common setup (like in helm) for a stairway db, and getting config names that are specific to a service.

ddietterich (approved):
This looks fine to me. I don't understand all of the choices, but I don't see anything wrong.

build.gradle
jaycarlton: This was intended as a prettier way to do that, so you don't have to dig in and find the spot to put `mavenLocal()`. You still have to bump version numbers though.

src/main/java/bio/terra/buffer/app/configuration/StairwayDatabaseConfiguration.java
jaycarlton: I never really landed on what felt like a good, intuitive boundary here. Might need to revisit at some point.

src/main/resources/application.yml
jaycarlton: @yonghaoy are you saying this file needs a change? Have I orphaned a setting somewhere?

src/test/resources/application-test.yml
```diff
@@ -17,12 +17,26 @@ buffer:
...
+      in-kubernetes: false
+      namespace-suffix: buffer
+    stairway:
+      jmx-enabled: false
+      pool-max-total: 8
+      pool-max-idle: 8
+      force-clean-start: true
+      migrate-upgrade: true
+      cluster-name: buffer-cluster # FIXME
 terra.common:
```
yonghaoy, line 40: merge with the tracing config. That is also terra.common

src/test/java/bio/terra/buffer/service/resource/FlightManagerTest.java
```diff
@@ -24,6 +24,7 @@
 @DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
 public class FlightManagerTest extends BaseUnitTest {
   @Mock private final StairwayComponent mockStairwayComponent = mock(StairwayComponent.class);
+
```
yonghaoy, line 27: remove newline

src/main/resources/application.yml
yonghaoy: This seems to be wrong to me. I don't think the common-lib's prosperity has Stairway DB setup configs. Those are still in buffer code

src/test/java/bio/terra/buffer/common/PoolIdTest.java
```diff
@@ -10,6 +10,7 @@
   public void storeAndRetrieveFromFlightMap() throws Exception {
     PoolId id = PoolId.create("poolId");
     FlightMap flightMap = new FlightMap();
+    flightMap.put("key", 123);
```
yonghaoy, line 13: that is this?

src/main/resources/application.yml
```diff
@@ -14,18 +14,16 @@ buffer:
     # Whether to update pool when config changes during server startup.
     # Might be useful to avoid potential backward compatibility caused by config change and make tests easier.
     update-pool-on-start: true
-  stairway:
```
yonghaoy, line 17: you should add them back

yonghaoy (changes requested):
yo

src/test/java/bio/terra/buffer/common/PoolIdTest.java
jaycarlton: I thought the map needed something in it to make the comparison meaningful. Otherwise it's just an empty structure, and might not catch as many bugs.

jaycarlton: I just put something in the map to make the comparison non-trivial.

src/main/resources/application.yml
```diff
@@ -19,13 +19,16 @@ buffer:
...
-    # This decides how soon we can fill a pool. Example: if total pools size is 3000 and each flight takes 10 minutes.
-    # having 500 here means it takes 3000 resources / 500 threads * 10minute = 60 minutes to create this pool
-    max-parallel-flights: 500
-    migrate-upgrade: true
-    quiet-down-timeout: 20s
-    terminate-timeout: 5s
+terra:
+  common:
+    stairway:
+      force-clean-start: false
```
yonghaoy, line 25: You also need to add the cluster suffix here?  the value should be buffer-stairwaycluster

yonghaoy: And delete it from test's application.yml since your test doesn't need to override

yonghaoy (approved):

PR #130, opened 2021-03-18
[PF-519]: setup perf and prod buffer pools. 

yonghaoy:


src/main/resources/config/prod/pool_schema.yml
```diff
@@ -0,0 +1,18 @@
+# Resource Buffer Service Pools Schema in prod environment
+---
+poolConfigs:
+- poolId: "cwb_ws_prod_v1"
+  size: 200
```
andy7i, line 5: If there isn't a large cost to having a larger pool size, I'd feel more comfortable if we had more than 200, maybe closer to 2000. Some workshops may include 50-100 or more participants and so this pool may deplete pretty quickly.

I think it may be valuable to have enough projects to last ~1 week in case something happens and it takes a week or two to get the CWB/Terra buffer working again. (@araina2018 do we have data for the # of workspaces created in a week? If we have that data we can find the week where we created the most workspaces and set the pool size to 5-10x that.)

Not related to this ticket, but since the buffer affects users' ability to create workspaces, it may be useful to have monitoring to ensure project creation is succeeding and/or to ensure that we have at least x projects available in the buffer. Do we have any monitoring like this?

src/main/resources/config/prod/pool_schema.yml
```diff
@@ -0,0 +1,18 @@
...
+  resourceConfigName: "cwb_ws_resource_prod_v1"
+- poolId: "aou_ws_prod_v1"
+  size: 200
```
andy7i, line 8: What is the current pool size for AoU Prod? I have some similar concerns for this pool size as mentioned for the CWB pool above.

src/main/resources/config/perf/resource-config/aou_ws_resource_perf_v1.yml
```diff
@@ -0,0 +1,30 @@
+# All Of Us buffered workspace template
+---
+configName: "aou_ws_resource_perf_v1"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "aou-rw-perf"
+    scheme: "RANDOM_CHAR"
+  # test.firecloud.org/terra_perf_aou_perf
+  parentFolderId: "791559643292"
+  billingAccount: "01A82E-CA8A14-367457"
```
andy7i, line 10: AoU and CWB use the same billing account?

src/main/resources/config/perf/resource-config/cwb_ws_resource_perf_v1.yml
```diff
@@ -0,0 +1,31 @@
+# Community Workbench buffered workspace template
+---
+configName: "cwb_ws_resource_perf_v1"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-perf"
+    scheme: "RANDOM_CHAR"
+  # qulity.firecloud.org/quality/CommunityWorkbench
+  parentFolderId: "180789000311"
```
andy7i, line 9: Could we put the perf projects into: `test.firecloud.org/perf/CommunityWorkbench` (quality.firecloud.org is used for FiaBs) instead? 

I think we do want a buffer created in the quality org, but maybe that's outside of the scope of this ticket. (Can we add that here or create a ticket so we don't forget about this?)

src/main/resources/config/prod/resource-config/cwb_ws_resource_prod_v1.yml
```diff
@@ -0,0 +1,31 @@
...
+  - "dns.googleapis.com"
+  - "genomics.googleapis.com"
+  - "lifesciences.googleapis.com"
+  - "logging.googleapis.com"
+  - "monitoring.googleapis.com"
+  - "notebooks.googleapis.com"
+  - "storage-api.googleapis.com"
+  - "storage-component.googleapis.com"
+  network:
+    enableNetworkMonitoring: "false"
```
andy7i, line 31: Maybe outside the scope of this ticket, but CWB should have 2 project pools, one with `enableNetworkMonitoring: "false"` and one with `enableNetworkMonitoring: "true"`.

andy7i (commented):
added some questions

src/main/resources/config/perf/pool_schema.yml
```diff
@@ -1,18 +1,24 @@
...
+  size: 20
+  resourceConfigName: "workspace_manager_v3"
+- poolId: "workspace_manager_v4"
+  size: 20
+  resourceConfigName: "workspace_manager_v4"
+- poolId: "workspace_manager_v5"
+  size: 20
+  resourceConfigName: "workspace_manager_v5"
+- poolId: "cwb_ws_perf_v1"
+  size: 200
```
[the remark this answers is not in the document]
yonghaoy: When the perf test start pointing to perf RBS? If that happens later, we can change the number later. Same for production numbers. 
I just want to have some projects for now, but increase the size later once we decide those numbers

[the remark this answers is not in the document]
ungwudik: > Since the performance tests run many calls to create workspace and clone workspace (and each new workspace will need a google project), I think we'll want a much larger buffer size.
> 
> @ungwudik do you know how many workspaces are created and cloned in a single perf test run?
> 
> We may want to set the buffer at 5-10x this number, because we may kick multiple test runs in a single day.

A single perf test run creates 120 concurrent workspaces and clones 40 workspaces.

src/main/resources/config/prod/pool_schema.yml
yonghaoy: The numbers here are fake, I am expecting to bumo this to a way larger number later. Right now I am trying to have a running prod RBS. 
(2) I do have the number:
https://docs.google.com/spreadsheets/d/12NBQp78G8FGk9TuKmzH1DUQnX0uDC_Q2uyKqbagFtqQ/edit#gid=0
~500/day at peek. 
(3) The monitoring is something else, (AoU's special thing, that for egress alert detection). I don't think CWB need that

yonghaoy: See http://broad.io/resource-buffer-playbook on what dashboard, alerts we have for general monitoring purpose

src/main/resources/config/prod/pool_schema.yml
yonghaoy: ditt, pool size is temporary. Bumping size is a simple step and we can do it later

src/main/resources/config/perf/resource-config/aou_ws_resource_perf_v1.yml
yonghaoy: I am not sure, that is the only prod billing account DevOps gives me. 
Does the current DM use the same BIlling Account? I think so, but not sure

src/main/resources/config/prod/resource-config/cwb_ws_resource_prod_v1.yml
yonghaoy: That is for AoU's egress detection, I through CWB does not do that

src/main/resources/config/perf/resource-config/cwb_ws_resource_perf_v1.yml
yonghaoy: Current FIAB is using this folder:
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/config/tools/resource-config/cwb_ws_resource_fiab_v2.yml#L8

is that wrong?

src/main/resources/config/perf/pool_schema.yml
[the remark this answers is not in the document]
yonghaoy: Thanks! To confirm, how often it is? Is that nightly test? 
And I think AoU has something like 100 concurrent workspaces test?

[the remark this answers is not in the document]
ungwudik: Yes, that's nightly sometime after midnight. Sometimes, we also kick the tests off manually during the day when troubleshooting failures, etc.

Here's the jenkins job for Workspaces perf test: https://fc-jenkins.dsp-techops.broadinstitute.org/job/BVDPWorkspacesNightlyPerformanceTest/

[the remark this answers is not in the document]
ungwudik: I believe AoU runs two jobs daily which create 110 workspaces and clone 110 workspaces in total. @aweng98 can confirm.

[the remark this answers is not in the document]
aweng98: Correct.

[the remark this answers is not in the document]
ungwudik: i.e.
- job 1 creates 100 workspaces and clones 100 workspaces
- job 2 creates 10 workspaces and clones 10 workspaces

https://fc-jenkins.dsp-techops.broadinstitute.org/job/AoUPerfTestRunner/

src/main/resources/config/perf/pool_schema.yml
```diff
@@ -1,18 +1,24 @@
...
+  resourceConfigName: "cwb_ws_resource_perf_v1"
+- poolId: "aou_ws_perf_v1"
+  size: 100
```
aweng98, line 23: Is this for AoU perf testing? If so, I recommend 300 (same as AoU billing buffer in test env)

src/main/resources/config/perf/resource-config/aou_ws_resource_perf_v1.yml
gjuggler: AoU specifies a billing account in the create-billing-project request to Rawls (https://github.com/all-of-us/workbench/blob/master/api/src/main/java/org/pmiops/workbench/firecloud/FireCloudServiceImpl.java#L242). Presumably, Rawls will swap out the billing account to the requested one before it hands a project up to AoU.

AIUI, this billing account ID is only needed to pay for the minimal API usage involved in provisioning a project.

src/main/resources/config/prod/pool_schema.yml
```diff
@@ -0,0 +1,18 @@
...
+  resourceConfigName: "aou_ws_resource_prod_v1"
+- poolId: "aou_ws_preprod_v1"
+  size: 100
+  resourceConfigName: "aou_ws_resource_preprod_v1"
+- poolId: "aou_ws_stable_v1"
+  size: 100
+  resourceConfigName: "aou_ws_resource_stable_v1"
+- poolId: "aou_ws_staging_v1"
+  size: 100
```
aweng98, line 17: Can you increase to 150 (same as AoU billing buffer size in staging env)?

aweng98 (approved):
Thanks for setting up. LGTM.

src/main/resources/config/prod/pool_schema.yml
gjuggler: FWIW, AoU in prod is at 200.

gjuggler (approved):
Configs look fine to me. I have less context on ideal folder locations for e.g. perf projects.

andy7i (approved):
👍  for initial prod release. Final sizing for prod, perf pools TBD. QA pool creation TBD.

PR #131, opened 2021-03-19
[PF-591] Switch to use Common Library to build a DataSource

yonghaoy:


src/main/java/bio/terra/buffer/app/StartupInitializer.java
```diff
@@ -23,13 +23,13 @@ public static void initialize(ApplicationContext applicationContext) {
...
-    if (bufferJdbcConfiguration.isRecreateDbOnStart()) {
-      migrateService.initialize(changelogPath, bufferJdbcConfiguration.getDataSource());
-    } else if (bufferJdbcConfiguration.isUpdateDbOnStart()) {
-      migrateService.upgrade(changelogPath, bufferJdbcConfiguration.getDataSource());
+    if (bufferDatabaseConfiguration.isRecreateDbOnStart()) {
+      migrateService.initialize(changelogPath, bufferDatabaseConfiguration.getDataSource());
+    } else if (bufferDatabaseConfiguration.isUpdateDbOnStart()) {
+      migrateService.upgrade(changelogPath, bufferDatabaseConfiguration.getDataSource());
     }
     applicationContext.getBean(StairwayComponent.class).initialize();
```
jaycarlton, line 34: How are you getting the datasource to StairwayComponnet? In my branch you pass it here along with the hooks and context object you want to use

jaycarlton (approved):

src/main/java/bio/terra/buffer/app/StartupInitializer.java
yonghaoy: Like the currene compomentm do 
`stairwayDatabaseConfiguration.getDataSource()`?

src/main/java/bio/terra/buffer/app/configuration/ApplicationConfiguration.java
```diff
@@ -20,7 +20,8 @@
 @Configuration
 public class ApplicationConfiguration {
   @Bean(JDBC_TEMPLATE)
```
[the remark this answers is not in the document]
yonghaoy: I renamed the bean name, but seems the actual value need to be jdbcTemplate for the same reason.

src/main/java/bio/terra/buffer/app/StartupInitializer.java
```diff
@@ -23,13 +24,15 @@ public static void initialize(ApplicationContext applicationContext) {
     applicationContext.getBean(StackdriverExporter.class).initialize();
```
ddietterich, line 24: You have a logger, but nothing is logged in this class. Seems like being chatty during initialization would be generally helpful.

src/main/java/bio/terra/buffer/app/StartupInitializer.java
```diff
@@ -23,13 +24,15 @@ public static void initialize(ApplicationContext applicationContext) {
...
+    BufferDatabaseConfiguration bufferDatabaseConfiguration =
+        applicationContext.getBean(BufferDatabaseConfiguration.class);
+    BufferDatabaseProperties bufferDatabaseProperties =
+        applicationContext.getBean(BufferDatabaseProperties.class);
 
-    if (bufferJdbcConfiguration.isRecreateDbOnStart()) {
-      migrateService.initialize(changelogPath, bufferJdbcConfiguration.getDataSource());
-    } else if (bufferJdbcConfiguration.isUpdateDbOnStart()) {
-      migrateService.upgrade(changelogPath, bufferJdbcConfiguration.getDataSource());
+    if (bufferDatabaseProperties.isRecreateDbOnStart()) {
```
ddietterich, line 32: Hmm... if this is the common pattern, we should add a method in the Migrator that takes "dropall" and "update" flags directly. Those are what TDR uses as well.

src/main/java/bio/terra/buffer/app/configuration/BaseDataBaseConfiguration.java
```diff
@@ -0,0 +1,22 @@
+package bio.terra.buffer.app.configuration;
+
+import bio.terra.common.db.BaseDatabaseProperties;
+import bio.terra.common.db.DataSourceInitializer;
+import javax.sql.DataSource;
+
+/** Base class for accessing JDBC configuration properties. */
+public class BaseDataBaseConfiguration {
+  private final BaseDatabaseProperties databaseProperties;
```
ddietterich, line 9: nit: IntelliJ says this can be a local variable. Looks to me like you don't need any variable since you are just passing the properties into the initialize.

src/main/java/bio/terra/buffer/app/controller/BufferApiController.java
```diff
@@ -28,7 +28,7 @@
 
   @Autowired
   BufferApiController(
-      BufferJdbcConfiguration jdbcConfiguration,
+      BufferDatabaseConfiguration jdbcConfiguration,
```
ddietterich, line 31: Why do you need this or BufferDatabaseConfiguration or StairwayComponent in this controller? They are unused.

src/main/java/bio/terra/buffer/app/controller/UnauthenticatedApiController.java
```diff
@@ -22,7 +22,7 @@
 
   @Autowired
   UnauthenticatedApiController(
-      BufferJdbcConfiguration jdbcConfiguration, StairwayComponent stairwayComponent) {
+      BufferDatabaseConfiguration jdbcConfiguration, StairwayComponent stairwayComponent) {
```
ddietterich, line 25: I disagree with the approach of treating Stairway as a component for status. If it isn't running, then this service shouldn't pass a liveness check. You are not reporting on any other inner service status.

ddietterich (commented):
Coding nits and an out-of-scope complaint.
Otherwise, fine.

I'm curious why you'd push this forward when Jay's common lib update is a small number of days away. You'll end up re-visiting and I'll end up re-reviewing.

src/main/java/bio/terra/buffer/app/StartupInitializer.java
yonghaoy: filed a https://broadworkbench.atlassian.net/browse/PF-595 to track

src/main/java/bio/terra/buffer/app/StartupInitializer.java
yonghaoy: don't know what to add. But I think it might be helpful to log that the post setup is happening...

src/main/java/bio/terra/buffer/app/StartupInitializer.java
yonghaoy: This helps Jay work I think. The reason I am pushing this PR forward is Jay is trying to figure out how to get DataSource if moving StairwayComponent into common-lib.
After this work, if will may Jay's PR easier.
What he needs to do here is:
`stariwayCompoment.initialize(stairwayDataBaseConfuration.getDataSource());
`
The stariwayCompoment is in common-lib, while stairwayDataBaseConfuration is my new one.

src/main/java/bio/terra/buffer/app/controller/UnauthenticatedApiController.java
yonghaoy: Stairway is not running, buffer be able to still handout resources.  Just can not kick off flights for new resource creation. I think we should let it pass the status check, and altering on lower-resource event.

yonghaoy:
> Coding nits and an out-of-scope complaint.
> Otherwise, fine.
> 
> I'm curious why you'd push this forward when Jay's common lib update is a small number of days away. You'll end up re-visiting and I'll end up re-reviewing.

Thanks Dan! The whole motivation to having the ticket and PR is to make Jay's work simpler. Initialize his PR need to have the DataSource, then he need to move the entire Jdbc base class, and StairwayJdbcConfituraiton into common lib. Then it comes to the question that how the common-lib takes care of DataSource creation. To reduce that ticket's complexity, and Jay's workload, I made this ticket and try to push it in asap to unblock him.  
I left comment on how this change Jay's PR in ons of the other comment. 
Summary: 
Jay's common-lib code will have that StairwayCompoment, with `initialize(DataSource dataSource, Hook...) `accept a DataSource, and that comes from the new class I made here.

src/main/java/bio/terra/buffer/app/configuration/BaseDataBaseConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+import bio.terra.common.db.BaseDatabaseProperties;
+import bio.terra.common.db.DataSourceInitializer;
+import javax.sql.DataSource;
+
+/** Base class for accessing JDBC configuration properties. */
+public class BaseDataBaseConfiguration {
+  private final DataSource dataSource;
+
+  public BaseDataBaseConfiguration(BaseDatabaseProperties databaseProperties) {
+    dataSource = DataSourceInitializer.initializeDataSource(databaseProperties);
```
jaycarlton, line 12: The solution I came up with was to do this inside the StairwayComponent (injecting the right properties into it). It's a little odd to expose a property on a configuration but not make it a bean, but I know there were issues with beanifying it.

src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -439,15 +438,13 @@ private Project assertProjectExists(ResourceId resourceId) throws Exception {
...
-            SUB_NETWORK_LABEL_KEY,
-            SUBNETWORK_NAME,
-            CONFIG_NAME_LABEL_LEY,
-            TEST_CONFIG_NAME),
-        project.getLabels());
+
+    assertThat(
+        project.getLabels().entrySet(),
+        Matchers.hasItems(
+            Map.entry(NETWORK_LABEL_KEY, NETWORK_NAME),
```
jaycarlton, line 445: 👏🏻 much more readable

jaycarlton (approved):
Thank you!

ddietterich (approved):

PR #132, opened 2021-03-25
[no ticket] Stop on error in render-config.sh

kemp-verily:
Without this, the script continues on through errors (like a missing
vault token or missing software like jq) and fails in a fairly opaque
way.  Stopping on the first error makes it more clear what needs to be
fixed.


kemp-verily:
Looks like I'll need to wait until I have a Broad identity to associate for vault access for the unit tests.  I'll try again once I have that in place.

yonghaoy:
For running test locally, no credential needed. 
For running test in GHA, you need to have access on Github secrets. We do have https://github.com/orgs/DataBiosphere/teams/verily-platform-foundations team that should let you are able to run any test in GHA. Idon't think Broad Identity is needed to join the team? @gjuggler

PR #133, opened 2021-03-26
[no ticket]: Create RetryRule object for each flight

yonghaoy:


src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
```diff
@@ -40,26 +41,38 @@ public GoogleProjectCreationFlight(FlightMap inputParameters, Object application
...
     GcpProjectIdGenerator idGenerator =
         ((ApplicationContext) applicationContext).getBean(GcpProjectIdGenerator.class);
-    addStep(new AssertResourceCreatingStep(bufferDao), INTERNAL_DEFAULT_RETRY);
-    addStep(new UndoCreatingDbEntityStep(bufferDao), INTERNAL_DEFAULT_RETRY);
-    addStep(new GenerateProjectIdStep(gcpProjectConfig, idGenerator), CLOUD_API_DEFAULT_RETRY);
-    addStep(new CreateProjectStep(rmCow, gcpProjectConfig), CLOUD_API_DEFAULT_RETRY);
-    addStep(new SetBillingInfoStep(billingCow, gcpProjectConfig), CLOUD_API_DEFAULT_RETRY);
-    addStep(new EnableServicesStep(serviceUsageCow, gcpProjectConfig), CLOUD_API_DEFAULT_RETRY);
-    addStep(new SetIamPolicyStep(rmCow, gcpProjectConfig), CLOUD_API_DEFAULT_RETRY);
+    addStep(new AssertResourceCreatingStep(bufferDao), newInternalDefaultRetryRule());
```
ddietterich, line 44: You could make a single default rule for the flight, but this way is also fine.

ddietterich (approved):
LGTM - one comment

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
yonghaoy: right, that is used in multiple clases, so I create a static helper in a common place

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
```diff
@@ -40,26 +41,38 @@ public GoogleProjectCreationFlight(FlightMap inputParameters, Object application
...
+    addStep(new UndoCreatingDbEntityStep(bufferDao), newInternalDefaultRetryRule());
```
jaycarlton, line 45: It might be easier to read this if you made helper methods like `addDefaultInternalRetryStep()` and `addCloudRetryStep()`, and they would only take two args.

src/main/java/bio/terra/buffer/service/resource/flight/StepUtils.java
```diff
@@ -12,19 +12,22 @@
...
 /** Utilities used in Stairway steps. */
 public class StepUtils {
+
   /**
    * The stairway retry rule for GCP API calls. Use longer wait time because cloud API quota,
    * outrage, and other unknown issues.
    */
-  public static final RetryRuleFixedInterval CLOUD_API_DEFAULT_RETRY =
-      new RetryRuleFixedInterval(/* intervalSeconds =*/ 60, /* maxCount =*/ 10);
+  public static RetryRuleFixedInterval newCloudApiDefaultRetryRule() {
```
jaycarlton, line 20: Do you need to create a new one every time? If they all have the same parameters, why have more than one?

jaycarlton (approved):

src/main/java/bio/terra/buffer/service/resource/flight/StepUtils.java
yonghaoy: that is the issue, the same instance is shared by all instances. So I have to create a new one everytime

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectCreationFlight.java
yonghaoy: I feel retry rule was a small thing, and worth to be splitted out in its own small helper method?

jaycarlton: All this would do would be to consolidate a bit more. Saves about 7 keystrokes and one method arg each time.

Or you could shorten it by having another class like `RetryRules` with members `cloudApiDefault()` and `internalDefault()`.

PR #134, opened 2021-03-29
[PF-619] Disable deleting resource when exceeding pool size

yonghaoy:
This is the short-term of 'Primary Job' in Buffer doc:
https://docs.google.com/document/d/1rSIXOrDQYCgptHZIxnxXjM-nHjb_ohxxNGKAl5ZarKw/edit#heading=h.5inn55b6pv26

As Doug suggested "For: down sizing... Currently we don't really need to actively reduce the size"
This also increases the frequency of scheduling service runs.

### DIFF
--- a/src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
+++ b/src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
@@ -25,7 +25,9 @@ import org.springframework.transaction.annotation.EnableTransactionManagement;
 @ConfigurationProperties(prefix = "buffer.primary")
 public class PrimaryConfiguration {
 
-  /** Whether to run the scheduler to periodically. */
+  /**
+   * Whether to run the scheduler to periodically.
+   */
   private boolean schedulerEnabled;
 
   /**
@@ -43,7 +45,7 @@ public class PrimaryConfiguration {
    * is no use case that we need to actively reduce pool size. Resource is only need to be deleted
    * when pool is inactivated.
    */
-  private boolean deleteExcessResources;
+  private boolean deleteExcessResources = false;
 
   /**
    * How many resource creation flights for a pool to process simultaneously because because we
--- a/src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
+++ b/src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
@@ -11,21 +11,40 @@ import org.springframework.transaction.annotation.EnableTransactionManagement;
  *
  * <p>Rate limiting: The bottleneck for GCP quota comes from ServiceUsage.batchEnable API. By
  * default, it is 20/100s. Here is how we estimate the time: A Creation flight takes 600~900
- * seconds. If scheduler runs every 10 seconds, in 900 seconds it schedulers runs 70 times. If we
+ * seconds. If a scheduler runs every 10 seconds, in 900 seconds it schedulers runs 70 times. If we
  * have 4 pools, it schedules ({@code resourceCreationPerPoolLimit} * 4) = 4 flights. In total 70
  * times run, 280 flights is scheduled. We estimate this number should works when pool number is
- * 1~10. If we see more errors, we will comeback and revise those configs.
+ * 1~10. If we see more errors, we will comeback and revise those configs. TODO(PF-617): The current
+ * 'primary' scheduling service is actually running in all instances, figure out a way to make this
+ * only runs in one pod.
  */
 @Component
 @EnableConfigurationProperties
 @EnableTransactionManagement
 @ConfigurationProperties(prefix = "buffer.primary")
 public class PrimaryConfiguration {
-  /** Whether to run the scheduler to periodically. */
+
+  /**
+   * Whether to run the scheduler to periodically.
+   */
   private boolean schedulerEnabled;
 
-  /** How often to query for flights to submit. */
-  private Duration flightSubmissionPeriod = Duration.ofSeconds(10);
+  /**
+   * How often to query for flights to submit.
+   *
+   * <p> Sets this to be a larger number because most of the service is deployed in a
+   * multi-instance world, and there will be more than one scheduling service running.
+   */
+  private Duration flightSubmissionPeriod = Duration.ofSeconds(20);
+
+  /**
+   * Whether to delete resource when resource count exceeds the pool size.
+   *
+   * <p>This to prevent multiple instances concurrent scheduler deleting extra resources. Also this
+   * is no use case that we need to actively reduce pool size. Resource is only need to be deleted
+   * when pool is inactivated.
+   */
+  private boolean deleteExceedPoolSizeResource = false;
 
   /**
    * How many resource creation flights for a pool to process simultaneously because because we
@@ -69,5 +88,13 @@ public class PrimaryConfiguration {
 
   public void setResourceDeletionPerPoolLimit(int resourceDeletionPerPoolLimit) {
     this.resourceDeletionPerPoolLimit = resourceDeletionPerPoolLimit;
+  }
+
+  public boolean isDeleteExceedPoolSizeResource() {
+    return deleteExceedPoolSizeResource;
+  }
+
+  public void setDeleteExceedPoolSizeResource(boolean deleteExceedPoolSizeResource) {
+    this.deleteExceedPoolSizeResource = deleteExceedPoolSizeResource;
   }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
```diff
@@ -11,21 +11,40 @@
...
 
-  /** How often to query for flights to submit. */
-  private Duration flightSubmissionPeriod = Duration.ofSeconds(10);
+  /**
+   * How often to query for flights to submit.
+   *
+   * <p> Sets this to be a larger number because most of the service is deployed in a
+   * multi-instance world, and there will be more than one scheduling service running.
+   */
+  private Duration flightSubmissionPeriod = Duration.ofSeconds(20);
```
wchamber, line 38: why not 3x? Does this need ot scale with the number of instances?

src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
```diff
@@ -11,21 +11,40 @@
...
+
+  /**
+   * Whether to delete resource when resource count exceeds the pool size.
+   *
+   * <p>This to prevent multiple instances concurrent scheduler deleting extra resources. Also this
+   * is no use case that we need to actively reduce pool size. Resource is only need to be deleted
+   * when pool is inactivated.
+   */
+  private boolean deleteExceedPoolSizeResource;
```
wchamber, line 47: how about `deleteExcessResources`?

src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
```diff
@@ -11,21 +11,40 @@
...
+   * multi-instance world, and there will be more than one scheduling service running.
+   */
+  private Duration flightSubmissionPeriod = Duration.ofSeconds(20);
+
+  /**
+   * Whether to delete resource when resource count exceeds the pool size.
+   *
+   * <p>This to prevent multiple instances concurrent scheduler deleting extra resources. Also this
+   * is no use case that we need to actively reduce pool size. Resource is only need to be deleted
+   * when pool is inactivated.
```
wchamber, line 45: Should we default this to false? Any thoughts on if we even want to keep the ability to delete excess resources? Would we want to revert this when we have PF-617?

src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
```diff
@@ -11,21 +11,40 @@
  *
  * <p>Rate limiting: The bottleneck for GCP quota comes from ServiceUsage.batchEnable API. By
  * default, it is 20/100s. Here is how we estimate the time: A Creation flight takes 600~900
- * seconds. If scheduler runs every 10 seconds, in 900 seconds it schedulers runs 70 times. If we
+ * seconds. If a scheduler runs every 10 seconds, in 900 seconds it schedulers runs 70 times. If we
  * have 4 pools, it schedules ({@code resourceCreationPerPoolLimit} * 4) = 4 flights. In total 70
  * times run, 280 flights is scheduled. We estimate this number should works when pool number is
- * 1~10. If we see more errors, we will comeback and revise those configs.
+ * 1~10. If we see more errors, we will comeback and revise those configs. TODO(PF-617): The current
```
wchamber, line 17: nit: add a `<p>` before the TODO

wchamber (changes requested):
I think you should file a ticket for this and discuss why we're doing this short term change and why it's not a good long term solution.

src/test/java/bio/terra/buffer/service/resource/FlightSchedulerTest.java
```diff
@@ -138,7 +159,22 @@ public void scheduleDeactivationFlights_poolDeactivated() throws Exception {
   }
 
   @Test
-  public void scheduleDeactivationFlights_poolSizeReduced() throws Exception {
+  public void scheduleDeactivationFlights_enableDeletion() throws Exception {
+    newPoolWithResourceCount(1, ImmutableMultiset.of(ResourceState.READY, ResourceState.READY));
```
jaycarlton, line 163: why two states if there's only one resource?

src/test/java/bio/terra/buffer/service/resource/FlightSchedulerTest.java
```diff
@@ -138,7 +159,22 @@ public void scheduleDeactivationFlights_poolDeactivated() throws Exception {
...
+
+    PrimaryConfiguration primaryConfiguration = newPrimaryConfiguration();
+    primaryConfiguration.setDeleteExceedPoolSizeResource(true);
+    initializeScheduler(primaryConfiguration);
+
+    Thread.sleep(4000);
```
jaycarlton, line 169: use `TimeUnit.SECONDS(4).sleep()`.

src/test/java/bio/terra/buffer/service/resource/FlightSchedulerTest.java
```diff
@@ -138,7 +159,22 @@ public void scheduleDeactivationFlights_poolDeactivated() throws Exception {
...
+
+    verify(flightManager, times(1))
```
jaycarlton, line 171: you may omit `times(1)` as it's the default.

src/test/java/bio/terra/buffer/service/resource/FlightSchedulerTest.java
```diff
@@ -151,6 +187,18 @@ public void scheduleDeactivationFlights_poolSizeReduced() throws Exception {
     verify(flightManager, never()).submitCreationFlight(any(Pool.class));
   }
 
+  @Test
+  public void notScheduleDeactivationFlights_disableDeletion() throws Exception {
+    newPoolWithResourceCount(1, ImmutableMultiset.of(ResourceState.READY, ResourceState.READY));
+
+    initializeScheduler();
+    Thread.sleep(4000);
```
jaycarlton, line 195: indicate how this time relates to other configured times.

Also, can you mock the clock or thread class, or do you need a real sleep here?

jaycarlton (approved):
The parts I understand look good. I don't have much context.

src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
yonghaoy: Changed to be 3x. Feels we don't need to change this frequently every-time number instance change, but we many want to change this if number of instances changes a lot(from 1 to 10?). 
I have a personal env (buffertest) that has multiple(2) instance enabled without this change. Seems it is able to create 1500 resources successfully without little 429 errors.  So I guess 10s still works with 2 instances, changing to 30 is an improvement based on our calculation above.
But all are estimation based on our calculation.

src/test/java/bio/terra/buffer/service/resource/FlightSchedulerTest.java
yonghaoy: pool size is 1, but there are two READY resources, that is why the delete happens

src/test/java/bio/terra/buffer/service/resource/FlightSchedulerTest.java
yonghaoy: I need a real sleep to make the scheduler really runs.

src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
wchamber: 2) I don't really follow. If you're trying to figure out what the service does by default, aren't you going to be looking at the code? If we default to false, we don't have to touch any of the `application.yml`, which seems like less understanding of the code.

wchamber (approved):
lgtm, but please consider changing the default value of `deleteExcessResources` so we don't need to set values in the application yml files.

src/main/java/bio/terra/buffer/app/configuration/PrimaryConfiguration.java
yonghaoy: This is based on the most recent work with Jay when he moves StairwayCompoment into common-lib.
Having default value for spring config in java code seems not to be a well-know practice. People with spring experience seems will go to appliction.yml to find all values.  Java -> Application/.yaml -> helm is 3 levels lookup. While removing Java having default value, we get one less level. 

But to me, I don't have strong preference, changing to use java file hold default value

PR #138, opened 2021-04-05
[PF-561]: Few minor improvements: Add index on resource table, avoid on unnecessary db query

yonghaoy:
1. AoU also has a similar buffer service. Recently we observe slowness in the AoU buffer service when handout resource. That is caused by too many non-READY resource slows down query. As our resource table is heavily used by handout, schedulling service, I think it make sense to add index.
2. Also, remove the unnecessary query when deleting resources(a Pool is deactivated for a while, we are still trying to make a query on that table everytime).
3. Fix env variable in local run script.

```
testdb=> select *
from pg_indexes where tablename = 'resource';
 schemaname | tablename |      indexname       | tablespace |                                   indexdef                                    
------------+-----------+----------------------+------------+-------------------------------------------------------------------------------
 public     | resource  | resource_pkey        |            | CREATE UNIQUE INDEX resource_pkey ON public.resource USING btree (id)
 public     | resource  | resource_state_index |            | CREATE INDEX resource_state_index ON public.resource USING btree (state DESC)
```

### DIFF
--- a/local-dev/run_local.sh
+++ b/local-dev/run_local.sh
@@ -6,7 +6,6 @@ export BUFFER_DB_RECREATE_DB_ON_START=true
 export BUFFER_DATABASE_NAME=testdb
 export BUFFER_STAIRWAY_DB_USERNAME=dbuser_stairway
 export BUFFER_STAIRWAY_DB_PASSWORD=dbpwd_stairway
-export BUFFER_STAIRWAY_DB_FORCE_CLEAN_START=true
 export BUFFER_STAIRWAY_DATABASE_NAME=testdb_stairway
 export GOOGLE_APPLICATION_CREDENTIALS="$(dirname $0)"/../src/test/resources/rendered/sa-account.json
 export BUFFER_CRL_TESTING_MODE=true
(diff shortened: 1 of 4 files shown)


wchamber (commented):
To follow our compliance requirements, you need to file a ticket for this change.

Do you have any performance numbers to show that adding the index helps? I'm not really worried about it hurting, but it'd be nice to have some way to see that it's helping.

wchamber:
We might also see slowness because we never delete resources from our tables. Having a time window after which we delete handed out/deleted resources might also help. Hard to say without some metric/test.

yonghaoy:
> To follow our compliance requirements, you need to file a ticket for this change.
> 
> Do you have any performance numbers to show that adding the index helps? I'm not really worried about it hurting, but it'd be nice to have some way to see that it's helping.

I don't have number for his, now all RBS are functioning well. This change is based on the recent issue happens in AoU. AoU starts seeing slowness >1 year after launch. So I think RBS would be fine for ~1 year. But I thought RBS tables are very small, and seems it couldn't hurt to add that now, , so I just make the change. I am fine of doing that later when we see the slowness.  But other part of this PR, remove the unnecessary db query when pool deactivated, I do want that to be fixed. 

Also, for cleanup old entities, Dan brought that up in RBS design doc, and I put that in a P2 feature, then I forgot it.  
Filed https://broadworkbench.atlassian.net/browse/PF-654

wchamber (approved):

jmczerk (approved):

PR #139, opened 2021-04-06
[PF-655] Merge buffer test runner's gradle project into Buffer's main project

yonghaoy:
Initially having two gradle project was a short-term solution to make test runner work easier. 
Ideally there should be one gralde projects and easier for local development. No need to seperate them at all.


wchamber (approved):
Worth explaining in the PR description why this is a good thing.

PR #141, opened 2021-04-08
[PF-652]Adjust RBS Pools

yonghaoy:
1. Setup pool size per CI's feedback 
2. setup quality for tools RBS
3. Remove some unused pools.

### DIFF
--- a/src/main/resources/config/perf/pool_schema.yml
+++ b/src/main/resources/config/perf/pool_schema.yml
@@ -1,9 +1,6 @@
 # RBS Pools Schema for running buffer service performance test.
 ---
 poolConfigs:
-  - poolId: "resource_perf_v2"
-    size: 1500
-    resourceConfigName: "resource_perf_v2"
   - poolId: "workspace_manager_v3"
     size: 20
     resourceConfigName: "workspace_manager_v3"
@@ -14,8 +11,8 @@ poolConfigs:
     size: 20
     resourceConfigName: "workspace_manager_v5"
   - poolId: "cwb_ws_perf_v1"
-    size: 200
+    size: 500
     resourceConfigName: "cwb_ws_perf_v1"
   - poolId: "aou_ws_perf_v1"
-    size: 300
+    size: 500
     resourceConfigName: "aou_ws_perf_v1"
--- a/src/main/resources/config/prod/pool_schema.yml
+++ b/src/main/resources/config/prod/pool_schema.yml
@@ -2,10 +2,10 @@
 ---
 poolConfigs:
   - poolId: "cwb_ws_prod_v1"
-    size: 500
+    size: 3000
     resourceConfigName: "cwb_ws_prod_v1"
   - poolId: "aou_ws_prod_v1"
-    size: 300
+    size: 500
     resourceConfigName: "aou_ws_prod_v1"
   - poolId: "aou_ws_preprod_v1"
     size: 100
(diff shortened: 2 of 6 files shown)


src/main/resources/config/prod/pool_schema.yml
```diff
@@ -2,10 +2,10 @@
 ---
 poolConfigs:
   - poolId: "cwb_ws_prod_v1"
-    size: 500
+    size: 3000
```
yonghaoy, line 5: I will do 5k in next release

wchamber: Are you just waiting to see what happens after the move to 3k?

wchamber (approved):
Thanks for the cleanup too!
(Note to future me, as Yonghao does, make sure to check the helmfile values for [WM in different environments](https://github.com/broadinstitute/terra-helmfile/blob/35a9fa819ff552f2768982a660af6d19b1e7cca9/terra/values/workspacemanager/live/perf.yaml#L24) before deleting pools to make sure they're not still used.)

src/main/resources/config/prod/pool_schema.yml
yonghaoy: that is one reason. Another reason is alert will be fired if pool size <15% for 20 minutes...so bump 5 times each time can avoid that alert.

src/main/resources/config/perf/pool_schema.yml
```diff
@@ -14,8 +11,8 @@ poolConfigs:
     size: 20
     resourceConfigName: "workspace_manager_v5"
   - poolId: "cwb_ws_perf_v1"
-    size: 200
+    size: 500
```
andy7i, line 14: nit: I'm noticing that version # of the pools are different. Could we update the poolId so that the version numbers are consistent?

```
  - poolId: "cwb_ws_dev_v2"
  - poolId: "cwb_ws_perf_v1"
  - poolId: "cwb_ws_prod_v1"
  - poolId: "cwb_ws_fiab_v2"
  - poolId: "cwb_ws_quality_v1"
 ```
```
  - poolId: "cwb_ws_dev_v2"
  - poolId: "cwb_ws_perf_v2"
  - poolId: "cwb_ws_prod_v2"
  - poolId: "cwb_ws_fiab_v2"
  - poolId: "cwb_ws_quality_v2"
 ```

This would make updating the Terra configs much cleaner so that we can programmatically set the poolId to something like `cwb_ws_{{$environment}}_v2` instead of updating each env's `poolId` manually.

it might be helpful to do this for the aou pools as well, since we'll need to add these at some point.

src/main/resources/config/prod/pool_schema.yml
andy7i: just curious, why not test 3k in a non-prod env first?

andy7i (approved):
couple questions/suggestions, but not blocking

src/main/resources/config/prod/pool_schema.yml
yonghaoy: 3k was tested in perf already, and seems working fine

src/main/resources/config/perf/pool_schema.yml
yonghaoy: Switched quality to v2. 
But I don't think `cwb_ws_{{$environment}}_v2` is a good idea because it is possible/common that we want to use different pool versions. e.g., we change folder for test pool but we don't want to bump pools in all env. Or we want have different switch time(deploy a new pool or backfill prod pool seems takes longer). I'd suggest that we do not assume `v_num` is the common pool suffix.

andy7i: we talked faces and talked about setting each env's poolId config individually to provide flexibility. We also talked about setting the pool version #s so that it's clear at a glance which pools are similar or which use different resource config settings. (for example, cwb_ws_dev_v2 added lifesciences APIs. with this knowledge, you know that all other v2 cwb pools also have the lifesciences API and the v1 cwb pools do not)

PR #142, opened 2021-04-08
[no ticket] Stop on error in render-config.sh

kemp-verily:
Without this, the script continues on through errors (like a missing
vault token or missing software like jq) and fails in a fairly opaque
way.  Stopping on the first error makes it more clear what needs to be
fixed.


kemp-verily:
Closing in favor of #143 ... working from a fork is not supported since the github workflow uses secrets.

PR #144, opened 2021-04-14
[PF-228] Add support for two word project IDs

kemp-verily:
This adds support for configuring GCP style adjective-noun project IDs
using a pre-populated list of word pairs.

The word list was generated using the `wn` tool from Princeton's Wordnet
project.

I started with a seed list of positive adjectives (happy, fast, etc) and
used `wn` to generate synonyms.  For nouns, I used a seed list of
categories (fruit, vegetable, jewel) and used `wn` to generate all
words with an "is a kind of" relation.

I filtered the words based on length, and then manually reviewed and
filtered the resulting lists before generating the pairwise combinations.

For posterity, another approach I tried and ultimately abandonded was to
use Sentinet to perform automated review of the lists.  There were too
many abberations in the sentiment data to make this a viable approach.
By starting with seed lists of words instead I was able to generate
relatively small lists that I could review manually.


src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java
```diff
@@ -41,4 +51,23 @@ private String generateRandomId() {
...
   }
+
+  private static class WordPairs {
+    static final ArrayList<String> words = loadWords();
+    static final Random random = new Random();
+
+    static ArrayList<String> loadWords() {
+      ArrayList<String> words = new ArrayList<String>();
+      Scanner scanner =
+          new Scanner(WordPairs.class.getResourceAsStream("/random_project_prefixes.txt"));
```
yonghaoy, line 62: This is an another slightly different pattern in Buffer to use ClassLoader to loader file.
https://source.corp.google.com/github/DataBiosphere/terra-resource-buffer/src/main/java/bio/terra/buffer/service/pool/PoolConfigLoader.java;l=48?q=file:%5EDatabiosphere%2F%20package:%5Egithub$%20getResourceAsStream&m=100&type=cs

That code path also read json file into a java class. But your use cases is just word, we may don't need a pojo class for this.` List<String>` LGTM. It is up to you.

src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java
```diff
@@ -27,6 +33,10 @@ public String generateId(ProjectIdSchema projectIdSchema) {
     }
 
     switch (projectIdSchema.getScheme()) {
+      case TWO_WORDS_NUMBER:
+        generatedId += WordPairs.getWithRandomSuffix();
+        break;
+
       case RANDOM_CHAR:
```
wchamber, line 40: optional nit: consider removing the `default` case. We should have all enum cases covered here. go/java-practices/switch#default

src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java
```diff
@@ -41,4 +51,23 @@ private String generateRandomId() {
         .toString()
         .substring(0, RANDOM_ID_SIZE);
   }
+
+  private static class WordPairs {
+    static final ArrayList<String> words = loadWords();
+    static final Random random = new Random();
+
+    static ArrayList<String> loadWords() {
+      ArrayList<String> words = new ArrayList<String>();
```
wchamber, line 60: java nit: Prefer to declare the most generic type relevant for the variable and function parameters. e.g. `List` unless it's important that the type is an `ArrayList`. (Or even `Collection` but that's happens so infrequently its rarely done.)
```
static final List<String> words = loadWords();
static List<String> loadWords() {
  List<String> words = new ArrayList<>(); // the <String> can be inferred, so omitted it's new ArrayList<>()
```

wchamber (approved):
LGTM. I did not review all 7k word pairs for offensiveness :)
I would like a `luminous-lettuce`.

src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java
kemp-verily: I think I'll leave as is since it's just a simple list but thanks for the pointer.

src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java
kemp-verily: Done

src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java
kemp-verily: Thanks; I forgot this bit of trivia :)

yonghaoy (approved):

PR #145, opened 2021-04-15
[PF-659] Setup vpc_sc pools for prod & perf & dev

yonghaoy:
Also delete all aou pools
Background discussion:  https://broadinstitute.slack.com/archives/C01K149FSB1/p1617889518008300

1. For VPC-SC pools, there will be one central pool called vpc_sc_unclaimed as placeholder.
2. Rawls grab project from this pool then assign this to right folder by service perimeter folder.


src/main/resources/config/prod/resource-config/vpc_sc_v1.yml
```diff
@@ -1,12 +1,12 @@
...
+configName: "vpc_sc_v1"
 gcpProjectConfig:
   projectIdSchema:
-    prefix: "aou-rw"
+    prefix: "terra-vpc-sc"
     scheme: "RANDOM_CHAR"
-  # firecloud.org/terra_dev_aou_prod
-  parentFolderId: "300547813290"
+  # firecloud.org/prod/for_vpc_sc_unclaimed
+  parentFolderId: "TODO"
```
yonghaoy, line 9: @dvoet Hi only have permissions to create folders in test.fc.org. 
Can you help to also create a folder in fc.org.prod? Would be great if we can make this PR in next release train. Thanks!

src/main/resources/config/prod/resource-config/vpc_sc_v1.yml
```diff
@@ -1,12 +1,12 @@
-# All Of Us buffered workspace template
+# Projects with VPC-SC configuration
 ---
-configName: "aou_ws_prod_v1"
+configName: "vpc_sc_v1"
 gcpProjectConfig:
   projectIdSchema:
-    prefix: "aou-rw"
+    prefix: "terra-vpc-sc"
```
yonghaoy, line 6: Not sure what could be project name prefix, but `terra-vpc-sc-{env} `does not sound perfect

src/main/resources/config/dev/resource-config/vpc_sc_v1.yml
```diff
@@ -1,12 +1,12 @@
-# All Of Us buffered workspace template
+# # Projects with VPC-SC configuration
```
dvoet, line 1:
```suggestion
# Projects with VPC-SC configuration
```

dvoet (approved):

src/main/resources/config/prod/resource-config/vpc_sc_v1.yml
```diff
@@ -21,6 +21,7 @@ gcpProjectConfig:
     - "dataproc.googleapis.com"
     - "dns.googleapis.com"
     - "genomics.googleapis.com"
+    - "lifesciences.googleapis.com"
```
wchamber, line 24: why is this being added now? Was it an oversight to not have it before? Why does only prod have it?

src/main/resources/config/prod/resource-config/vpc_sc_v1.yml
wchamber: @dvoet What's the right prefix to use here, or does it not matter to CWB what the projectIds look like?

wchamber:
Could you also summarize the discussion from the slack thread in this PR description?

src/main/resources/config/dev/pool_schema.yml
```diff
@@ -1,14 +1,13 @@
 # Resource Buffer Service Pools Schema in dev environment
 ---
 poolConfigs:
-  # Pool for GCP projects to be used by workspaces in the AoU test environment. Rawls is the proximal owner of these
-  # projects; AoU sits as a layer atop Rawls.
-  - poolId: "aou_ws_test_v2"
```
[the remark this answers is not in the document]
yonghaoy: All cwb and aou pools are not used at all

src/main/resources/config/prod/resource-config/vpc_sc_v1.yml
yonghaoy: Good catch. This should be added everywhere(for using cromewell, added it perf as well)

src/main/resources/config/prod/resource-config/vpc_sc_v1.yml
dvoet: I think what you have is fine

src/main/resources/config/perf/resource-config/vpc_sc_v1.yml
```diff
@@ -1,12 +1,12 @@
...
   projectIdSchema:
-    prefix: "aou-rw-perf"
+    prefix: "terra-vpc-sc-perf"
     scheme: "RANDOM_CHAR"
-  # test.firecloud.org/terra_perf_aou_perf
-  parentFolderId: "791559643292"
+  # test.firecloud.org/perf/for_vpc_sc_unclaimed
+  parentFolderId: "533417334224"
   billingAccount: "01A82E-CA8A14-367457"
   enabledApis:
```
wchamber, line 11: I don't see lifesciences api enabled on this one. Please double check you've got them all and they're all in sync.

wchamber (approved):

src/main/resources/config/perf/resource-config/vpc_sc_v1.yml
yonghaoy: forgot to push, my bad

PR #150, opened 2021-05-12
Add config for data repo 

snf2ye:


src/main/resources/config/dev/resource-config/datarepo_v1.yml
```diff
@@ -0,0 +1,18 @@
+---
+configName: "datarepo_v1"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "datarepo-dev"
+    scheme: "RANDOM_CHAR"
+  parentFolderId: "270278425081" #data.test-terra.bio/repos/jade-dev
```
snf2ye, line 7: This is the folder we currently use. But, I think we probably need to switch to something in the firecloud org with a larger project quota.

src/main/resources/config/dev/resource-config/datarepo_v1.yml
```diff
@@ -0,0 +1,18 @@
...
+  enabledApis:
+    - "bigquery-json.googleapis.com"
+    - "firestore.googleapis.com"
+    - "firebaserules.googleapis.com"
+    - "storage-component.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "cloudbilling.googleapis.com"
+    - "appengine.googleapis.com"
+  network:
+    enableNetworkMonitoring: "false"
```
snf2ye, line 18: What are these settings for?

src/main/resources/config/dev/resource-config/datarepo_v1.yml
```diff
@@ -0,0 +1,18 @@
+---
+configName: "datarepo_v1"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "datarepo-dev"
+    scheme: "RANDOM_CHAR"
+  parentFolderId: "270278425081" #data.test-terra.bio/repos/jade-dev
+  billingAccount: "01A82E-CA8A14-367457"
```
yonghaoy, line 8: Is that okay to use broad billing account?

src/main/resources/config/dev/resource-config/datarepo_v1.yml
yonghaoy: That is for VPC-SC(Service Perimeter) flow logs. I don't think you need this.

src/main/resources/config/dev/resource-config/datarepo_v1.yml
snf2ye: I checked our settings and this is what we use for our dev environment. Is there any reason why this would be an issue? It also looks like it's the same what WSM uses.

yonghaoy: Then I think it should be fine. I was not clear how Broad share billing accounts between each teams.

tlangs (approved):

yonghaoy (commented):
The test fails on https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/service/pool/GcpResourceConfigValidator.java#L21
(initially we want to make sure compute and storage API are enabled in all Terra project). If TDR I can make the change to relax the restriction or we also enable compute API in TDR project. I am fine with both options

tlangs:
What's the reason for the compute API needing to be enabled?

yonghaoy:
> What's the reason for the compute API needing to be enabled?

All GCP projects we are buffering is Terra  `workspace` project, and compute is the essential API, so we add this as restriction. But I guess it is not true now with TDR case.

tlangs:
This is definitely not a requirement for TDR projects.

yonghaoy:
> This is definitely not a requirement for TDR projects.

https://github.com/DataBiosphere/terra-resource-buffer/pull/157 PTAL!

tlangs:
Hi @yonghaoy, I'm not sure why the test is still failing.

tlangs:
or, does this need to merge first? https://github.com/broadinstitute/terraform-ap-deployments/pull/328

yonghaoy:
> > This is definitely not a requirement for TDR projects.
> 
> #157 PTAL!

Hi, you need to plan & apply the terraform change first. (I am doing it for you now).
The test failed for different reason, it should be fixed in my suggested commit

tlangs:
Ugh quotes...

yonghaoy (approved):
Pool config LG! Just wait until the terraform PR applied fist

tlangs:
When will we know that the terraform PR has been applied?

yonghaoy:
> When will we know that the terraform PR has been applied?

Terraform will be applied right after you type atlantis apply in that PR comment(like what I just did). 
Now I think we are ready to go

tlangs:
Great! I'm not authorized to merge this PR, so someone else will have to.

yonghaoy:
> Great! I'm not authorized to merge this PR, so someone else will have to.

I think we have to wait until the test pass(the test takes ~1hr as they are creating GCP projects)

tlangs:
Tests pass! still can't merge 😢

yonghaoy:
> Tests pass! still can't merge 😢

Then we have to 

> Tests pass! still can't merge 😢
I merged that for you. 
I think we grant https://github.com/orgs/DataBiosphere/teams/broadwrite/members permission to merge this repo. Not sure how you can join that github team.

tlangs:
thanks for merging it!

PR #151, opened 2021-05-14
[CA-1139] Increase RBS Pool sizes to accommodate integration tests

s-rubenstein:
This should increase the RBS dev pool sizes to 1000 (if I am understanding the configs correctly). This is an attempt to make the dev pool sizes large enough for the integration tests to run comfortably. I figure starting with 1000, we can increase by more over time as we see what our demand looks like.

### DIFF
--- a/src/main/resources/config/dev/pool_schema.yml
+++ b/src/main/resources/config/dev/pool_schema.yml
@@ -2,12 +2,12 @@
 ---
 poolConfigs:
   - poolId: "cwb_ws_dev_v2"
-    size: 300
+    size: 1000
     resourceConfigName: "cwb_ws_resource_dev_v2"
   - poolId: "workspace_manager_v6"
-    size: 300
+    size: 1000
     resourceConfigName: "workspace_manager_v6"
   - poolId: "vpc_sc_v1"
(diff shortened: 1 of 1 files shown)


src/main/resources/config/dev/pool_schema.yml
```diff
@@ -2,12 +2,12 @@
...
-    size: 300
+    size: 1000
     resourceConfigName: "cwb_ws_resource_dev_v2"
   - poolId: "workspace_manager_v6"
-    size: 300
+    size: 1000
     resourceConfigName: "workspace_manager_v6"
   - poolId: "vpc_sc_v1"
-    size: 300
+    size: 1000
```
andy7i, line 11: not sure if these 2 are used enough to warrant an increase but I'm not too concerned about increasing it either

andy7i (approved):
LGTM. One comment but not blocking.

yonghaoy (changes requested):
Per offline sync. Fiab is actually using `tools` RBS. And the error is not caused by pool limit

gpolumbo-broad (approved):

PR #156, opened 2021-05-25
[PF-615] Bump TCL version, handle new exception from DB retry

zloery:
The change https://github.com/DataBiosphere/terra-common-lib/pull/57 in Terra Common Library changed how `DatabaseRetryUtils.executeAndRetry` handles an operation that exceeds the maximum number of retries. Previously, it threw an InterruptedException. Now, it surfaces the most recent DB failure. This change handles the new exception in the same way as the old one, so RBS behavior is unchanged. This is the only existing usage (across all services) of this TCL method.
I think we want to surface the same error as before instead of the raw database error, though because `DataAccessException` is unchecked, we could do either.


src/main/java/bio/terra/buffer/service/pool/PoolService.java
```diff
@@ -124,7 +125,7 @@ private Resource handoutResourceTransactionally(
                 "No resource is ready to use at this moment for pool: %s. Please try later",
                 poolId));
       }
-    } catch (InterruptedException e) {
+    } catch (InterruptedException | DataAccessException e) {
```
jaycarlton, line 128: I'd include the caught exception in the one you're throwing so we don't lose anything potentially useful.

jaycarlton (approved):

yonghaoy (approved):

PR #157, opened 2021-06-02
[PF-806]remove compute api restriction

yonghaoy:
Initially we have restriction that all pool config must have compute API enabled because those projects power “Terra workspace” which means user need compute. But with Buffer Service adopted by Data Repo, that’s true anymore. 

We may want to remove this restriction and unblock TDR integration


tlangs (approved):

wchamber (approved):
I'd be ok with removing `REQUIRED_SERVICES` entirely, but up to you. No strong opinions. I think you'll notice pretty quick if your projects don't have the right services enabled.

PR #159, opened 2021-06-03
[no ticket]: Enable Compute API for TDR project

yonghaoy:


tlangs (approved):

tlangs (changes requested):
I'm wondering if we need the firewall created in projects which won't be using compute.

tlangs (approved):
Lets do this for now. We'd like to not have to enable the compute API in the future. An option to disable the creation of the firewall, since we don't need compute, would be much appreciated.

yonghaoy:
> I'm wondering if we need the firewall created in projects which won't be using compute.

I know for the general Broad security suggestion, that we should delete the default network, add related by a high security one. Not sure if TDR's project need to follow this. 
Also, have you heard there are some discussion around whether to put TDR's project protected by Google VPC-SC to prevent data exfiltration? If that is in road map, we will need this as well. 
All those API calls seems need the created project has compute API enabled. 
We might be able to make this optional and controlled by resource config. But it may take some effort, and we will have to evaluate this effort first.

tlangs:
As long as its talked about and planned, I think we can just disable the compute API manually for each project we pull from the pool. Keep us updated on the status of the work to make this optional, or if this work could be done at all.

PR #163, opened 2021-06-15
PF-858: add two APIs

jaycarlton:
We need the APIs for GCS and BQ data transfers.

### DIFF
--- a/src/main/resources/config/dev/pool_schema.yml
+++ b/src/main/resources/config/dev/pool_schema.yml
@@ -4,9 +4,9 @@ poolConfigs:
   - poolId: "cwb_ws_dev_v2"
     size: 300
     resourceConfigName: "cwb_ws_resource_dev_v2"
-  - poolId: "workspace_manager_v6"
+  - poolId: "workspace_manager_v7"
     size: 300
-    resourceConfigName: "workspace_manager_v6"
+    resourceConfigName: "workspace_manager_v7"
   - poolId: "vpc_sc_v1"
     size: 300
     resourceConfigName: "vpc_sc_v1"
--- a/src/main/resources/config/dev/pool_schema.yml
+++ b/src/main/resources/config/dev/pool_schema.yml
@@ -4,6 +4,9 @@ poolConfigs:
   - poolId: "cwb_ws_dev_v2"
     size: 300
     resourceConfigName: "cwb_ws_resource_dev_v2"
+  - poolId: "workspace_manager_v6"
+    size: 300
+    resourceConfigName: "workspace_manager_v6"
   - poolId: "workspace_manager_v7"
     size: 300
     resourceConfigName: "workspace_manager_v7"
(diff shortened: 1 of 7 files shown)


yonghaoy (changes requested):
Direct modify pool config is not allowed as backfill and controlling rolling out across development cycle is hard. So you need new pool for this:
https://github.com/DataBiosphere/terra-resource-buffer#upgrade-pool-configuration

src/main/resources/config/dev/pool_schema.yml
```diff
@@ -4,9 +4,9 @@ poolConfigs:
   - poolId: "cwb_ws_dev_v2"
     size: 300
     resourceConfigName: "cwb_ws_resource_dev_v2"
-  - poolId: "workspace_manager_v6"
+  - poolId: "workspace_manager_v7"
```
yonghaoy, line 7: I think now you should add v7 pool and also keep the existing v6. 
If v6 is deleted here, RBS will delete that pool, but WSM is stilling using that pool until you change it in wsm helmfile. 
Same URL: https://github.com/DataBiosphere/terra-resource-buffer#upgrade-pool-configuration
Detailed steps are listed here, you are doing step1 now, please read step 2-5 see when to upgrade pool in wsm and when to delete v6 pool

yonghaoy (changes requested):

src/main/resources/config/dev/pool_schema.yml
jaycarlton: Good catch. I was confused because this file only had one active pool.

yonghaoy: Yes, delete that pool is still required, but that is the last step of the whole process.

yonghaoy (approved):

zloery (approved):
Just to confirm: the new APIs are `bigquerydatatransfer.googleapis.com` and `storagetransfer.googleapis.com`? both of those make sense to me to enable copying cloned data for cloning resources

yonghaoy:
Wait, actually do we want that API be enabled in WSM projects or the user project? 
Rawls workspace projects does not have this permission but still working in the same way. 

https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L578-L584

jaycarlton:
> Wait, actually do we want that API be enabled in WSM projects or the user project?
> Rawls workspace projects does not have this permission but still working in the same way.
> 
> https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L578-L584

I need to enable it in the user's existing project and the destination project for the clone (another user project). I don't know if we need it in the WSM project.

jaycarlton:
After discussing with @wchamber, I'm going to try to implement the clone functionality in CRL rather than bring in these services. Closing this for now. Thanks.

PR #164, opened 2021-06-18
[PF-845]: Groom old Stairway flights in Buffer 

yonghaoy:
Current config: Scan every 6 hour and delete flight older than 90days.

Not sure those are correct.


ddietterich (approved):
LGTM

src/main/resources/application.yml
```diff
@@ -22,12 +22,14 @@ buffer:
 terra:
   common:
     stairway:
```
jmczerk, line 24: `-fligh-` -> `-flight-` ?

jmczerk (changes requested):

src/main/resources/application.yml
yonghaoy: good catch! Thanks!

jmczerk (approved):

PR #167, opened 2021-06-30
PF-538 (part 2): Add workspace_manager_v8 pools for tools, dev environments.

marikomedlock:
Added the `workspace_manager_v8` pool and accompanying resource config to the tools and dev environment. These resource configs are identical to the `_v5` for dev and `_v4` for tools. They are copy/pasted into a new file to make future cleanup easier (see review comments).

The new pools do not delete the default compute engine service account or the default VPC network. This was to support Nextflow, which required both of these things. Now there is a Nextflow edge release that allows overriding both defaults.

I've tested this against my personal environment WSM and updated CLI code. Next I will update the helmfile for all WSM dev/tools deployments to point to these pools.

### DIFF
--- a/src/main/resources/config/dev/pool_schema.yml
+++ b/src/main/resources/config/dev/pool_schema.yml
@@ -7,6 +7,9 @@ poolConfigs:
   - poolId: "workspace_manager_v6"
     size: 300
     resourceConfigName: "workspace_manager_v6"
+  - poolId: "workspace_manager_v7"
+    size: 500
+    resourceConfigName: "workspace_manager_v5"
   - poolId: "vpc_sc_v1"
     size: 300
     resourceConfigName: "vpc_sc_v1"
--- a/src/main/resources/config/dev/pool_schema.yml
+++ b/src/main/resources/config/dev/pool_schema.yml
@@ -7,9 +7,9 @@ poolConfigs:
   - poolId: "workspace_manager_v6"
     size: 300
     resourceConfigName: "workspace_manager_v6"
-  - poolId: "workspace_manager_v7"
+  - poolId: "workspace_manager_v8"
     size: 500
-    resourceConfigName: "workspace_manager_v5"
+    resourceConfigName: "workspace_manager_v8"
   - poolId: "vpc_sc_v1"
     size: 300
     resourceConfigName: "vpc_sc_v1"
--- a/src/main/resources/config/dev/resource-config/workspace_manager_v8.yml
+++ b/src/main/resources/config/dev/resource-config/workspace_manager_v8.yml
@@ -1,0 +1,30 @@
+# Workspace Manager buffered workspace template
+---
+configName: "workspace_manager_v8"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-wsm-dev"
+    scheme: "RANDOM_CHAR"
+  parentFolderId: "421312654454" #test.firecloud.org/dev/buffer-dev/workspace_manager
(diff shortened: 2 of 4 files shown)


yonghaoy (commented):
How about create a v7(just copy from v5) pool instead of using v5?

marikomedlock:
> How about create a v7(just copy from v5) pool instead of using v5?

This does create a new v7 pool, but it reuses the v5 pool config. Do you mean to create a duplicate of the v5 pool config and name it v7?

yonghaoy:
> > How about create a v7(just copy from v5) pool instead of using v5?
> 
> This does create a new v7 pool, but it reuses the v5 pool config. Do you mean to create a duplicate of the v5 pool config and name it v7?

Right. Create a v7 pool config feels less confusing.

marikomedlock:
> > > How about create a v7(just copy from v5) pool instead of using v5?
> > 
> > 
> > This does create a new v7 pool, but it reuses the v5 pool config. Do you mean to create a duplicate of the v5 pool config and name it v7?
> 
> Right. Create a v7 pool config feels less confusing.

What is the purpose of specifying both the pool id and the pool config if they should always match? I think this should be enforced in RBS code instead of having identically named id/config by convention.

yonghaoy:
> > > > How about create a v7(just copy from v5) pool instead of using v5?
> > > 
> > > 
> > > This does create a new v7 pool, but it reuses the v5 pool config. Do you mean to create a duplicate of the v5 pool config and name it v7?
> > 
> > 
> > Right. Create a v7 pool config feels less confusing.
> 
> What is the purpose of specifying both the pool id and the pool config if they should always match? I think this should be enforced in RBS code instead of having identically named id/config by convention.

That is just recommendation, not requirement. RBS does not do the check. As long as the resource config that pool use exists. And we does have test for that: https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/test/java/bio/terra/buffer/config/PoolSchemaTest.java#L21
The reason I recommended the match this is for future easier cleanup.  Eventually resource-config/ may have hundreds of resource configs, and most of them are not used. Would be easier to just delete those files. Feel 1 to 1 mapping makes the kind of cleanup easier. But that is totally up to you, no strong preference.

yonghaoy (approved):

wchamber (approved):

marikomedlock:
> > > > > How about create a v7(just copy from v5) pool instead of using v5?
> > > > 
> > > > 
> > > > This does create a new v7 pool, but it reuses the v5 pool config. Do you mean to create a duplicate of the v5 pool config and name it v7?
> > > 
> > > 
> > > Right. Create a v7 pool config feels less confusing.
> > 
> > 
> > What is the purpose of specifying both the pool id and the pool config if they should always match? I think this should be enforced in RBS code instead of having identically named id/config by convention.
> 
> That is just recommendation, not requirement. RBS does not do the check. As long as the resource config that pool use exists. And we does have test for that: https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/test/java/bio/terra/buffer/config/PoolSchemaTest.java#L21
> The reason I recommended the match this is for future easier cleanup. Eventually resource-config/ may have hundreds of resource configs, and most of them are not used. Would be easier to just delete those files. Feel 1 to 1 mapping makes the kind of cleanup easier. But that is totally up to you, no strong preference.

OK, I'll make the pool id and resource config name identical in dev and tools, so that it conforms with other pools. If we want these to be identical for cleanup purposes, then consider enforcing it in code instead.

yonghaoy (approved):

PR #168, opened 2021-06-30
[CA-1434] VPC pools for environments that don't already have one

andy7i:
Set up VPC pools for envs that don't already have one. 

https://broadworkbench.atlassian.net/browse/CA-1434

Related firecloud-develop PR: https://github.com/broadinstitute/firecloud-develop/pull/2589

### DIFF
--- a/src/main/resources/config/alpha/pool_schema.yml
+++ b/src/main/resources/config/alpha/pool_schema.yml
@@ -4,6 +4,9 @@ poolConfigs:
 - poolId: "cwb_ws_alpha_v1"
   size: 100
   resourceConfigName: "cwb_ws_resource_alpha_v1"
+- poolId: "cwb_ws_alpha_v2"
+  size: 100
+  resourceConfigName: "cwb_ws_resource_alpha_v2"
 - poolId: "cwb_ws_alpha_vpc_v1"
   size: 100
   resourceConfigName: "cwb_ws_resource_alpha_vpc_v1"
--- a/src/main/resources/config/alpha/pool_schema.yml
+++ b/src/main/resources/config/alpha/pool_schema.yml
@@ -4,6 +4,9 @@ poolConfigs:
 - poolId: "cwb_ws_alpha_v1"
   size: 100
   resourceConfigName: "cwb_ws_resource_alpha_v1"
+- poolId: "cwb_ws_alpha_vpc_v1"
+  size: 100
+  resourceConfigName: "cwb_ws_resource_alpha_vpc_v1"
 - poolId: "datarepo_v1"
   size: 300
   resourceConfigName: "datarepo_v1"
--- a/src/main/resources/config/alpha/pool_schema.yml
+++ b/src/main/resources/config/alpha/pool_schema.yml
@@ -7,9 +7,9 @@ poolConfigs:
 - poolId: "cwb_ws_alpha_v2"
   size: 100
   resourceConfigName: "cwb_ws_resource_alpha_v2"
-- poolId: "cwb_ws_alpha_vpc_v1"
+- poolId: "vpc_sc_v1"
   size: 100
-  resourceConfigName: "cwb_ws_resource_alpha_vpc_v1"
+  resourceConfigName: "vpc_sc_v1"
 - poolId: "datarepo_v1"
   size: 300
   resourceConfigName: "datarepo_v1"
--- a/src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v1.yml
+++ b/src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v1.yml
@@ -3,7 +3,7 @@
 configName: "cwb_ws_resource_alpha_v1"
 gcpProjectConfig:
   projectIdSchema:
-    prefix: "terra-alpha"
+    prefix: "terra-dev"
     scheme: "RANDOM_CHAR"
   # test.firecloud.org/alpha/CommunityWorkbench
   parentFolderId: "102618195587"
--- a/src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v1.yml
+++ b/src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v1.yml
@@ -3,7 +3,7 @@
 configName: "cwb_ws_resource_alpha_v1"
 gcpProjectConfig:
   projectIdSchema:
-    prefix: "terra-dev"
+    prefix: "terra-alpha"
     scheme: "RANDOM_CHAR"
   # test.firecloud.org/alpha/CommunityWorkbench
   parentFolderId: "102618195587"
--- a/src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v2.yml
+++ b/src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v2.yml
@@ -1,0 +1,30 @@
+# Community Workbench buffered workspace template
+---
+configName: "cwb_ws_resource_alpha_v2"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-alpha"
+    scheme: "RANDOM_CHAR"
+  # test.firecloud.org/alpha/CommunityWorkbench
+  parentFolderId: "102618195587"
+  billingAccount: "01A82E-CA8A14-367457"
+  enabledApis:
+  - "bigquery-json.googleapis.com"
+  - "compute.googleapis.com"
+  - "container.googleapis.com"
+  - "containerregistry.googleapis.com"
+  - "cloudbilling.googleapis.com"
+  - "clouderrorreporting.googleapis.com"
+  - "cloudkms.googleapis.com"
+  - "cloudtrace.googleapis.com"
+  - "dataflow.googleapis.com"
+  - "dataproc.googleapis.com"
+  - "dns.googleapis.com"
+  - "genomics.googleapis.com"
+  - "lifesciences.googleapis.com"
+  - "logging.googleapis.com"
+  - "monitoring.googleapis.com"
+  - "storage-api.googleapis.com"
+  - "storage-component.googleapis.com"
+  network:
+    enableNetworkMonitoring: "false"
--- a/src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_vpc_v1.yml
+++ b/src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_vpc_v1.yml
@@ -1,12 +1,12 @@
 # Community Workbench buffered workspace template
 ---
-configName: "cwb_ws_resource_staging_v1"
+configName: "cwb_ws_resource_alpha_vpc_v1"
 gcpProjectConfig:
   projectIdSchema:
-    prefix: "terra-staging"
+    prefix: "terra-alpha-vpc"
     scheme: "RANDOM_CHAR"
-  # test.firecloud.org/staging/CommunityWorkbench
-  parentFolderId: "137690849844"
+  # test.firecloud.org/alpha/CommunityWorkbench
+  parentFolderId: "102618195587"
   billingAccount: "01A82E-CA8A14-367457"
   enabledApis:
   - "bigquery-json.googleapis.com"
@@ -27,4 +27,4 @@ gcpProjectConfig:
   - "storage-api.googleapis.com"
   - "storage-component.googleapis.com"
   network:
-    enableNetworkMonitoring: "false"
+    enableNetworkMonitoring: "true"
--- a/src/main/resources/config/alpha/resource-config/vpc_sc_v1.yml
+++ b/src/main/resources/config/alpha/resource-config/vpc_sc_v1.yml
@@ -6,7 +6,7 @@ gcpProjectConfig:
     prefix: "terra-vpc-sc-alpha"
     scheme: "RANDOM_CHAR"
   # test.firecloud.org/alpha/for_vpc_sc_unclaimed
-  parentFolderId: "TODO"
+  parentFolderId: "38455243798"
   billingAccount: "01A82E-CA8A14-367457"
   enabledApis:
   - "bigquery-json.googleapis.com"
--- a/src/main/resources/config/alpha/resource-config/vpc_sc_v1.yml
+++ b/src/main/resources/config/alpha/resource-config/vpc_sc_v1.yml
@@ -1,12 +1,12 @@
 # Community Workbench buffered workspace template
 ---
-configName: "cwb_ws_resource_alpha_vpc_v1"
+configName: "vpc_sc_v1"
 gcpProjectConfig:
   projectIdSchema:
-    prefix: "terra-alpha-vpc"
+    prefix: "terra-vpc-sc-alpha"
     scheme: "RANDOM_CHAR"
-  # test.firecloud.org/alpha/CommunityWorkbench
-  parentFolderId: "102618195587"
+  # test.firecloud.org/alpha/for_vpc_sc_unclaimed
+  parentFolderId: "TODO"
   billingAccount: "01A82E-CA8A14-367457"
   enabledApis:
   - "bigquery-json.googleapis.com"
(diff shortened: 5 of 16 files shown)


src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v1.yml
```diff
@@ -3,7 +3,7 @@
 configName: "cwb_ws_resource_alpha_v1"
 gcpProjectConfig:
   projectIdSchema:
-    prefix: "terra-dev"
+    prefix: "terra-alpha"
```
andy7i, line 6: does this require a v2 file/pool?

src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_vpc_v1.yml
```diff
@@ -0,0 +1,30 @@
+# Community Workbench buffered workspace template
+---
+configName: "cwb_ws_resource_alpha_vpc_v1"
```
andy7i, line 3: these yml's were copied from the non-vpc file. for example, this file ("cwb_ws_resource_alpha_vpc_v1.yml" was copied from "cwb_ws_resource_alpha_v1.yml" so they should be equivalent except for the configName and enableNetworkMonitoring (and maybe the "prefix").

src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v1.yml
yonghaoy: yes

src/main/resources/config/staging/resource-config/cwb_ws_resource_staging_v1.yml
```diff
@@ -3,7 +3,7 @@
 configName: "cwb_ws_resource_staging_v1"
 gcpProjectConfig:
   projectIdSchema:
-    prefix: "terra-alpha"
+    prefix: "terra-staging"
```
yonghaoy, line 6: ditto, you need v2 here as well

yonghaoy (changes requested):
VPC-SC GCP projects is only used by AoU at this moment.  And AoU only use Terra dev and prod. So I did not create vpc-sc pool in other envs. 
Is the requirement changing?

yonghaoy (changes requested):
See https://github.com/DataBiosphere/terra-resource-buffer/blob/d19530b63d2912386be6de85a1498c19f4898447/src/main/resources/config/dev/resource-config/vpc_sc_v1.yml as example. 
There is the pattern we use as discussed with in slack before:
1: config name: configName: "vpc_sc_v1"
2: You need to create those new folders in each env.  Example: test.firecloud.org/dev/for_vpc_sc_unclaimed
3: Prefix: terra-vpc-sc-{env}
4: pool name is also vpc_sc_v1
5: If no clients want to use those projects, I suggest not to create that large pool(20~100) is a reasonable pool size.

src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_vpc_v1.yml
yonghaoy: see mty another comment. The current pattern we are using. Config name can be `vpc_sc_v1`

src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_vpc_v1.yml
```diff
@@ -0,0 +1,30 @@
...
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-alpha-vpc"
+    scheme: "RANDOM_CHAR"
+  # test.firecloud.org/alpha/CommunityWorkbench
```
yonghaoy, line 8: You should use another folder.

src/main/resources/config/tools/pool_schema.yml
```diff
@@ -4,9 +4,15 @@ poolConfigs:
...
     size: 500
     resourceConfigName: "cwb_ws_resource_fiab_v2"
+  - poolId: "cwb_ws_dev-fiab_vpc_v2"
+    size: 500
+    resourceConfigName: "cwb_ws_resource_dev-fiab_vpc_v2"
   - poolId: "cwb_ws_quality_v2"
     size: 500
     resourceConfigName: "cwb_ws_resource_quality_v2"
+  - poolId: "cwb_ws_qa-fiab_vpc_v2"
+    size: 500
```
yonghaoy, line 14: 500 is too large for a pool that no clients

andy7i:
@yonghaoy  thanks for the feedback!


> VPC-SC GCP projects is only used by AoU at this moment. And AoU only use Terra dev and prod. So I did not create vpc-sc pool in other envs.
> Is the requirement changing?

VPC-SC gcp projects can be used outside of AoU so this is to help set that up. I'm not 100% sure if we will need projects with network monitoring (and flow logs) outside of service perimeters, in which case maybe it makes sense to create a new pool for non-serviceperimeter projects?

Question for you:
why do perf and prod have `"notebooks.googleapis.com"`? (The other envs do not)

yonghaoy:
> @yonghaoy thanks for the feedback!
> 
> > VPC-SC GCP projects is only used by AoU at this moment. And AoU only use Terra dev and prod. So I did not create vpc-sc pool in other envs.
> > Is the requirement changing?
> 
> VPC-SC gcp projects can be used outside of AoU so we need to set up a pool for this scenario. We may also need projects with network monitoring (and flow logs) that will not use service perimeters, in which case maybe it makes sense to create a new pool for non-serviceperimeter projects since the current pool lives in `for_vpc_sc_unclaimed` folders? (I think part of this answer will come from https://broadworkbench.atlassian.net/browse/CA-1429)
> 
> Question for you:
> why do perf and prod have `"notebooks.googleapis.com"` for vpc and non-vpc pools? (The other envs do not have it. Should they? Looks like WSM also has it.)
> 
> Screen Shot 2021-07-01 at 5 38 11 PM

RE: Notebook API is optional for Rawls but required for WSM(WSM supports creates Cloud API Notebook, but I don't know if Rawsl needs). I may copied/paste from WSM template when setting up Rawls Prod. I think you can leave as it or add that API for other envs in case Rawls need to support Terra-cli(https://github.com/DataBiosphere/terra-cli/tree/main/notebooks). Sorry for the confusion.

src/main/resources/config/staging/resource-config/vpc_sc_v1.yml
```diff
@@ -0,0 +1,30 @@
+# Community Workbench buffered workspace template
+---
+configName: "vpc_sc_v1"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-vpc-sc-staging"
```
yonghaoy, line 6: FYI: GCP has name prefix of 30 character. This is already 21(including the extra - appended). 
But it should still work here because we hard coded the random char after it is 8
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java#L20

yonghaoy (approved):

andy7i:
> > @yonghaoy thanks for the feedback!
> > > VPC-SC GCP projects is only used by AoU at this moment. And AoU only use Terra dev and prod. So I did not create vpc-sc pool in other envs.
> > > Is the requirement changing?
> > 
> > 
> > VPC-SC gcp projects can be used outside of AoU so we need to set up a pool for this scenario. We may also need projects with network monitoring (and flow logs) that will not use service perimeters, in which case maybe it makes sense to create a new pool for non-serviceperimeter projects since the current pool lives in `for_vpc_sc_unclaimed` folders? (I think part of this answer will come from https://broadworkbench.atlassian.net/browse/CA-1429)
> > Question for you:
> > why do perf and prod have `"notebooks.googleapis.com"` for vpc and non-vpc pools? (The other envs do not have it. Should they? Looks like WSM also has it.)
> > Screen Shot 2021-07-01 at 5 38 11 PM
> 
> RE: Notebook API is optional for Rawls but required for WSM(WSM supports creates Cloud API Notebook, but I don't know if Rawsl needs). I may copied/paste from WSM template when setting up Rawls Prod. I think you can leave as it or add that API for other envs in case Rawls need to support Terra-cli(https://github.com/DataBiosphere/terra-cli/tree/main/notebooks). Sorry for the confusion.

Ideally we wouldn't have any extraneous permissions. I think we should check with @broad-dsp-appsec to see if it's ok for the terra/rawls projects in the prod and perf environments to have the extra api enabled.

src/main/resources/config/alpha/pool_schema.yml
```diff
@@ -4,6 +4,12 @@ poolConfigs:
 - poolId: "cwb_ws_alpha_v1"
   size: 100
   resourceConfigName: "cwb_ws_resource_alpha_v1"
+- poolId: "cwb_ws_alpha_v2"
+  size: 100
+  resourceConfigName: "cwb_ws_resource_alpha_v2"
+- poolId: "vpc_sc_v1"
```
jacmrob, line 10: what does sc stand for?

src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v2.yml
```diff
@@ -0,0 +1,30 @@
+# Community Workbench buffered workspace template
+---
+configName: "cwb_ws_resource_alpha_v2"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-alpha"
+    scheme: "RANDOM_CHAR"
+  # test.firecloud.org/alpha/CommunityWorkbench
+  parentFolderId: "102618195587"
+  billingAccount: "01A82E-CA8A14-367457"
```
jacmrob, line 10: is it okay that this is hard-coded?

src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v2.yml
```diff
@@ -0,0 +1,30 @@
+# Community Workbench buffered workspace template
+---
+configName: "cwb_ws_resource_alpha_v2"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-alpha"
```
jacmrob, line 6: so do these get created in their own google projects then rather than the projects that correlate with their environments?

src/main/resources/config/alpha/pool_schema.yml
jacmrob: and what are the two pools (`cwb_ws...` and `vpc_sc...`) in each environment for?

src/main/resources/config/alpha/resource-config/vpc_sc_v1.yml
```diff
@@ -0,0 +1,30 @@
+# Community Workbench buffered workspace template
+---
+configName: "vpc_sc_v1"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-vpc-sc-alpha"
```
jacmrob, line 6: curious on choice of naming schema... if `vpc_sc_v1` -> `terra-vpc-sc-alpha`, should `cwb_ws_resource_alpha_v2` -> `terra-cwb-ws-resource-alpha`? 

Not asking that it be changed if this is consistently the naming convention in the repo. maybe understanding the difference between these two pools will help too.

jacmrob (commented):
I'm assuming `prod` and `dev` already have have VPC pools?

marctalbott (approved):
Looks good. Plus one to keeping the enabled apis across environments standardized. The more we can keep our projects homogenous the better.

yonghaoy:
> > > @yonghaoy thanks for the feedback!
> > > > VPC-SC GCP projects is only used by AoU at this moment. And AoU only use Terra dev and prod. So I did not create vpc-sc pool in other envs.
> > > > Is the requirement changing?
> > > 
> > > 
> > > VPC-SC gcp projects can be used outside of AoU so we need to set up a pool for this scenario. We may also need projects with network monitoring (and flow logs) that will not use service perimeters, in which case maybe it makes sense to create a new pool for non-serviceperimeter projects since the current pool lives in `for_vpc_sc_unclaimed` folders? (I think part of this answer will come from https://broadworkbench.atlassian.net/browse/CA-1429)
> > > Question for you:
> > > why do perf and prod have `"notebooks.googleapis.com"` for vpc and non-vpc pools? (The other envs do not have it. Should they? Looks like WSM also has it.)
> > > Screen Shot 2021-07-01 at 5 38 11 PM
> > 
> > 
> > RE: Notebook API is optional for Rawls but required for WSM(WSM supports creates Cloud API Notebook, but I don't know if Rawsl needs). I may copied/paste from WSM template when setting up Rawls Prod. I think you can leave as it or add that API for other envs in case Rawls need to support Terra-cli(https://github.com/DataBiosphere/terra-cli/tree/main/notebooks). Sorry for the confusion.
> 
> Ideally we wouldn't have any extraneous permissions. I think we should check with @broad-dsp-appsec to see if it's ok for the terra/rawls projects in the prod and perf environments to have the extra api (which is already enabled for WSM projects) enabled.

Or we can create a new pool without the extra permissions in prod and perf if we want

src/main/resources/config/alpha/pool_schema.yml
andy7i: the SC stands for Service Control which is used for controlling access and data exfiltration https://cloud.google.com/vpc-service-controls

I'm following the existing naming schemes -- `cwb_ws` stands for community workbench workspace, which is the "regular" project type in terra. The vpc_sc projects are used for projects that live inside a service perimeter.

src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v2.yml
andy7i: they get created into the org/folder that is specified by `parentFolderId` on lines 8-9 below. In this case, it is created in `test.firecloud.org/alpha/CommunityWorkbench` (dev, alpha, perf, staging all live in the test.firecloud.org org)

src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v2.yml
andy7i: I do think ideally we would be able to get the billingAccount and enabledApis (below) from a config file so that we don't have to copy/paste these into each yml, but I don't think we can do that yet.

src/main/resources/config/alpha/resource-config/vpc_sc_v1.yml
andy7i: I'm mainly following the existing naming convention here. I think the vpc_sc pools will be used primarily for AoU, so maybe that's why?

yonghaoy: I am not sure what would be the best name prefix. That name came from: https://github.com/DataBiosphere/terra-resource-buffer/pull/145#discussion_r614068570

andy7i:
I'll handle the perf and prod changes as part of a separate PR to keep things more contained.

andy7i:
> I'm assuming `prod` and `dev` already have have VPC pools?

yep!

src/main/resources/config/alpha/pool_schema.yml
jacmrob: Gotcha, thanks

src/main/resources/config/alpha/resource-config/cwb_ws_resource_alpha_v2.yml
jacmrob: Okay, no problem

jacmrob (approved):
Lgtm

andy7i:
@yonghaoy github tells me `Merging is blocked The base branch restricts merging to authorized users. Learn more about protected branches.` Should we have access to merge to master? / could you merge it?

yonghaoy:
> @yonghaoy github tells me `Merging is blocked The base branch restricts merging to authorized users. Learn more about protected branches.` Should we have access to merge to master? / could you merge it?

Hi @andy7i The repo access was messed up when I setup 2 reviewer policy with GHA auto push. Now it should be fixed. Can you try again?

andy7i:
@yonghaoy works for me now -- thanks!

PR #169, opened 2021-06-30
Expand subnets BT-326

kshakir:


yonghaoy (commented):
Do we need to migrate all existing projects? 
For Buffer service: It preallocated GCP projects. If you make change here, there would be /20 and /16 projects coexists in the pool and waiting on Raws/AoU pick it up...

kshakir:
@yonghaoy

> Do we need to migrate all existing projects?

Yes. This should be a "create or update subnets" per project

> For Buffer service: It preallocated GCP projects. If you make change here, there would be /20 and /16 projects coexists in the pool and waiting on Raws/AoU pick it up...

I'm a bit new to RBS beyond what I've heard in DSP Workbench sprint reviews. Can you point me to an example of RBS performing resource updates & deletes for existing projects?

For example, there is already code that was creating subnets across an older list of regions [here](https://github.com/broadinstitute/gcp-dm-templates/pull/28/files). I'd like to try to make sure whatever RBS modifier runs ensures that all projects have the same list of GCP regions with subnets.

yonghaoy:
> @yonghaoy
> 
> > Do we need to migrate all existing projects?
> 
> Yes. This should be a "create or update subnets" per project
> 
> > For Buffer service: It preallocated GCP projects. If you make change here, there would be /20 and /16 projects coexists in the pool and waiting on Raws/AoU pick it up...
> 
> I'm a bit new to RBS beyond what I've heard in DSP Workbench sprint reviews. Can you point me to an example of RBS performing resource updates & deletes for existing projects?
> 
> For example, there is already code that was creating subnets across an older list of regions [here](https://github.com/broadinstitute/gcp-dm-templates/pull/28/files). I'd like to try to make sure whatever RBS modifier runs ensures that all projects have the same list of GCP regions with subnets.

1: The DM code you pointed will be replaced by RBS soon(along with PPW). After PPW, we will need a GCP project for each workspace creation, and creating a GCP project takes 5~10 minutes. We can not use DM anymore, that is why we have RBS. You can imagine RBS is doing the same thing as DM, just pre-allocate them.
2: RBS just pre-allocates resources. It does not update/delete resources. So if you make change here, once code is released, only projects created after that time will have the new change, all existing projects(including the ones handed and used by clients, and the ones still existing in resource pool waiting to be handed out) is not updated.

mcovarr (approved):

evanbernstein:
The problem we were trying to solve with CloudNAT was that we wanted to make it easy for users to be able to run more jobs in parallel, and we knew that GCP had strict limits on IP quotas. Last year when we talked to GCP, they highly recommended CloudNAT as a solution around IP limits. Today, they aren't pushing it as strongly, and it feels like they are more interested in moving towards IPv6 as the solution. They have also said that we can escalate any IP quotas and get them all approved quickly, which solves the original problem (in addition, we will also have to remove a limit within Cromwell, but we were going to do that anyway). 

The biggest problem with CloudNAT is that there are ingress and egress costs to any data that is not accessed via a GCP API. The way things currently work, this could be fine because we resolve DRS URLs to GCS links. However, the future of DRS is Signed URLs. Those would not use the GCP API (because they are simply https links). Every use of a DRS Signed URL would incur costs. These CloudNAT costs are not simple to explain. Given that we were trying to use CloudNAT to resolve a limit for all clients, and it turned from something we thought was simple, cheaper, and basically free, to something that could have extreme costs that are hard to explain, it no longer is the right solution for our problem. We do think that CloudNAT will have uses for specific clients, but that is not the goal we were currently going after, so this work is now deferred.

PR #170, opened 2021-07-01
PF-538 (part 3): Remove workspace_manager_v6, v7 pools in tools and dev.

marikomedlock:
Removed the `workspace_manager_v6` pool in dev and the `v6` and `v7` pools in tools. Both are superseded by the new `v8` pool in dev and tools.

I searched the terra-helmfile repo and found no use of the `v6` and `v7` pools.


src/main/resources/config/dev/pool_schema.yml
```diff
@@ -4,9 +4,6 @@ poolConfigs:
   - poolId: "cwb_ws_dev_v2"
     size: 300
     resourceConfigName: "cwb_ws_resource_dev_v2"
-  - poolId: "workspace_manager_v6"
-    size: 300
-    resourceConfigName: "workspace_manager_v6"
```
marikomedlock, line 9: @yonghaoy Should I also delete the associated resource config file if it's not used in any other pool schema? I see there are other unused resource config files, so not sure what the best practice is.

yonghaoy: There is no best practice around that as that is a new service with only few activate clients. 
But it's totally fine to delete it.All pool and resource configs are also stored in DB, we will not lose anything. 
just to make sure that v6 is not used by WSM by the time you delete that.

yonghaoy (approved):

wchamber (approved):

PR #172, opened 2021-07-08
[ca-1434] created new pools for perf/prod in order to remove unused notebook API

andy7i:
In order to keep projects consistent across envs, we discussed creating new pools: #168 (comment) This PR handles that extra change.

related PR: https://github.com/broadinstitute/firecloud-develop/pull/2596

### DIFF
--- a/src/main/resources/config/perf/pool_schema.yml
+++ b/src/main/resources/config/perf/pool_schema.yml
@@ -14,13 +14,13 @@ poolConfigs:
     size: 500
     resourceConfigName: "cwb_ws_perf_v1"
   - poolId: "cwb_ws_perf_v2"
-    size: 1000
+    size: 500
     resourceConfigName: "cwb_ws_perf_v2"
   - poolId: "vpc_sc_v1"
     size: 500
     resourceConfigName: "vpc_sc_v1"
   - poolId: "vpc_sc_v2"
-    size: 1000
+    size: 500
     resourceConfigName: "vpc_sc_v2"
   - poolId: "datarepo_v1"
     size: 300
--- a/src/main/resources/config/perf/pool_schema.yml
+++ b/src/main/resources/config/perf/pool_schema.yml
@@ -13,9 +13,15 @@ poolConfigs:
   - poolId: "cwb_ws_perf_v1"
     size: 500
     resourceConfigName: "cwb_ws_perf_v1"
+  - poolId: "cwb_ws_perf_v2"
+    size: 1000
+    resourceConfigName: "cwb_ws_perf_v2"
   - poolId: "vpc_sc_v1"
     size: 500
     resourceConfigName: "vpc_sc_v1"
+  - poolId: "vpc_sc_v2"
+    size: 1000
+    resourceConfigName: "vpc_sc_v2"
   - poolId: "datarepo_v1"
     size: 300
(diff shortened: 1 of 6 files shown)


src/main/resources/config/perf/pool_schema.yml
```diff
@@ -13,9 +13,15 @@ poolConfigs:
...
     size: 500
     resourceConfigName: "cwb_ws_perf_v1"
+  - poolId: "cwb_ws_perf_v2"
+    size: 1000
+    resourceConfigName: "cwb_ws_perf_v2"
   - poolId: "vpc_sc_v1"
     size: 500
     resourceConfigName: "vpc_sc_v1"
+  - poolId: "vpc_sc_v2"
+    size: 1000
```
andy7i, line 23: @aweng98 do you have thoughts on what size the project buffer should be for Terra/AoU perf testing?

src/main/resources/config/perf/resource-config/cwb_ws_perf_v2.yml
```diff
@@ -0,0 +1,30 @@
+# Community Workbench buffered workspace template
+---
+configName: "cwb_ws_perf_v2"
```
andy7i, line 3: these ymls are equivalent to the v1 versions except for: updated version #, removed unused `notebooks.googleapis.com` API.

src/main/resources/config/perf/pool_schema.yml
aweng98: AoU perf tests use up ~250 projects per test run. If 1000 projects is for AoU perf tests only, then it is more than enough. No change needed. If it's for all perf tests, make it a 50% more. On some days, perf tests are needed to rerun 1+ times. Thanks.

andy7i: Thanks for the input!

This pool will be used for workspaces in a service perimeter, which I assume is mostly just used for AoU. 

There's another pool of 1000 for regular workspaces. Do you know if 1000 projects should be enough for the nightly perf tests?

yonghaoy: +1 1000 sounds too much for AoU, and creating 2000 at one time for RBS is also a heavy job...
Also too much for CWB.  CWB only need < 200 concurrent runs per: 
https://github.com/DataBiosphere/terra-resource-buffer/pull/130#discussion_r597927657

yonghaoy (approved):

src/main/resources/config/perf/pool_schema.yml
andy7i: thanks yonghao -- I'll bump it back to 500 then?

yonghaoy: Also, large pool size is good for large QPS or large concurrent request. Since perf with nightly runs, no need to have too large pool, RBS will fill the pool up pretty quick

aweng98: More than enough.

andy7i: sounds good -- thanks all

aweng98 (approved):

jacmrob (approved):
LGTM

marctalbott (approved):

PR #179, opened 2021-07-30
[no ticket]: Increase how long to keep the resource before asking Janitor to delete it.  

yonghaoy:
Most of perf tests(AoU, Buffer, Terra) happens at midnight - 4am. We should avoid keeping Janitor busy during that time. Change to 10 hrs seems can make janitor avoid that busy time slot.

As-is, we've seen the janitor 429 trying to delete projects with the following error message.
`Quota exceeded for quota metric 'Write requests' and limit 'Write requests per minute' of service 'cloudresourcemanager.googleapis.com' for consumer 'project_number:191683885592'`

### DIFF
--- a/src/main/java/bio/terra/buffer/app/configuration/CrlConfiguration.java
+++ b/src/main/java/bio/terra/buffer/app/configuration/CrlConfiguration.java
@@ -45,7 +45,7 @@ public class CrlConfiguration {
   /** The client name required by CRL. */
   public static final String CLIENT_NAME = "terra-resource-buffer";
   /** How long to keep the resource before Janitor do the cleanup. */
-  public static final Duration TEST_RESOURCE_TIME_TO_LIVE = Duration.ofMinutes(60);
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/buffer/app/configuration/CrlConfiguration.java
```diff
@@ -45,7 +45,7 @@
   /** The client name required by CRL. */
   public static final String CLIENT_NAME = "terra-resource-buffer";
   /** How long to keep the resource before Janitor do the cleanup. */
-  public static final Duration TEST_RESOURCE_TIME_TO_LIVE = Duration.ofMinutes(60);
+  public static final Duration TEST_RESOURCE_TIME_TO_LIVE = Duration.ofHours(10);
```
wchamber, line 48: let's add the same explanation as you have in the PR description to the comment here.

wchamber (approved):

zloery (approved):

PR #180, opened 2021-08-02
PF-228: Validate generated project id. Restrict prefix length.

marikomedlock:
- Added length validation to the GCP project id generation. Check that the length is <= 30 ([ref](https://cloud.google.com/resource-manager/docs/creating-managing-projects#:~:text=A%20project%20ID%20must%20start,between%206%20and%2030%20characters.&text=You%20can't%20use%20certain,new%20project%20with%20the%20projects.)).
- Moved the check for whether the project id is already in use from the `CreateProjectStep` to the `GenerateProjectIdStep`, and try to generate another id if the generated one is already in use.
- Restricted the project id prefix length to 12 characters. This only applies when using the `TWO_WORDS_NUMBER` naming scheme. I chose 12 because on average we get a valid project id on the second try. Some prefixes will have to be shortened to meet this requirement (e.g. `terra-wsm-test` -> `terra-wsm-t`) Below are some statistics for the number of tries to get a valid project id with the given prefix length:

| Prefix Length | Mean | Max | 95 Percentile | 99 Percentile |
| ----------- | ----------- | ----------- | ----------- | ----------- |
| 11 | 1 | 15 | 4 | 6 |
| 12 | 2 | 28 | 7 | 11 |
| 13 | 5 | 65 | 16 | 26 |
| 14 | 15 | 164 | 45 | 70 |
| 15 | 76 | 853 | 228 | 354 |
| 16 | 785 | 7507 | 2351 | 3629 |

- Added a new tools project pool for WSM personal environments that uses the `TWO_WORDS_NUMBER` naming scheme. `workspace_manager_v9` uses the prefix `terra-wsm-t` instead of the previous `terra-wsm-test` so that it meets the prefix length requirement.

### DIFF
--- a/src/main/java/bio/terra/buffer/service/resource/flight/GenerateProjectIdStep.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/GenerateProjectIdStep.java
@@ -41,7 +41,7 @@ public class GenerateProjectIdStep implements Step {
           CLOUD_RESOURCE_UID,
           new CloudResourceUid().googleProjectUid(new GoogleProjectUid().projectId(projectId)));
       return StepResult.getStepResultSuccess();
-    } catch (IOException ioEx) {
+    } catch (IOException | InterruptedException ioEx) {
       logger.info("Error when generating GCP project id.", ioEx);
       return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, ioEx);
     }
--- a/src/main/java/bio/terra/buffer/service/resource/flight/GenerateProjectIdStep.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/GenerateProjectIdStep.java
@@ -2,6 +2,7 @@ package bio.terra.buffer.service.resource.flight;
 
 import static bio.terra.buffer.service.resource.FlightMapKeys.CLOUD_RESOURCE_UID;
 import static bio.terra.buffer.service.resource.FlightMapKeys.GOOGLE_PROJECT_ID;
+import static bio.terra.buffer.service.resource.flight.GoogleUtils.retrieveProject;
 
 import bio.terra.buffer.generated.model.CloudResourceUid;
 import bio.terra.buffer.generated.model.GcpProjectConfig;
@@ -35,15 +36,33 @@ public class GenerateProjectIdStep implements Step {
     FlightMap workingMap = flightContext.getWorkingMap();
     try {
       String projectId =
-          projectIdGenerator.generateIdWithRetries(gcpProjectConfig.getProjectIdSchema(), rmCow);
+          projectIdGenerator.generateIdWithRetries(gcpProjectConfig.getProjectIdSchema());
+
+      // check that a project with this id does not already exist. to do this check, we try
+      // to retrieve the project. this will only succeed if RBS has permission to get that
+      // project. we expect this to be the case most of the time because of the Terra-specific
+      // prefix, but it's possible this step will generate a project that is in use outside of
+      // Terra. in that case, the project creation step will fail.
+      if (retrieveProject(rmCow, projectId).isPresent()) {
+        logger.info("Generated GCP project is already in use: {}", projectId);
+
+        // here we use Stairway retry, instead of retrying immediately in order to rate limit
+        // the number of project get calls made per step execution. i.e. we could retry the
+        // project id generation immediately since we haven't hit a quota issue yet, but that
+        // would mean each execution of this step may make a variable number of project get
+        // calls. that would make it hard to decide how many flights could run in parallel
+        // without hitting project get quota issues.
+        return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY);
+      }
+
       flightContext.getWorkingMap().put(GOOGLE_PROJECT_ID, projectId);
       workingMap.put(
           CLOUD_RESOURCE_UID,
           new CloudResourceUid().googleProjectUid(new GoogleProjectUid().projectId(projectId)));
       return StepResult.getStepResultSuccess();
-    } catch (IOException | InterruptedException ioEx) {
-      logger.info("Error when generating GCP project id.", ioEx);
-      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, ioEx);
+    } catch (IOException | InterruptedException ex) {
+      logger.info("Error when generating GCP project id.", ex);
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, ex);
     }
   }
 
--- a/src/main/java/bio/terra/buffer/service/resource/flight/GenerateProjectIdStep.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/GenerateProjectIdStep.java
@@ -9,9 +9,14 @@ import bio.terra.buffer.generated.model.GoogleProjectUid;
 import bio.terra.buffer.service.resource.projectid.GcpProjectIdGenerator;
 import bio.terra.cloudres.google.cloudresourcemanager.CloudResourceManagerCow;
 import bio.terra.stairway.*;
+import java.io.IOException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Generates Project Id and put it in working map. */
 public class GenerateProjectIdStep implements Step {
+  private final Logger logger = LoggerFactory.getLogger(GenerateProjectIdStep.class);
+
   private final GcpProjectConfig gcpProjectConfig;
   private final GcpProjectIdGenerator projectIdGenerator;
   private final CloudResourceManagerCow rmCow;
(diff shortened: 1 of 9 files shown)


src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java
```diff
@@ -22,10 +30,58 @@
...
+   * @param rmCow resource manager wrapper to retrieve the project
+   * @return generated project id (prefix + naming scheme)
+   */
+  public String generateIdWithRetries(
+      ProjectIdSchema projectIdSchema, CloudResourceManagerCow rmCow) {
+    for (int numTries = 0; numTries < MAX_RETRIES; numTries++) {
+      String projectId = generateId(projectIdSchema);
+      if (projectId.length() <= MAX_LENGTH_GCP_PROJECT_ID)
+        try {
+          if (retrieveProject(rmCow, projectId).isEmpty()) {
```
yonghaoy, line 66: this pattern may not work well in limited GCP API quota environment. `retrieveProject` is a google call, and we expect GCP calls' retry interval is slightly longer because those quota is x numbers/**minute** 
That's why we have [newCloudApiDefaultRetryRule](https://github.com/DataBiosphere/terra-resource-buffer/blob/af480aeda9f0dd336065536daa6241b78b7123ec/src/main/java/bio/terra/buffer/service/resource/flight/StepUtils.java#L20) and [newInternalDefaultRetryRule](https://github.com/DataBiosphere/terra-resource-buffer/blob/af480aeda9f0dd336065536daa6241b78b7123ec/src/main/java/bio/terra/buffer/service/resource/flight/StepUtils.java#L28)

Suggest to split generateId and retrieveProject like what we are doing now, 

this method here just retry this method with no sleep:
```
 if (projectId.length() <= MAX_LENGTH_GCP_PROJECT_ID) {
   return projectId;
} 
```

In Flight step: 
```
try {
generateStep()
retrieveProject()
} catch (IOE | RuntimeE) {
return step retryable error;
}
```

One more note is in the flight step which calls this method, shall we catch exception that throwed by this method and return retry-able failure?  If just throw RuntimeException, Stairway may just fail that flight with no retries: 
https://github.com/DataBiosphere/stairway/blob/develop/stairway/src/main/java/bio/terra/stairway/Flight.java#L301

src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java
```diff
@@ -22,10 +30,58 @@
   /** The largest number to use for random suffixes for adjective/noun project IDs. */
   @VisibleForTesting static final int RANDOM_SUFFIX_LIMIT = 10000;
 
+  /** The maximum allowed length for a GCP project id. */
+  private static final int MAX_LENGTH_GCP_PROJECT_ID = 30;
+
+  /** The maximum allowed length for a GCP project id prefix (e.g. "terra-"). */
+  public static final int MAX_LENGTH_GCP_PROJECT_ID_PREFIX = 12;
```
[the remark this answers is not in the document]
marikomedlock: No, because this length restriction only applies to the `TWO_WORDS_NUMBER` naming scheme. I'll update the comment to note this.

src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java
marikomedlock: Good point about potential quota issues. Two things I'm not sure about:
- In this case, we're not retrying the `retrieveProject` call, we're retrying the project id generation. So I think it only makes sense to use Stairway retries for sleeping after an `IOException`. If we find an existing project with this name, then we can just try to generate a new id immediately.
- I don't think we need to use Stairway retries for calls to `generateId`. This just does a random word lookup, so sleeping won't help anything. I think we can just fail the flight at that point.

yonghaoy: 1: What you mean by "we're not retrying the retrieveProject call, we're retrying the project id generation." I think you do retry `retrieveProject` call when the previous project id is already being used. Also the quota issues happens when we submit large number of flights at the same time. For example, creating a new pool, and that require create large number of CreationFlight at one time. There will be many retrieveProject call fired by other same steps.
2: In your code, your throw RuntimeException. That one is not captuered by Stairway. Would Stairway just treat this as a failure flight with no retry? I think so. But I don't think it is a 'retry-able' Stairway step.

yonghaoy: The suggestion above is not to make all retries in Stairway retried. it make sense to retry immediately if project id too long. 
Then the  `GenerateProjectIdStep` get the short project id back from the static method, it use that project id to make a retrieve call, that is part we Stairway retry can be used.

marikomedlock: 1: If `retrieveProject` finds that project id 123 is in use, then we generate a new project id 456 and call `retrieveProject` on that. It's not the same call that's being retried over and over. That's why I'm not sure sleeping between calls makes sense. e.g. if `retrieveProject(123)` finds a project, then sleeping 10 minutes before a call to `retrieveProject(456)` shouldn't affect the outcome.  That's why I didn't include a sleep (via stairway retries) between calls to `retreiveProject`.

However, if `retrieveProject` fails with an `IOException` then that may be a quota issue as you pointed out. So I agree we should sleep (via stairway retries) before the next call. Happy to discuss on GVC if that's better.

2: Yes, stairway would just fail this flight with no retry. Would it be better if I made that explicit? (e.g. throw a checked exception and then return a FATAL status in the step)

yonghaoy: 1: `if retrieveProject(123) finds a project, then sleeping 10 minutes before a call to retrieveProject(456)` where is that logic happened? 
My understanding of your code: 
```
for (int numTries = 0; numTries < MAX_RETRIES; numTries++) {
      String projectId = generateId(projectIdSchema);
      if (projectId.length() <= MAX_LENGTH_GCP_PROJECT_ID)
        if (retrieveProject(rmCow, projectId).isEmpty()) {
          logger.info("Generated GCP project id after {} tries.", numTries);
          return projectId;
        } else {
          logger.info("Generated GCP project is already in use: {}", projectId);
        }
    }
```
The sequence can be: 
 a: generate 123
b: retrieve 123, used
c: generate 456
d: retrieve 456, also used,
e, f, g....

I did not find the 10 min sleep during b and d. 

2: I am fine with what you have now: Not catching anything and just let Stairway failed with no retry. Initially I throught catch exception then return **a retryable status in the step** because it is retryable IMO. But I think what you have is also fine, just let it fail, and a new flight will be kicked off, and do the same thing..

yonghaoy: Sorry, if my comment is confusing. My suggestion is similar to the the code in my previous comment:
https://github.com/DataBiosphere/terra-resource-buffer/pull/180#discussion_r681788451 pull retrieveProject() out from the helper method : `generateId()`, and make it back to project creation step. Your helper generateId(). Let me make a PR to explain it

yonghaoy: There it is: https://github.com/DataBiosphere/terra-resource-buffer/pull/181/files 
What you said: `I move retrieveProject into the step and return a retry exception if it failed` I did not see that is happening. retrieveProject is not in the step, it is still in generateId() method with no retry

yonghaoy: I don't want to block your PR. Given the 12 char limit you add, and the low chance of having conflict, no sleep between  `retrieveProject` is not my big concern for now. Totally fine to proceed with your change. 
LGTM

yonghaoy (approved):

src/main/java/bio/terra/buffer/service/resource/projectid/GcpProjectIdGenerator.java
marikomedlock: OK to summarize, I think the two options we're talking about are:
option A: retrieve project in generate project id step, and try to generate another one if it's already used.
option B: retrieve the project only in the create project step, and fail the flight if it's already used.

From your earlier comment, I think these are similar because if a flight fails then we will just kick off a new flight to fill the pool. The main difference is that option A retries within a step and option B retries the whole flight. This means the time between attempts to generate an unused project id will be different (zero for option A, however long it takes to start a new flight for option B).

I'm going to put some time on your calendar to make sure we're on the same page.

marikomedlock: Talked to @yonghaoy and we decided on the following:
- Use stairway retry if the generated project id is already found. This is so that we know each stairway step execution only makes a single get call, which makes it easier to decide how many flights can run in parallel. We could immediately try generating a new project id and checking if it exists, but that would mean each stairway step execution could make a variable number of get calls, which would make the above calculation harder.
- Move the `retreiveProject` call from the `CreateProjectStep` to the `GenerateProjectIdStep`, so that we lower the risk of the undo failing because of a project get quota issue. The generate project id step undo is a no-op, because no project has been created yet. The create project step undo includes another project get call so we can try to delete it.

yonghaoy (approved):

src/main/java/bio/terra/buffer/service/resource/flight/GenerateProjectIdStep.java
```diff
@@ -2,33 +2,68 @@
...
-    String projectId = projectIdGenerator.generateId(gcpProjectConfig.getProjectIdSchema());
-    flightContext.getWorkingMap().put(GOOGLE_PROJECT_ID, projectId);
-    workingMap.put(
-        CLOUD_RESOURCE_UID,
-        new CloudResourceUid().googleProjectUid(new GoogleProjectUid().projectId(projectId)));
-    return StepResult.getStepResultSuccess();
+    try {
+      String projectId =
+          projectIdGenerator.generateIdWithRetries(gcpProjectConfig.getProjectIdSchema());
+
```
jmczerk, line 40: Thanks for the detailed docs here, it is really helpful in reviewing!

jmczerk (approved):
Thank you for the detailed info in the PR, as well as the helpful comments.

The only suggestion I'd make (not nesc in this PR unless it's easy) would be to add a test to exercise the "I tried to generate a name 100 times and gave up" path.

marikomedlock:
> The only suggestion I'd make (not nesc in this PR unless it's easy) would be to add a test to exercise the "I tried to generate a name 100 times and gave up" path.

Good idea. I'll plan to merge this PR so we have a new `tools` pool that I can test in my personal env. Then will investigate including the test in the follow-on PR to add a `dev` pool.

PR #182, opened 2021-08-05
CA-1466 Enable default compute SA in perf and fiab

gpolumbo-broad:
https://broadworkbench.atlassian.net/browse/CA-1466

### DIFF
--- a/src/main/resources/config/perf/resource-config/cwb_ws_perf_v3.yml
+++ b/src/main/resources/config/perf/resource-config/cwb_ws_perf_v3.yml
@@ -28,5 +28,6 @@ gcpProjectConfig:
     - "storage-component.googleapis.com"
   network:
     enableNetworkMonitoring: "false"
+    enablePrivateGoogleAccess: "true"
   computeEngine:
     keepDefaultServiceAcct: "true"
--- a/src/main/resources/config/perf/resource-config/cwb_ws_perf_v3.yml
+++ b/src/main/resources/config/perf/resource-config/cwb_ws_perf_v3.yml
@@ -1,0 +1,32 @@
+# Community Workbench buffered workspace template
+---
+configName: "cwb_ws_perf_v3"
+gcpProjectConfig:
+  projectIdSchema:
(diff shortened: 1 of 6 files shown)


Qi77Qi:
I don't think I'm eligible to review this PR since I don't know much how RBS functions and how these configs work..there might be other solutions other than enabling default SA for all VMs, but here's the context for the change:

Leonardo needs the default SA to exist only for creating galaxy app (or any gke app) because gke nodes need the default SA to function (note we can't use a PET SA for this because gke cluster can be shared among multiple users)....

andy7i (approved):
👍  disclosure: i was part of the mob that made this change

yonghaoy (approved):

marikomedlock (approved):

src/main/resources/config/perf/resource-config/cwb_ws_perf_v3.yml
```diff
@@ -0,0 +1,32 @@
...
+    - "dataproc.googleapis.com"
+    - "dns.googleapis.com"
+    - "genomics.googleapis.com"
+    - "lifesciences.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
+  network:
+    enableNetworkMonitoring: "false"
```
yonghaoy, line 30: FYI, you can add 
`enablePrivateGoogleAccess = true` under network section, that will be needed by  Cromwell soon, and it would be much easier if we make the change & test before PPW launch. I think this should not break anything since AoU is already in enablePrivateGoogleAccess mode, and Leo works fine with that.

yonghaoy (approved):

gpolumbo-broad:
Added `enablePrivateGoogleAccess` per https://github.com/DataBiosphere/terra-resource-buffer/pull/184 and conversation here:  https://broadinstitute.slack.com/archives/CNXF2BAPJ/p1628194823020500

PR #183, opened 2021-08-05
PF-228 (part 2): Add new dev WSM pool with new naming scheme.

marikomedlock:
- Added a new dev WSM pool that uses the `TWO_WORDS_NUMBER` project naming scheme, and a shorter prefix than before (`terra-wsm-dev`->`terra-wsm-d`). There is already a tools WSM pool that I tested on my personal environment WSM deployment and confirmed that projects with the new naming scheme are returned when creating a new workspace.
- Added a unit test for when the maximum number of retries have been exhausted in generating a new project id.

I'm planning to remove the previous `workspace_manager_v8` pools in a follow-on PR once I move all the deployments over to using the new `workspace_manager_v9` pools.

### DIFF
--- a/src/main/resources/config/dev/pool_schema.yml
+++ b/src/main/resources/config/dev/pool_schema.yml
@@ -7,6 +7,9 @@ poolConfigs:
   - poolId: "workspace_manager_v8"
     size: 500
     resourceConfigName: "workspace_manager_v8"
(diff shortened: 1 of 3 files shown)


yonghaoy (approved):

jmczerk (approved):
Thank you for adding the test!

PR #184, opened 2021-08-05
[PF-924]Add private Google Access flag to controll whether to enable private Google Access

yonghaoy:
Background: When RBS is designed, private Google Access + Flow logs is just used by AoU, so we combined them into enableNetworkMonitroring. 
But with the new Cromwell requirements, it wants all CWB projects having private Google Access Enabled but disable flow logs.
So I split them back to two.
If private Google Access is true, we will:

1. Set this value in Subnets
2.  Create private DNS record
3. Record sets
4. Create Custom static Route.
All steps is https://cloud.google.com/vpc/docs/configure-private-google-access
And https://github.com/broadinstitute/gcp-dm-templates/blob/bf2ec422869d108670ea34f93db3e19d5d07b017/firecloud_project.py

### DIFF
--- a/src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectConfigUtils.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectConfigUtils.java
@@ -35,8 +35,9 @@ public class GoogleProjectConfigUtils {
    * true regardless.
    */
   public static boolean isUsePrivateGoogleAccess(GcpProjectConfig gcpProjectConfig) {
-    return isNetworkMonitoringEnabled(gcpProjectConfig) || (gcpProjectConfig.getNetwork() != null
-        && gcpProjectConfig.getNetwork().isUsePrivateGoogleAccess() != null
-        && gcpProjectConfig.getNetwork().isUsePrivateGoogleAccess());
+    return isNetworkMonitoringEnabled(gcpProjectConfig)
+        || (gcpProjectConfig.getNetwork() != null
+            && gcpProjectConfig.getNetwork().isUsePrivateGoogleAccess() != null
+            && gcpProjectConfig.getNetwork().isUsePrivateGoogleAccess());
   }
 }
--- a/src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectConfigUtils.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectConfigUtils.java
@@ -29,10 +29,14 @@ public class GoogleProjectConfigUtils {
         && gcpProjectConfig.getComputeEngine().isKeepDefaultServiceAcct();
   }
 
-  /** Checks if private Google Access enabled. */
+  /**
+   * Checks if private Google Access enabled. Using network monitoring requires private Google
+   * Access. So if {@link #isNetworkMonitoringEnabled(GcpProjectConfig)} is true, this will also be
+   * true regardless.
+   */
   public static boolean isUsePrivateGoogleAccess(GcpProjectConfig gcpProjectConfig) {
-    return gcpProjectConfig.getNetwork() != null
+    return isNetworkMonitoringEnabled(gcpProjectConfig) || (gcpProjectConfig.getNetwork() != null
         && gcpProjectConfig.getNetwork().isUsePrivateGoogleAccess() != null
-        && gcpProjectConfig.getNetwork().isUsePrivateGoogleAccess();
+        && gcpProjectConfig.getNetwork().isUsePrivateGoogleAccess());
   }
 }
--- a/src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectConfigUtils.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectConfigUtils.java
@@ -34,10 +34,10 @@ public class GoogleProjectConfigUtils {
    * Access. So if {@link #isNetworkMonitoringEnabled(GcpProjectConfig)} is true, this will also be
    * true regardless.
    */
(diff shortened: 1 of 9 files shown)


src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectConfigUtils.java
```diff
@@ -28,4 +28,16 @@ public static boolean keepComputeEngineServiceAcct(GcpProjectConfig gcpProjectCo
         && gcpProjectConfig.getComputeEngine().isKeepDefaultServiceAcct() != null
         && gcpProjectConfig.getComputeEngine().isKeepDefaultServiceAcct();
   }
+
+  /**
+   * Checks if private Google Access enabled. Using network monitoring requires private Google
+   * Access. So if {@link #isNetworkMonitoringEnabled(GcpProjectConfig)} is true, this will also be
+   * true regardless.
+   */
+  public static boolean isUsePrivateGoogleAccess(GcpProjectConfig gcpProjectConfig) {
```
wchamber, line 37: nit: maybe just `use` instead of `isUse`?

src/main/resources/config/resource_schema.yaml
```diff
@@ -101,12 +101,19 @@ components:
           description: |-
             Make network traffic is measured. If enabled, turn on flow logs, enable private google access.
             See https://docs.google.com/document/d/1ccz2kzDL68CPofZ-b95ykQIudCgh7OCSFJ4Ym4Oz0r0/edit for more context.
+            Note: Having NetworkMonitoring requires usePrivateGoogleAccess to be true. So usePrivateGoogleAccess will be true
```
wchamber, line 104: nit: `Having enableNetworkMonitoring...`

src/main/resources/config/resource_schema.yaml
```diff
@@ -101,12 +101,19 @@ components:
...
+            if this is set to true
           type: boolean
         keepDefaultNetwork:
           description: |-
             Keep the default VPC network if this flag is true, otherwise delete it.
           type: boolean
           default: false
+        usePrivateGoogleAccess:
+          description: |-
+            Whether to config Private Google Access See: https://cloud.google.com/vpc/docs/configure-private-google-access
```
wchamber, line 114: Maybe duplicate the comment about this always being true if enableNetworkMonitoring is true

src/main/java/bio/terra/buffer/service/resource/flight/GoogleProjectConfigUtils.java
```diff
@@ -28,4 +28,16 @@ public static boolean keepComputeEngineServiceAcct(GcpProjectConfig gcpProjectCo
...
+    return isNetworkMonitoringEnabled(gcpProjectConfig)
+        || (gcpProjectConfig.getNetwork() != null
+            && gcpProjectConfig.getNetwork().isUsePrivateGoogleAccess() != null
+            && gcpProjectConfig.getNetwork().isUsePrivateGoogleAccess());
```
wchamber, line 41: now that this has a little bit of logic, consider adding a unit test to verify that this always returns true if networkmonitoring is enabled

src/main/resources/config/resource_schema.yaml
```diff
@@ -101,12 +101,19 @@ components:
...
             See https://docs.google.com/document/d/1ccz2kzDL68CPofZ-b95ykQIudCgh7OCSFJ4Ym4Oz0r0/edit for more context.
+            Note: Having NetworkMonitoring requires usePrivateGoogleAccess to be true. So usePrivateGoogleAccess will be true
+            if this is set to true
           type: boolean
         keepDefaultNetwork:
           description: |-
             Keep the default VPC network if this flag is true, otherwise delete it.
           type: boolean
           default: false
+        usePrivateGoogleAccess:
```
wchamber, line 112: consider `enablePrivateGoogleAccess` to match `enableNetworkMonitoring`

wchamber (approved):

src/main/resources/config/resource_schema.yaml
```diff
@@ -101,12 +101,19 @@ components:
           description: |-
             Make network traffic is measured. If enabled, turn on flow logs, enable private google access.
             See https://docs.google.com/document/d/1ccz2kzDL68CPofZ-b95ykQIudCgh7OCSFJ4Ym4Oz0r0/edit for more context.
+            Note: Having NetworkMonitoring requires usePrivateGoogleAccess to be true. So usePrivateGoogleAccess will be true
+            if this is set to true
           type: boolean
```
marikomedlock, line 106: nit: could you have a `default: false` here as well?

src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -167,7 +167,7 @@ public void testCreateGoogleProject_witIamBindings() throws Exception {
   }
 
   @Test
-  public void testCreateGoogleProject_enableNetworkMonitoring() throws Exception {
+  public void testCreateGoogleProject_enablePrivateGoogleAccessAndFlowLog() throws Exception {
```
marikomedlock, line 170: Maybe you can add a new, similar test for just PrivateGoogleAccess, but not FlowLog.

marikomedlock (approved):

src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java
yonghaoy: I did not add integration test here because each of them takes 10~15 minutes, and they can not run in parallel. 
I tried to not making this integration test not covering different GcpConfig if not too complicated. But using integration test to make sure those steps is able to execute and get what we want.

PR #187, opened 2021-08-24
Data repo - Fake prod project pool

snf2ye:
Data repo has a legacy instance that we refer to as "fake prod." We hope to retire it soon, but in the mean time, it's still in use. It lives in and references a different google project folder than our "real production" instance, so we need a project pool that can live in this different folder.


nmalfroy (approved):
thanks!

samanehsan:
We'll also need to update the `RBS_POOL_ID ` in the fake prod config: https://github.com/broadinstitute/datarepo-helm-definitions/blob/master/fakeprod/datarepo-api.yaml#L24

src/main/resources/config/prod/resource-config/datarepo_fakeprod_v1.yml
```diff
@@ -0,0 +1,20 @@
+# Data Repo buffered workspace template
+---
+configName: "datarepo_fakeprod_v1"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "datarepo"
+    scheme: "RANDOM_CHAR"
+  parentFolderId: "270278425081" #data.test-terra.bio/repos/jade-dev
```
yonghaoy, line 8: have you granted RBS SA permission to that folder?

yonghaoy (commented):
It is a 'fake' prod instance, is that used in production? If not, I think we should not use prod RBS, but use dev/tools instead.

src/main/resources/config/prod/resource-config/datarepo_fakeprod_v1.yml
snf2ye: Yes, the folder is shared between our dev, alpha, staging, etc. environments, so it should already RBS SA permission.

yonghaoy: Yup, dev RBS does have access on that folder.  If using dev RBS, no terraform change is needed!

src/main/resources/config/dev/resource-config/datarepo_fakeprod_v1.yml
```diff
@@ -0,0 +1,20 @@
+# Data Repo buffered workspace template
+---
+configName: "datarepo_fakeprod_v1"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "datarepo"
```
yonghaoy, line 6: nit: not sure if you want to use `datarepo-dev` as prefix to differentiate with `prod` projects

src/main/resources/config/dev/resource-config/datarepo_fakeprod_v1.yml
```diff
@@ -0,0 +1,20 @@
...
+    scheme: "RANDOM_CHAR"
+  parentFolderId: "270278425081" #data.test-terra.bio/repos/jade-dev
+  billingAccount: "0106B0-41CAA9-427C96"
```
yonghaoy, line 9: Is this prod billing accounts? I saw other dev data repo is using 01A82E-CA8A14-367457

yonghaoy (approved):

samanehsan (approved):

PR #188, opened 2021-09-01
[no ticket]: Bring default compute engine SA back in all environments. 

yonghaoy:
The previous PR updates pool configs in 2 envs. This PR made the same setup for other envs.


gpolumbo-broad (approved):

rtitle (approved):
Looks good, thanks. I just hit this in a dev fiab but seems this will fix it. (Seems like qa and perf were previously updated.)

andy7i (changes requested):
I'm not seeing the corresponding pool_scheme.yml changes for prod and the dev fiabs

andy7i (approved):
👍  LGTM!

PR #191, opened 2021-10-20
Yyu insit

yonghaoy:


buffer-clienttests/src/main/resources/configs/HandoutResourcesThenWaitForRefill.json
```diff
@@ -7,7 +7,7 @@
   "testScripts": [
     {
       "name": "HandoutResource",
-      "numberOfUserJourneyThreadsToRun": 500,
+      "numberOfUserJourneyThreadsToRun": 400,
```
wchamber, line 10: Can we comment this JSON about why we're reducing it to less than 500 and why we still think 400 is a good number? Or can we not add comments to JSON, and maybe add a README.md or NOTES.md in this folder?
Up to you on if you think this is worth doing.

PR #192, opened 2021-11-03
[no ticket][risk=no]Increase how long to keep resource before Janitor delete it

yonghaoy:
Set it to longer period so developers can have time to check projects and debug test failures...


Qi77Qi (approved):
thanks a lot! appreciate it!

gpcarr (approved):

yonghaoy:
One test is broken due to a recent Broadinsitution.org policy change. 
Disable that test for now to unblock this feature.

PR #193, opened 2021-11-05
[PF-1154][RW-7391][RW-7554][risk=medium] Config Firewall rules and DNS zone for AoU projects

yonghaoy:
Added two configs that will control whether to make VMs private and DNS zone for gcr.io.
This is needed for AoU workflow and dataproc support, and already been approved by Broad and DRC. 

https://docs.google.com/document/d/1mlbesp3IIO5tO0d_l_ZFnwKQC_No68TeHSBAQGSGwcM/edit
Screen Shot 2021-11-05 at 9 26 38 AM

Screen Shot 2021-11-05 at 9 28 32 AM
Background doc: https://docs.google.com/document/d/1PS7Z8h-kfn8LHhwhrEUQRtbj4h3GZwe2wfOlnOLcNeo/edit?resourcekey=0-uaDxzYW3ZZrB31K2i6hPBQ#


src/main/resources/config/toolsalpha/pool_schema.yml
```diff
@@ -1,12 +1,6 @@
 # RBS Pools Schema for local testing and on personal environment
 ---
 poolConfigs:
-  - poolId: "resource_toolsalpha_v1"
-    size: 3
-    resourceConfigName: "resource_toolsalpha_v1"
-  - poolId: "resource_toolsalpha_v2"
-    size: 3
-    resourceConfigName: "resource_toolsalpha_v2"
```
yonghaoy, line 9: Those two have keep default network and default SA. And it is not working in Broad deployment because BITs add a org policy "skipDefaultNetwork"... So I removed those two pools, and I believe that is not needed anymore after nextflow support using custom SA and network

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -28,21 +30,174 @@
...
+          .setDirection("EGRESS")
+          .setDestinationRanges(Arrays.asList("199.36.153.4/30"))
+          .setPriority(FIREWALL_RULE_PRIORITY_MAP.get(ALLOW_EGRESS_PRIVATE_ACCESS_RULE_NAME))
+          .setAllowed(Arrays.asList(new Firewall.Allowed().setIPProtocol("all")));
+
+  @VisibleForTesting
+  public static final Firewall ALLOW_EGRESS_INTERNEL =
+      new Firewall()
+          .setName(ALLOW_EGRESS_INTERNEL_RULE_NAME)
+          .setDescription("Allow egress ingress traffic on the network.")
```
marikomedlock, line 161: nit: egress internal?

src/main/java/bio/terra/buffer/service/resource/flight/CreateResourceRecordSetStep.java
```diff
@@ -56,45 +76,53 @@ public StepResult doStep(FlightContext flightContext) throws RetryException {
...
+  private void createRecordSetForDnsZone(
+      String projectId, String zoneName, List<ResourceRecordSet> resourceRecordSets)
+      throws IOException {
+    // ManagedZone is already create in previous step so it won't be empty here.
+    // If we got NPE, that means something went wrong in previous step, fine to just throw NPE
+    // here.
+    ManagedZone managedZone =
+        getResource(() -> dnsCow.managedZones().get(projectId, zoneName).execute(), 404).get();
+
+    // Find all ResourceRecordSets to check if A and CNAME already created.
```
marikomedlock, line 107: When would this already be created -- in a step retry?

src/main/resources/config/dev/resource-config/vpc_sc_v3.yml
```diff
@@ -0,0 +1,35 @@
...
+    - "lifesciences.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
+  network:
+    enableNetworkMonitoring: "true"
+    enablePrivateGoogleAccess: "true"
+    enableCloudRegistryPrivateGoogleAccess: "true"
+    blockInternetAccess: "true"
```
marikomedlock, line 33: I think these are not needed since `enableNetworkMonitoring` is true -- is that right? If so, fine to include them for documentation/readability purposes.

src/main/resources/config/resource_schema.yaml
```diff
@@ -113,7 +113,21 @@ components:
...
         enablePrivateGoogleAccess:
           description: |-
-            Whether to config Private Google Access See: https://cloud.google.com/vpc/docs/configure-private-google-access
+            Whether to config Private Google Access for storage and bigquery API. See: https://cloud.google.com/vpc/docs/configure-private-google-access
+            Note: Having NetworkMonitoring requires usePrivateGoogleAccess to be true. So usePrivateGoogleAccess will be true
+            if this is set to true.
+          type: boolean
+          default: false
+        enableCloudRegistryPrivateGoogleAccess:
+          description: |-
```
marikomedlock, line 122: nit: I think the "NetrworkMonitoring=true forces this to be true" comment also applies to this field

src/main/resources/config/toolsalpha/pool_schema.yml
marikomedlock: Agreed I don't think we need those flags anymore. They are no longer set in the WSM pools.

marikomedlock (approved):

src/main/java/bio/terra/buffer/service/resource/flight/CreateDnsZoneStep.java
```diff
@@ -65,24 +79,36 @@ public StepResult doStep(FlightContext flightContext) throws RetryException {
...
       return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
     }
     return StepResult.getStepResultSuccess();
   }
 
+  private void createManagedDnsZone(
+      String projectId, Network network, ManagedZone managedZoneTemplatae) throws IOException {
+    // Skip ManagedZone creation if ManagedZone already present.
+    ManagedZone managedZone =
+        managedZoneTemplatae
```
calbach, line 99: template

src/main/java/bio/terra/buffer/service/resource/flight/CreateDnsZoneStep.java
```diff
@@ -65,24 +79,36 @@ public StepResult doStep(FlightContext flightContext) throws RetryException {
...
+            .clone()
+            .setPrivateVisibilityConfig(
+                new ManagedZonePrivateVisibilityConfig()
+                    .setNetworks(
+                        ImmutableList.of(
+                            new ManagedZonePrivateVisibilityConfigNetwork()
+                                .setNetworkUrl(network.getSelfLink()))));
+
+    createResourceAndIgnoreConflict(
+        () -> dnsCow.managedZones().create(projectId, managedZone).execute());
```
calbach, line 109: 🐮

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -28,21 +30,174 @@
...
+      DEFAULT_NETWORK_NAME + "-vpc-" + LEONARDO_SSL_FOR_VPC_NETWORK_RULE_NAME;
+
+  /**
+   * Firewall rules to make VM private but still allowing leonardo can access internet(exclude
+   * leonardo-worker which is used by Leo created dataproc worker).
+   */
+  @VisibleForTesting public static final String DENY_EGRESS_RULE_NAME = "deny-egress-all";
+
+  @VisibleForTesting
+  public static final String DENY_EGRESS_LEONARDO_WORKER_RULE_NAME = "deny-egress-leonardo-worker";
```
calbach, line 57: This would require coordination with @rtitle , but I'm wondering whether we should instead name this tag something like:

`leonardo-internal`, `leonardo-private`, `leonardo-restricted`, `leonardo-limited`

Basically these firewall rules could be applied to all projects, but the presence of the network tag dictates whether egress limiting is actually enforced. If we call it `leonardo-worker`, it's a bit awkward that Leo only tags workers with this when specially requested (the workers are always workers - that part doesn't change).

I suppose if we're always planning to have the blockage driven by the RBS folder configuration, the current pattern is OK though.

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -28,21 +30,174 @@
...
+          .setPriority(FIREWALL_RULE_PRIORITY_MAP.get(ALLOW_EGRESS_PRIVATE_ACCESS_RULE_NAME))
+          .setAllowed(Arrays.asList(new Firewall.Allowed().setIPProtocol("all")));
+
+  @VisibleForTesting
+  public static final Firewall ALLOW_EGRESS_INTERNEL =
+      new Firewall()
+          .setName(ALLOW_EGRESS_INTERNEL_RULE_NAME)
+          .setDescription("Allow egress ingress traffic on the network.")
+          .setDirection("EGRESS")
+          .setDestinationRanges(Arrays.asList("10.128.0.0/9"))
```
calbach, line 163: Please add a comment for how you got this

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -28,21 +30,174 @@
...
+  public static final String DENY_EGRESS_LEONARDO_WORKER_RULE_NAME = "deny-egress-leonardo-worker";
+
+  @VisibleForTesting
+  public static final String ALLOW_EGRESS_LEONARDO_RULE_NAME = "allow-egress-leonardo";
+
+  @VisibleForTesting
+  public static final String ALLOW_EGRESS_PRIVATE_ACCESS_RULE_NAME = "allow-egress-private-access";
+
+  @VisibleForTesting
+  public static final String ALLOW_EGRESS_INTERNEL_RULE_NAME = "allow-egress-internal";
```
calbach, line 66: INTERNAL

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -28,21 +30,174 @@
...
+
+  /**
+   * Firewall rule to priority map. The lower number is, the higher priority.
+   *
+   * <p>To make sure only Leo VMs have private internet access when {@code blockInternetAccess} is
+   * true, the priority of egress firewall rules need to be:
+   *
+   * <ul>
+   *   <oi>ALLOW_EGRESS_PRIVATE_ACCESS_RULE_NAME <oi>ALLOW_EGRESS_INTERNEL(equal)
```
calbach, line 75: INTERNET

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -28,21 +30,174 @@
...
+                  new Firewall.Allowed().setIPProtocol("udp").setPorts(Arrays.asList("0-65535"))));
+
+  @VisibleForTesting
+  public static final Firewall ALLOW_EGRESS_LEONARDO =
+      new Firewall()
+          .setName(ALLOW_EGRESS_LEONARDO_RULE_NAME)
+          .setDescription("Allow Leonardo created VMs accessing internet")
+          .setDirection("EGRESS")
+          .setDestinationRanges(Arrays.asList("0.0.0.0/0"))
+          .setTargetTags(Arrays.asList("leonardo"))
```
calbach, line 178: Is this correct? Why is it targetTags and not sourceTags?

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -28,21 +30,174 @@
...
+          .setAllowed(Arrays.asList(new Firewall.Allowed().setIPProtocol("all")));
+
+  @VisibleForTesting
+  public static final Firewall DENY_EGRESS_LEONARDO_WORKER =
+      new Firewall()
+          .setName(DENY_EGRESS_LEONARDO_WORKER_RULE_NAME)
+          .setDescription("Block Leonardo-worker VMs accessing internet")
+          .setDirection("EGRESS")
+          .setDestinationRanges(Arrays.asList("0.0.0.0/0"))
+          .setTargetTags(Arrays.asList("leonardo-worker"))
```
calbach, line 189: Same here

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -28,21 +30,174 @@
...
+      ImmutableMap.<String, Integer>builder()
+          .put(ALLOW_INTERNAL_FOR_VPC_NETWORK_RULE_NAME, 65534)
+          .put(LEONARDO_SSL_FOR_VPC_NETWORK_RULE_NAME, 65534)
+          .put(ALLOW_INTERNAL_FOR_DEFAULT_NETWORK_RULE_NAME, 65534)
+          .put(LEONARDO_SSL_FOR_DEFAULT_NETWORK_RULE_NAME, 65534)
+          .put(DENY_EGRESS_RULE_NAME, 4000)
+          .put(DENY_EGRESS_LEONARDO_WORKER_RULE_NAME, 2000)
+          .put(ALLOW_EGRESS_LEONARDO_RULE_NAME, 3000)
+          .put(ALLOW_EGRESS_PRIVATE_ACCESS_RULE_NAME, 1000)
+          .put(ALLOW_EGRESS_INTERNEL_RULE_NAME, 1000)
```
calbach, line 90: internal

src/main/resources/config/resource_schema.yaml
```diff
@@ -113,7 +113,21 @@ components:
...
+            Whether to config Private Google Access for gcr.io.
+          type: boolean
+          default: false
+        blockInternetAccess:
+          description: |-
+            Whether to allow virtual machines have internet access. If ture, few firewall rules will be created to block egress
```
calbach, line 128: many typos on this line

src/main/resources/config/resource_schema.yaml
```diff
@@ -113,7 +113,21 @@ components:
...
+            Note: Having NetworkMonitoring requires usePrivateGoogleAccess to be true. So usePrivateGoogleAccess will be true
+            if this is set to true.
+          type: boolean
+          default: false
+        enableCloudRegistryPrivateGoogleAccess:
+          description: |-
+            Whether to config Private Google Access for gcr.io.
+          type: boolean
+          default: false
+        blockInternetAccess:
```
calbach, line 126: I think this should be named more specifically. blockBatchInternetAccess, perhaps? Or maybe even more specifically: `blockLeonardoWorkerEgress`

yonghaoy: This also blocks PAPI VMs. (all VM without `leonardo `tag)

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
rtitle: I think I like `leonardo-private`, it's a bit more descriptive what the firewall rule actually does. @yonghaoy shall we agree on `leonardo-private`?

yonghaoy: @rtitle leonardo-private sounds good!

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
yonghaoy: This is correct, per https://cloud.google.com/vpc/docs/firewalls#rule_assignment

`For an egress (outbound) rule, the target designates the source instances.`

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
yonghaoy: ditto

src/main/resources/config/resource_schema.yaml
yonghaoy: networkMonitoring is just enable flow logs now. It doesn't require gcr

src/main/java/bio/terra/buffer/service/resource/flight/CreateResourceRecordSetStep.java
yonghaoy: Yup!

src/main/resources/config/dev/resource-config/vpc_sc_v3.yml
yonghaoy: enableNetworkMonitoring only turn on flow logs on the GCP now. (previously it turns on flow logs and setup private Googel access, not those two are separated configs)

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
calbach: ✅

src/main/resources/config/resource_schema.yaml
calbach: OK, then how about my first suggestion?

My point is that it doesn't block all Internet, e.g. Leonardo GCE or Dataproc master node, perhaps other things in the future.

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
calbach: I suppose it should be all regions for generality of the feature. I believe that should be OK, but I am also reviewing the VPC network docs - they are fairly complicated..

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
calbach: I guess this was actually supposed to say "INTERNAL" - not fixed yet

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
calbach: still says INTERNEL

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
yonghaoy: There is another alternative being discussed in `dsp-callisto` channel. i.e. Let Leo create the firewall rule by the region. It might be preferred if we want to add regional control, i.e. disallow users use certain regions

yonghaoy: RE: VPC network docs. VPC has two mode: Auto create or manually create. Because we want to custom things like enable flow logs, so we have to use the manually mode. 
In the manually mode, we will have to create the subnet ourself, so we tried to follow the "auto-mode" subnet steps to create them, including which IP range to use: https://cloud.google.com/vpc/docs/vpc#ip-ranges

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
yonghaoy: ah, sorry, not pushed my commit yet

src/main/resources/config/resource_schema.yaml
marikomedlock: Yes, but if you enable networkMonitoring, then I think this is forced to true in `GoogleProjectConfigUtils. enableGcrPrivateGoogleAccess`.

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -28,21 +30,174 @@
...
+          .build();
+
+  @VisibleForTesting
+  public static final Firewall ALLOW_INTERNAL_VPC_NETWORK =
```
yonghaoy, line 98: @rtitle Here is the allow internal rule that I added per slack discussion. PTAL

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -28,21 +31,177 @@
...
+          .setPriority(FIREWALL_RULE_PRIORITY_MAP.get(ALLOW_EGRESS_PRIVATE_ACCESS_RULE_NAME))
+          .setAllowed(Arrays.asList(new Firewall.Allowed().setIPProtocol("all")));
+
+  @VisibleForTesting
+  public static final Firewall ALLOW_EGRESS_INTERNAL =
+      new Firewall()
+          .setName(ALLOW_EGRESS_INTERNAL_RULE_NAME)
+          .setDescription("Allow internal egress traffic on the network.")
+          .setDirection("EGRESS")
+          .setDestinationRanges(new ArrayList(REGION_TO_IP_RANGE.values()))
```
yonghaoy, line 167: @rtitle here is the allow egress rule

src/main/resources/config/resource_schema.yaml
yonghaoy: ah, good catch! That is. copy paste error, I removed `isNetworkMonitoringEnabled` from the new added method. 
Thanks!

src/main/resources/config/resource_schema.yaml
yonghaoy: I see, changed to blockLeonardoWorkerEgress

yonghaoy (commented):
@calbach I removed the code to create new resource pool for AoU because we identity a multi-region issue. I will have a follow up PR to fix that, along with new poll creation

calbach (approved):
LGTM for the rules (pending the multiregion discussion). I'm not familiar with the resource buffer codebase so I did not closely review the rest of the code changes.

PR #194, opened 2021-11-09
[IA-3040]: setup lables for Leo https and allow internal firewall rules.

yonghaoy:
Also bump CWB and AoU pool version. 
Screen Shot 2021-11-09 at 9 22 00 PM
Screen Shot 2021-11-09 at 9 21 34 PM


src/main/java/bio/terra/buffer/service/resource/flight/CreateProjectStep.java
```diff
@@ -25,6 +27,16 @@
   @VisibleForTesting public static final String NETWORK_LABEL_KEY = "vpc-network-name";
   @VisibleForTesting public static final String SUB_NETWORK_LABEL_KEY = "vpc-subnetwork-name";
   @VisibleForTesting public static final String CONFIG_NAME_LABEL_KEY = "buffer-config-name";
+  // Firewall rule name to allow https traffic for leoanrdo VMs. Empty if not having such firewall
+  // rule.
+  @VisibleForTesting
+  public static final String LEONARDO_ALLOW_HTTPS_FIREWALL_RULE_NAME_LABEL_KEY =
```
yonghaoy, line 33: @Qi77Qi here are the label keys

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -31,7 +31,7 @@
 public class CreateFirewallRuleStep implements Step {
   /** Names for firewall rules on the high-security network (called 'network'). */
   @VisibleForTesting
-  public static final String ALLOW_INTERNAL_FOR_VPC_NETWORK_RULE_NAME = "allow-internal";
```
yonghaoy, line 34: label values

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -100,6 +100,7 @@
           .setName(ALLOW_INTERNAL_FOR_VPC_NETWORK_RULE_NAME)
           .setDescription("Allow internal ingress traffic on the network.")
           .setDirection("INGRESS")
+          .setTargetTags(Arrays.asList("leonardo"))
```
yonghaoy, line 103: it was filtered by with `leonardo` before. I tried to launch a PAPI job without this and PAPI job succeeded.So I

src/main/java/bio/terra/buffer/service/resource/flight/CreateProjectStep.java
Qi77Qi: I'm wondering if it makes sense to introduce a `version` field for RBS config file in labels...that way leo can check this key to make certain assumptions

yonghaoy: That is the config name(pool config is versioned, and labled) is created for, but it is hard for Leo to use because:
(1) Not all the pools are the same, and upgrade together, e.g., TDR upgrade less frequently than AoU, then than CWB.
(2) not all envs upgrade at the same time, dev iterates faster than prod usually.
(3) Terra-AoU Env is not 1 to 1 mapping, one poll is used by AoU prod, preprod, stable. 

That means there won't be a single 'version' field can applied to all pools.

Qi77Qi: I'm not saying using an existing label...I just meant in project labels, can we have a say `rbs_config_version` key, this will indicate the rbs definition that the project is created with. Say v2 doesn't support region-x, but v3 added support for region-x etc.

> not all envs upgrade at the same time, dev iterates faster than prod usually.
that's fine. It just means each env might have different versions.

yonghaoy: > existing label

The existing label is created for track RBS pool version upgrade. And I believe that is the most common use case. I don't think we have an easy way/better solution to have a unified version number on a project to track project upgrade.  RBS is used by AoU, Rawls, TDR, and WSM. Supporting a feature or not may also depends on the pool schema it is using, for example, my most recent PR created few firewall rules, but it is only enabled for AoU. If we put an "RBS v3" label on all projects, it would confusing WSM and CWB. The right way is to track by the pool schema. Moving forward, it's quite possible that we add another firewall to block AoU projects use non-US regions.

the exiting label is a special case that we globally added that(which is unnecessary for TDR actually). Leo checking version might be broken after we introduce new rule/pool configs(e.g. a config called blockNonUsCompute for AoU). I would trust making a GCP call to pull all firewall rules to decide whether to create my own firewall rule rather than rely on few combination of RBS labels.
Can you use existing labels to compish what Leo needs?  Other ideas other than use firewall rules: Use pool config lables, for example: In prod vpc_sc_v3 or cwb_v3 would have that regions enabled

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -83,10 +83,10 @@
   @VisibleForTesting
   public static final ImmutableMap<String, Integer> FIREWALL_RULE_PRIORITY_MAP =
       ImmutableMap.<String, Integer>builder()
-          .put(ALLOW_INTERNAL_FOR_VPC_NETWORK_RULE_NAME, 65534)
-          .put(LEONARDO_SSL_FOR_VPC_NETWORK_RULE_NAME, 65534)
-          .put(ALLOW_INTERNAL_FOR_DEFAULT_NETWORK_RULE_NAME, 65534)
-          .put(LEONARDO_SSL_FOR_DEFAULT_NETWORK_RULE_NAME, 65534)
+          .put(ALLOW_INTERNAL_FOR_VPC_NETWORK_RULE_NAME, 1000)
```
calbach, line 86: is it correct for these all to have the same priority? Aren't there conflicts?

nit: sort by priority?

yonghaoy: those all allow ingress, and not conflict with each other, so should be fine. (also copied from existing Leo's firewall rule)

yonghaoy:
I will add pools for all envs. Then wait for one release cycle. On Thanks giving week, I will make rawls change to let it use the new created pools.

calbach (approved):

Qi77Qi (approved):
lgtm

PR #196, opened 2021-11-23
[RW-7587][risk=no]: Add flow log filter that filter out inter-VPC network traffic 

yonghaoy:
Background context: We are seeing increase cost when using dataproc. And it turns out caused by flow log cost.
I think master node and worker nodes communicates a lot hence generates lot of flow logs.
This PR filter out those logs that generated within VPC-SC.


rtitle (approved):
Makes sense.. LGTM

marikomedlock (approved):

PR #198, opened 2021-11-24
[RW-7587] Improve flow log filters

yonghaoy:
The previous log filter does not work due to a log filter CEL support limitation.
When dest_src is null (in such case, the destination is outside VPC-SC, so it does not such thing), it errors out did not generate any flow logs. 
I searched around, and found another team had similar issues, and here is the only workaround we can come up with. 

Tested locally:
```
 gcloud compute networks subnets update projects/terra-vpc-sc-dev-da501acd/regions/us-central1/subnetworks/subnetwork --logging-filter-expr="reporter=='SRC' && \!inIpRange(connection.dest_ip, '10.128.0.0/20') &&  \!inIpRange(connection.dest_ip, '199.36.153.4/30')"
``` 
Then run some dataproc, seems to be working well.
I also filtered out the traffic that using Google APIs (GCS and BQ) which is already protected by service perimeter.
All those filters are from:
https://github.com/broadinstitute/terraform-ap-deployments/blob/master/perimeters/vpc-flow-egress-alert.tf#L120L124


rtitle (approved):
Interesting.. looks ok to me, as long as it’s tested in dev

PR #199, opened 2021-11-30
[DDO-1729] [WIP] Expose metrics to Prometheus

jack-r-warren:
Exposes an `/actuator/prometheus` endpoint from Buffer. It grabs all the OpenCensus stats and includes some other [JVM and system stuff](https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.metrics.supported) that we don't really need.


jack-r-warren:
Closing in favor of #200

PR #200, opened 2021-12-01
[DDO-1729] Expose Prometheus /metrics on port 9098

jack-r-warren:
Exposes a Prometheus server exposing the `/metrics` endpoint on another port, 9098 by default. The port being different from the app itself means that the metrics server won't be exposed outside the cluster. Metrics themselves are pulled from OpenCensus just like what gets exported to Stackdriver.

This is enabled by default but not actually scraped by default (telling Prometheus about the endpoint happens in Helm, we're working on abstracting that out to a Helm library).

Supersedes an alternative approach from #199. The functionality developed here was ported into TCL in https://github.com/DataBiosphere/terra-common-lib/pull/62.


src/main/java/bio/terra/buffer/app/configuration/PrometheusConfiguration.java
```diff
@@ -0,0 +1,31 @@
+package bio.terra.buffer.app.configuration;
```
yonghaoy, line 1: For SpringConfiguration in common-lib, it would be two classes:
A Property: https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/stairway/StairwayProperties.java
And a Config: https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/stairway/StairwayConfig.java

Then for Buffer to use this common lib, it would be 
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/application.yml#L22L34

To override them in Helm, they would be: 
https://github.com/broadinstitute/terra-helm/blob/8bde75705e96b966413562b7aa1aac19f035494f/charts/buffer/templates/deployment.yaml#L172L178

src/main/java/bio/terra/buffer/service/prometheus/PrometheusEndpoint.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import java.io.IOException;
+import java.util.Objects;
+
+/**
+ * A component that automatically runs a Prometheus endpoint server on a different port,
+ * exporting stats from OpenCensus.
+ */
+@Component
```
yonghaoy, line 20: https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/app/Main.java#L12
is how to let Spring scan common-lib's package

src/main/java/bio/terra/buffer/app/configuration/PrometheusConfiguration.java
jack-r-warren: Saved me several hours lol thank you

src/main/java/bio/terra/buffer/service/prometheus/PrometheusEndpoint.java
```diff
@@ -0,0 +1,59 @@
...
+    private final PrometheusConfiguration prometheusConfiguration;
+
+    private HTTPServer prometheusServer;
+
+    @Autowired
+    public PrometheusEndpoint(PrometheusConfiguration prometheusConfiguration) {
+        this.prometheusConfiguration = prometheusConfiguration;
+    }
+
+    @PostConstruct
```
yonghaoy, line 32: I never use this annotation before. What I know before is this pattern:
https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/tracing/TracingConfig.java
InitializingBean. But seems both fine: 
https://stackoverflow.com/questions/8519187/spring-postconstruct-vs-init-method-attribute

yonghaoy (approved):
We might want to turn this off in local test:
https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/test/resources/application-test.yml#L32

em-may (approved):

PR #202, opened 2021-12-10
[PF-1246] Force update log4j transitive dependencies

zloery:
Force transitive log4j api and slf4j library dependencies to the latest version due to security concerns. We do not use log4j core, so nothing to update there.


yonghaoy (commented):
I saw log4j in the gradle lockfile, but I did not see the package is used anywhere in our code. 
Then do we still need this?

zloery:
Per initial discussion with infosec:

> That's a good question, I'm not sure. If it's possible to upgrade the underlying version of log4j or to implement the mitigation that would be better than not.

We don't use the log4j core implementation anywhere (we use Logback instead) but we do pull in the log4j api and slf4j adaptor libraries as transitive dependencies. I think we don't strictly need this, but I'm not 100% confident in that.

yonghaoy (approved):

ungwudik (approved):

zarsky-broad:
See https://github.com/DataBiosphere/terra-external-credentials-manager/pull/84 and https://docs.spring.io/spring-boot/docs/2.6.1/gradle-plugin/reference/htmlsingle/#managing-dependencies.dependency-management-plugin.customizing

PR #205, opened 2021-12-15
PF-1009: Read config files as system file instead of reading it from the jar.

yuhuyoyo:


src/main/java/bio/terra/buffer/service/pool/PoolConfigLoader.java
```diff
@@ -33,9 +38,17 @@
...
-  public static List<PoolWithResourceConfig> loadPoolConfig(String folderName) {
-    PoolConfigs poolConfigs = parsePools(folderName);
-    Map<String, ResourceConfig> resourceConfigNameMap = parseResourceConfig(folderName);
+  public static List<PoolWithResourceConfig> loadPoolConfig(
+      String folderName, Optional<String> systemFilePath) {
+    PoolConfigs poolConfigs;
+    Map<String, ResourceConfig> resourceConfigNameMap;
+    if (systemFilePath.isPresent()) {
+      poolConfigs = parsePoolsAsSystemFile(systemFilePath.get(), folderName);
+      resourceConfigNameMap = parseResourceConfigAsSystemFile(systemFilePath.get(), folderName);
```
yonghaoy, line 47: 1: Having both `filePath` and `folderName` is a bit confusing and necessary, we can combine this into one path. It can also just be single 'folder name'.  
The existing `configPath` is already a config file that used here: https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/service/pool/PoolService.java#L67

If you want to support system file, instead of setting systemFilePath and configPath, I suggest to just set one env variable, similar to the configFile that we are using. 

2: And actually, systemFile also works if configs live within a jar, you can quickly check this by letting the load a file from path: `config/toolsalpha`.

Whether to use `ClassLoader` that only supports files within a jar or use java `File` class to support all system files is two option we have, we choose ClassLoader because it makes integration test easier (integration test can also load files from main/resources/....)

yonghaoy (commented):
Two questions: 
1: System file may also works for

yonghaoy (commented):
files in a jar OR system file as one of might be not the ideal end state Verily deployment wants. From my understanding, what @jmczerk wants is 
1: still inside jar: Buffer loads 'major configs' from the existing location, i.e. src/main/resources, 
2: plus system file: but also override some secret values (billing account, folder id ) from a different path.

jmczerk:
> files in a jar OR system file as one of might be not the ideal end state Verily deployment wants. From my understanding, what @jmczerk wants is 1: still inside jar: Buffer loads 'major configs' from the existing location, i.e. src/main/resources, 2: plus system file: but also override some secret values (billing account, folder id ) from a different path.

For this ticket (PF-1009) I think the OR approach is fine.  I think the templating discussed in PF-260 is a different issue that's out of scope for this PR;  when implementing that, we'll need a way to reference a base configuration to apply any templated parameters to, so I don't imagine it would be too hard to make that flexible enough to point at a config in the Jar or on the file system.

src/main/java/bio/terra/buffer/service/pool/PoolConfigLoader.java
yuhuyoyo: i agree that having `filePath` and `folderName` together is redundant. I changed it there. 

it's true that I can just remove ClassLoader and set the configPath to the fullpath for broad deployment but just to be cautious and not break everything, I want to flag-guard the change. We can later on set systemFilePath for the broad environments as well and clean up that path. WDYT?

src/main/resources/application.yml
```diff
@@ -14,6 +14,7 @@ buffer:
     # Whether to update pool when config changes during server startup.
     # Might be useful to avoid potential backward compatibility caused by config change and make tests easier.
     update-pool-on-start: true
+    read-config-from-system-path: false
```
jmczerk, line 17: Does this need to change to `system-file-path: ` now?  Would be good to have a comment describing as well.

jmczerk (approved):

src/main/resources/application.yml
```diff
@@ -14,6 +14,8 @@ buffer:
     # Whether to update pool when config changes during server startup.
     # Might be useful to avoid potential backward compatibility caused by config change and make tests easier.
     update-pool-on-start: true
+    # Alternative system file path to read the config files from. If not set, fall back to reading from config-path. 	    
+    system-file-path:
```
yonghaoy, line 18: feels system-file-path should be part of buffer's pool config, like `config-path` in line 13. 
Can you also put it under buffer.pool, then it would be `config-path.system-file-path`

yonghaoy (approved):

src/main/java/bio/terra/buffer/service/pool/PoolConfigLoader.java
yonghaoy: SG. Can you file a ticket under PF-82 with a TODO here to remind us to migrate everything  to system-file path?

src/main/resources/application.yml
yuhuyoyo: It is already in buffer.pool (PoolConfiguraiton.java), below where config-path is defined.

src/main/java/bio/terra/buffer/service/pool/PoolConfigLoader.java
yuhuyoyo: done. https://broadworkbench.atlassian.net/browse/PF-1273

PR #206, opened 2021-12-17
[PF-1152] Add network.blocked_regions to resource schema

melissachang:
Tested by changing `resource_toolsalpha_v4.yml` and observing logging.

- Works if no blocked regions
- Works if some blocked regions
- If there's a typo, see InvalidArgumentException


src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -140,6 +143,26 @@ public StepResult undoStep(FlightContext flightContext) {
...
+  private Map<String, String> getRegionToIpRange() {
+    // Convert to HashMap so we can call removeAll().
+    Map<String, String> regiontoIpRange = new HashMap<>(REGION_TO_IP_RANGE);
+    List<String> blockedRegions = GoogleProjectConfigUtils.blockedRegions(gcpProjectConfig);
+
+    // Throw an exception if blocklist contains an invalid region.
+    Set<String> blockedRegionsCopy = new HashSet<>(blockedRegions);
+    blockedRegionsCopy.removeAll(regiontoIpRange.keySet());
+    if (!blockedRegionsCopy.isEmpty()) {
+      throw new IllegalArgumentException(
```
jmczerk, line 156: Should we consider warning here instead of throwing?  The thought being that if we decide to remove a region from `regiontoIpRange` for some reason, if we have existing pools that block-list that region, they will fail here.  Since we can't update pool configs after they are created, that would necessitate removing the existing pool to allow startup.

src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -140,6 +143,26 @@ public StepResult undoStep(FlightContext flightContext) {
     return StepResult.getStepResultSuccess();
   }
 
+  /** Gets a map of region to IP range. */
+  private Map<String, String> getRegionToIpRange() {
+    // Convert to HashMap so we can call removeAll().
+    Map<String, String> regiontoIpRange = new HashMap<>(REGION_TO_IP_RANGE);
```
jmczerk, line 149: Nit: `regionToIpRange`?

src/main/resources/config/resource_schema.yaml
```diff
@@ -132,6 +132,12 @@ components:
             See: https://docs.google.com/document/d/1elBUrHVoFzT8k5bofDxPi1bLPzJAYed-mc10ZfNz8CY/edit?resourcekey=0-7pDBBKNL6n3K7imr2V0fpA
           type: boolean
           default: false
+        blockedRegions:
+          description: |-
```
jmczerk, line 136: We probably don't need to call this out as Verily-specific; I could see it being useful elsewhere.

src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
melissachang: Done

src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
melissachang: Done

src/main/resources/config/resource_schema.yaml
melissachang: Done

jmczerk:
This looks good, though I'd like to see a test to cover both a valid and invalid region name.  This might be a good model to follow:

https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java#L287

src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
yonghaoy: I don't think we can get warn too much from here. If that is true, the whole method can be something like
```

REGION_TO_IP_RANGE.entrySet().stream().filter(e -> !GoogleProjectConfigUtils.blockedRegions(gcpProjectConfig).contains(e.getKey())).collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()))
```

yonghaoy: And if we do want to keep the warning message, it is still bit hard to read, suggestion:
```
if(REGION_TO_IP_RANGE.keySet().containsAll(blockedRegions)) {
  log.warn()
}
```

melissachang:
> This looks good, though I'd like to see a test to cover both a valid and invalid region name. This might be a good model to follow:
> 
> https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java#L287

I see 2 approaches:

1. In CreateProjectFlightIntegrationTest.java, add `assertSubnetRegions()`. Add Subnetworks.listUsable to CloudComputeCow.java, which returns all regions for a project.
2. Have [existing CreateProjectFlightIntegrationTest.assertSubnetsExist()](https://github.com/DataBiosphere/terra-resource-buffer/blob/e033fec6823a2c1167230410509acff176a8b8b5/src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java#L662) take an additional parameter `blockedRegions`.

I'll do 1, since we're just checking subnet region and not the [other parts of subnet](https://github.com/DataBiosphere/terra-resource-buffer/blob/e033fec6823a2c1167230410509acff176a8b8b5/src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java#L662). And, I'll learn about terra-cloud-resource-lib.

Let me know if anyone prefers 2.

src/main/resources/config/resource_schema.yaml
melissachang: Done

src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
melissachang: I thought warning would be useful if there's a typo. Advantage of current code is that warning message prints the invalid regions. Agree it is hard to read though.

melissachang (commented):
Added integration tests. I ended up not using Subnetworks.AggregatedList. That returns all regions, regardless of whether there's a subnet in that region. If there's no subnet in the region, it returns a warning. Instead of checking for warnings, I think it's cleaner to iterate over blocked regions and call Subnetworks.List(project, region).

It was a fun learning experience though. :)

src/main/resources/config/resource_schema.yaml
```diff
@@ -132,6 +132,13 @@ components:
...
+            Subnetworks will be created for the regions in CreateSubnetsStep.REGION_TO_IP_RANGE, excluding these
+            blockedRegions.
+          type: array
+          items:
+            type: string
```
jmczerk, line 141: Sorry for the late add, but yesterday's hiccup with Spring Boot's handling of the yaml made me think of this.  Does `blockedRegions` need a `default: []` so that this doesn't fail parsing existing configs without this config value present?

jmczerk: Thanks @yonghaoy !

jmczerk (approved):

yonghaoy (approved):

src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
yonghaoy: It's still bit unnecessary to create a Set copy and just for the warn check. 
Is my suggestion above sufficient to check the typo? 
```
if(!REGION_TO_IP_RANGE.keySet().containsAll(blockedRegions)) {
  log.warn()
}
```

zloery (approved):

src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
melissachang: Simplified warning code. Thanks, I agree with you. :)

Can I keep the log.debug? John mentioned it's useful to confirm the binary has this commit.

I'll remove it after everything is done, as part of this Issue. (I was wondering if I should keep it permanently. Thanks for saying that Broad deployment prints all debug -- so I shouldn't keep permanently.)

yonghaoy: that SG!

PR #209, opened 2022-01-07
PF-1009: handle the cases when the system file path is symbolic link 

yuhuyoyo:
when mounting config files to k8s containers, the path is actually a symbolic link. Exception is thrown when trying to read those path.


src/test/java/bio/terra/buffer/config/PoolSchemaTest.java
```diff
@@ -41,6 +42,11 @@ public void testConfigValid_readFromSystemFile() {
     }
   }
 
+  @Test
+  public void loadPoolConfig_systemFilePathIsSymbolicLink_configValid() throws IOException {
+    loadPoolConfig("config", Optional.ofNullable("./src/test/java/bio/terra/buffer/config"));
```
yonghaoy, line 47: nit: unnecessary of using `ofNullable`

jmczerk (commented):
The code change looks good to me, but I'm not sure if it's OK to put config files under `src/test/java`

src/test/java/bio/terra/buffer/config/PoolSchemaTest.java
yonghaoy, line 47: you can check `loadPoolConfig` result to make sure poolSchema finds all resources configs.
```
assertEqual(loadPoolConfig("config", Optional.of("./src/test/java/bio/terra/buffer/config")).size()), 2);
```

yuhuyoyo:
> The code change looks good to me, but I'm not sure if it's OK to put config files under `src/test/java`

hmm I cannot put the symlink file under resources/ though so I just put it closest to where the test is.

yonghaoy (approved):

yonghaoy:
> > The code change looks good to me, but I'm not sure if it's OK to put config files under `src/test/java`
> 
> hmm I cannot put the symlink file under resources/ though so I just put it closest to where the test is.

In Java unit test, it's common to put test data files in the same directory as the unit test to make them close to each other.

jmczerk (approved):

PR #211, opened 2022-01-11
[PF-1278] Allow customize test resource time to live per env

yonghaoy:
Background: Using a static variable as time to live make it impossible to customize this value per env. 
However, we did see a request from TDR and Leo that wants to have different values.


zloery (approved):
Seems fine to make this configurable, but if this is a configuration property, wouldn't there still be a single value per instance? (i.e. Leo and TDR would still have to use the same time if they're talking to the same instance)

yonghaoy:
> Seems fine to make this configurable, but if this is a configuration property, wouldn't there still be a single value per instance? (i.e. Leo and TDR would still have to use the same time if they're talking to the same instance)

yeah, TDR and Leo would still have the same time if they use the same instance. TDR is fine to use another Buffer, (buffertest) as they don't want to have special-case in their end to config tools RBS. 
See https://broadinstitute.slack.com/archives/C016V54EK0B/p1641915415025100

PR #212, opened 2022-01-19
[PF-1152] Delete region blocklist warning

melissachang:
There are now regions in blocklist that aren't in `REGION_TO_IP_RANGE` (see [here](https://github.com/verily-src/terra-deployment/pull/443)) so the warning no longer makes sense.

### DIFF
--- a/src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
@@ -25,6 +25,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


jmczerk (approved):

src/main/java/bio/terra/buffer/service/resource/flight/CreateSubnetsStep.java
```diff
@@ -146,14 +146,7 @@ public StepResult undoStep(FlightContext flightContext) {
     // Convert to HashMap so we can call removeAll().
```
yonghaoy, line 146: The method can just be
```
return REGION_TO_IP_RANGE.entrySet().stream()
.filter(e -> !GoogleProjectConfigUtils.blockedRegions(gcpProjectConfig).contains(e.getKey()))
.collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()))
```

melissachang: Done, except I kept the `blockedRegions` variable

yonghaoy (approved):

PR #213, opened 2022-01-20
[no ticket] Update RBS perf test config to consume less resources

yonghaoy:
Recently RBS perf tests became flaky and often times out. Seems there are more and more services that consuming GCP quota from terra-kernel-k8s project, and slow down project creation.
Decrease the number of projects to create per perf test, and increase the timeout. 

The new config (200 concurrent) is still good enough to capture RBS latency issues, the actual QPS in Community workbench is <0.001/s.

### DIFF
--- a/buffer-clienttests/src/main/resources/configs/HandoutResourcesThenWaitForRefill.json
+++ b/buffer-clienttests/src/main/resources/configs/HandoutResourcesThenWaitForRefill.json
@@ -7,9 +7,9 @@
   "testScripts": [
     {
       "name": "HandoutResource",
-      "numberOfUserJourneyThreadsToRun": 400,
-      "userJourneyThreadPoolSize": 200,
-      "expectedTimeForEach": 2,
+      "numberOfUserJourneyThreadsToRun": 200,
+      "userJourneyThreadPoolSize": 100,
+      "expectedTimeForEach": 3,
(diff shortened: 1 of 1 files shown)


buffer-clienttests/src/main/resources/configs/HandoutResourcesThenWaitForRefill.json
```diff
@@ -7,9 +7,9 @@
   "testScripts": [
     {
       "name": "HandoutResource",
-      "numberOfUserJourneyThreadsToRun": 400,
-      "userJourneyThreadPoolSize": 200,
-      "expectedTimeForEach": 2,
+      "numberOfUserJourneyThreadsToRun": 200,
+      "userJourneyThreadPoolSize": 100,
```
marikomedlock, line 11: This will handout 200 times, up to 100 concurrently. Is that what you intended? Just checking because in the PR description it says the "new config (200 concurrent)".

yonghaoy: what is what I need. In total it handout 200 times, and 100 each. But for Buffer, those two 100 are too close to each other, and it is 200 'concurrent' to Buffer. (Buffer needs to create 200 projects at one time), sorry for the confusion

marikomedlock (approved):

PR #216, opened 2022-02-07
[PF-1228] Use a standalone cloud_resource_uid in Buffer service

yonghaoy:
See  https://broadworkbench.atlassian.net/browse/PF-1228

It's bit unfortunate that we better to keep all the type of resources because that is part of response.


jmczerk (commented):
Code change looks OK, but looks like maybe it needs a dependency lockfile update?

PR #217, opened 2022-02-08
Fix perf test flakiness by reducing numberOfUserJourneyThreadsToRun

melissachang:
Last night's run timed out:

https://github.com/DataBiosphere/terra-resource-buffer/runs/5109523341?check_suite_focus=true

`HandoutResourcesThenWaitForRefill` timed out after 2 hours.

A passing run from a couple days ago:

https://github.com/DataBiosphere/terra-resource-buffer/runs/5032679956?check_suite_focus=true

`HandoutResourcesThenWaitForRefill` took about 1 hr 50 mins.

### DIFF
--- a/buffer-clienttests/src/main/resources/configs/HandoutResourcesThenWaitForRefill.json
+++ b/buffer-clienttests/src/main/resources/configs/HandoutResourcesThenWaitForRefill.json
@@ -7,9 +7,9 @@
   "testScripts": [
     {
       "name": "HandoutResource",
-      "numberOfUserJourneyThreadsToRun": 200,
+      "numberOfUserJourneyThreadsToRun": 150,
       "userJourneyThreadPoolSize": 100,
-      "expectedTimeForEach": 3,
+      "expectedTimeForEach": 2,
       "expectedTimeForEachUnit": "HOURS"
     }
   ],
--- a/src/main/resources/config/buffertest/pool_schema.yml
+++ b/src/main/resources/config/buffertest/pool_schema.yml
@@ -2,5 +2,5 @@
 ---
 poolConfigs:
   - poolId: "resource_buffer_test_v1"
-    size: 1250
+    size: 1500
     resourceConfigName: "resource_buffer_test_v1"
--- a/src/main/resources/config/buffertest/pool_schema.yml
+++ b/src/main/resources/config/buffertest/pool_schema.yml
@@ -2,5 +2,5 @@
 ---
 poolConfigs:
   - poolId: "resource_buffer_test_v1"
-    size: 1500
+    size: 1000
     resourceConfigName: "resource_buffer_test_v1"
--- a/src/main/resources/config/buffertest/pool_schema.yml
+++ b/src/main/resources/config/buffertest/pool_schema.yml
@@ -2,5 +2,5 @@
 ---
 poolConfigs:
   - poolId: "resource_buffer_test_v1"
-    size: 1000
+    size: 1250
     resourceConfigName: "resource_buffer_test_v1"


jmczerk (approved):
This works for me if @yonghaoy is good with it.  My impulse would be to maybe go to 1250, since 1500 is right at the edge, but I'm not super picky.

melissachang:
FYI I manually ran perf test and it failed. I'll investigate later today.

https://github.com/DataBiosphere/terra-resource-buffer/actions/runs/1813964968

melissachang:
> FYI I manually ran perf test and it failed. I'll investigate before merging.
> 
> https://github.com/DataBiosphere/terra-resource-buffer/actions/runs/1813964968

N/m, it failed because perf action does a checkout.

Changed to 1250.

yonghaoy:
The pool size is 1500, but one test run actually just verify it can create 200 projects. Reducing pool size here won't help. (but still good to reduce through because 1500 is a bit too high). 
The actual config is controlled by test runner config. Here is my recent example to reduce the number: 
https://github.com/DataBiosphere/terra-resource-buffer/pull/213

But I throught I changed the timeout to 3hr? Is 2hr is a hard limit in test runner?  @marikomedlock

melissachang:
> The pool size is 1500, but one test run actually just verify it can create 200 projects. Reducing pool size here won't help. (but still good to reduce through because 1500 is a bit too high). The actual config is controlled by test runner config. Here is my recent example to reduce the number: #213
> 
> But I throught I changed the timeout to 3hr? Is 2hr is a hard limit in test runner? @marikomedlock

[Handout step is capped at 30 mins](https://github.com/DataBiosphere/terra-resource-buffer/blob/1d7226593c9f579344ab26e4ba74d3dc105a1e73/buffer-clienttests/src/main/java/scripts/testscripts/HandoutResource.java#L48)

But I would like to advocate for shorter test time. I was actually wondering if we could decrease from 2 hours. I'm not sure if 3 hours gives additional data over 1.5 hours. And it makes things easier if we don't have to wait 3 hours for a run.

I decreased `numberOfUserJourneyThreadsToRun` from 200 to 150 (and changed 3 hours back to 2 hours), how does that look?

yonghaoy (approved):

marikomedlock:
> But I throught I changed the timeout to 3hr? Is 2hr is a hard limit in test runner? @marikomedlock

Test runner doesn't have a hard limit.

The `expectedTimeForEach` in the test runner config applies to each user journey thread. The user journey threads in this perf test just pull a project from the pool. So we give them up to 3hr to do that (with your previous change).

RBS uses test runner in a kind of special case here, because it's actually doing the assertion in the test cleanup method. There are no time limits for setup and cleanup methods in test runner. In this test, there is a hard-coded 2hr timeout [in the cleanup method](https://github.com/DataBiosphere/terra-resource-buffer/blob/1d7226593c9f579344ab26e4ba74d3dc105a1e73/buffer-clienttests/src/main/java/scripts/testscripts/HandoutResource.java#L71). I think that's why your previous change didn't up the timeout in the way we were expecting (sorry I didn't notice this then!).

So looking at this again, I think the timeout for the user journey threads for this test is probably extremely generous (2hrs to retry pulling a project from the pool, which we know is full in the setup method). To make this more configurable, we could pass in the timeout used in the cleanup method(and also the startup method while we're at it) as a test script parameter. That would allow us to adjust the timeouts just from the config, like other tests. ([example](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/integration/src/main/java/scripts/testscripts/ServiceStatus.java) of using parameters in a test script)

PR #219, opened 2022-02-11
[RW-7838] Remove default compute engine SA account for Community Workbench projects

yonghaoy:
Per discussion offline, the default compute engine SA will not be used:
https://broadinstitute.slack.com/archives/C02PC66F6D8/p1644533971276899?thread_ts=1644424280.461739&cid=C02PC66F6D8


src/main/resources/config/dev/pool_schema.yml
```diff
@@ -1,18 +1,15 @@
 # Resource Buffer Service Pools Schema in dev environment
 ---
 poolConfigs:
-  - poolId: "cwb_ws_dev_v4"
-    size: 300
```
yonghaoy, line 5: dev and fiab has been using v5 now, safe to remove v4.

src/main/resources/config/prod/pool_schema.yml
```diff
@@ -8,8 +8,11 @@ poolConfigs:
     size: 3000
     resourceConfigName: "cwb_ws_prod_v4"
   - poolId: "cwb_ws_prod_v5"
-    size: 3000
+    size: 1000
```
yonghaoy, line 11: v5 has not been in production yet, and highly possible that we will skip this version. So change pool size to a smaller number just as a back up.

Qi77Qi (commented):
I can't think of anything uses this..but is there a way to test this in fiabs before we merge? if not, I'm fine with merging and give it a try

Qi77Qi (approved):

jmczerk (approved):
I don't have much context on these pools, but I did review for syntax, agreement of `resourceConfigName` with config file path, etc and LGTM.

PR #221, opened 2022-03-08
Add batch network config settings for VPC-SC prod projects

calbach:
**This is a potential launch blocking issue for AoU**.

The expectation has been that this has been enabled for AoU since 2021. Presumably this was an oversight and the settings should have been configured on https://github.com/DataBiosphere/terra-resource-buffer/pull/194/files#diff-0e1bb8265c93a909f273642f9cfcd825df74f4bf4c337b340ec1da6cbbec8351R29

I don't know whether this also should have been included on the other pool configs in that PR, I'm unfamiliar with those. Not high on my list right now though.

I need guidance in deploying this change.

@Qi77Qi @rtitle @gjuggler FYI

### DIFF
--- a/src/main/resources/config/prod/resource-config/vpc_sc_v7.yml
+++ b/src/main/resources/config/prod/resource-config/vpc_sc_v7.yml
@@ -1,0 +1,35 @@
+# Projects with VPC-SC configuration
+---
+configName: "vpc_sc_v7"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-vpc-sc"
+    scheme: "RANDOM_CHAR"
+  # firecloud.org/prod/for_vpc_sc_unclaimed
+  parentFolderId: "160283235721"
+  billingAccount: "0106B0-41CAA9-427C96"
+  enabledApis:
+    - "bigquery-json.googleapis.com"
+    - "compute.googleapis.com"
+    - "container.googleapis.com"
+    - "cloudbilling.googleapis.com"
+    - "clouderrorreporting.googleapis.com"
+    - "cloudkms.googleapis.com"
+    - "cloudtrace.googleapis.com"
+    - "containerregistry.googleapis.com"
+    - "dataflow.googleapis.com"
+    - "dataproc.googleapis.com"
+    - "dns.googleapis.com"
+    - "genomics.googleapis.com"
+    - "lifesciences.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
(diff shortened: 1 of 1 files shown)


src/main/resources/config/prod/resource-config/vpc_sc_v7.yml
```diff
@@ -0,0 +1,35 @@
...
+    - "lifesciences.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
+  network:
+    enableNetworkMonitoring: "true"
+    enablePrivateGoogleAccess: "true"
+    enableCloudRegistryPrivateGoogleAccess: "true"
+    blockBatchInternetAccess: "true"
```
calbach, line 33: FYI: along with the pool name, L32-33 are the only changes from v6.

calbach:
@zloery if the change looks good to you in the AM - please merge this on my behalf. Want to ensure sure no one is waiting on me to wake up and click the merge button.

src/main/resources/config/prod/resource-config/vpc_sc_v7.yml
```diff
@@ -0,0 +1,35 @@
+# Projects with VPC-SC configuration
```
jmczerk, line 1: This also requires an entry to be added for `vpc_sc_v7` in [config/prod/pool_schema.yml](https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/config/prod/pool_schema.yml).

rtitle: +1

Qi77Qi:
would be nice to roll it out to envs before prod first....looking at a [sample PR](https://github.com/DataBiosphere/terra-resource-buffer/pull/218)

rtitle:
@Qi77Qi I think these 2 settings are already enabled in all non-prod.. see https://github.com/DataBiosphere/terra-resource-buffer/blob/ch/vpc-prod-batch/src/main/resources/config/staging/resource-config/vpc_sc_v5.yml#L32-L33.

Not sure the history here, but looks like this was rolled out everywhere except prod.

jmczerk:
> would be nice to roll it out to envs before prod first....looking at a [sample PR](https://github.com/DataBiosphere/terra-resource-buffer/pull/218)

More discussion [here](https://broadinstitute.slack.com/archives/C6DTFUCDD/p1646703331692809) but I believe it already exists in other envs, there was just an oversight that it hadn't yet made it to `prod`.

rtitle (approved):

zloery (approved):

PR #222, opened 2022-03-14
[RW-7885, PF-1376] Pool config for BigQuery Daily Usage Quota Override

jaycarlton:
Add new pool quota for RBS.

### DIFF
--- a/src/main/resources/config/dev/pool_schema.yml
+++ b/src/main/resources/config/dev/pool_schema.yml
@@ -10,9 +10,9 @@ poolConfigs:
   - poolId: "workspace_manager_v9"
     size: 500
     resourceConfigName: "workspace_manager_v9"
-  - poolId: "vpc_sc_v6"
+  - poolId: "vpc_sc_v7"
     size: 300
-    resourceConfigName: "vpc_sc_v6"
+    resourceConfigName: "vpc_sc_v7"
   - poolId: "datarepo_v3"
     size: 300
     resourceConfigName: "datarepo_v3"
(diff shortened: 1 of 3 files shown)


yuhuyoyo (approved):

src/main/resources/config/dev/pool_schema.yml
```diff
@@ -10,9 +10,9 @@ poolConfigs:
   - poolId: "workspace_manager_v9"
     size: 500
     resourceConfigName: "workspace_manager_v9"
-  - poolId: "vpc_sc_v6"
+  - poolId: "vpc_sc_v7"
```
melissachang, line 13: There's already a `vpc_sc_v7` in prod, so this should be `vpc_sc_v8`.

https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/config/prod/resource-config/vpc_sc_v7.yml

jaycarlton: good catch

src/main/resources/config/dev/pool_schema.yml
```diff
@@ -10,9 +10,9 @@ poolConfigs:
   - poolId: "workspace_manager_v9"
     size: 500
     resourceConfigName: "workspace_manager_v9"
-  - poolId: "vpc_sc_v6"
```
calbach, line 13: I believe this broke dev, this removedd the old pool which was still in use

calbach: Sent https://github.com/DataBiosphere/terra-resource-buffer/pull/223

PR #226, opened 2022-03-15
[RW-7885, PF-1376] update from bytes to MiB

jaycarlton:
I had the base unit wrong; it was supposed to be in MiB, so I was off by a factor of a million or so.

Add a new config, `vp_sc_v9`.

### DIFF
--- a/src/main/resources/config/dev/resource-config/vpc_sc_v8.yml
+++ b/src/main/resources/config/dev/resource-config/vpc_sc_v8.yml
@@ -37,4 +37,4 @@ gcpProjectConfig:
   serviceUsage:
     bigQuery:
       overrideBigQueryDailyUsageQuota: true
-      bigQueryDailyUsageQuotaOverrideValueBytes: 43980465111040 # 40 TB
+      bigQueryDailyUsageQuotaOverrideValueMebibytes: 43980465111040 # 40 TB
--- a/src/main/resources/config/dev/resource-config/vpc_sc_v8.yml
+++ b/src/main/resources/config/dev/resource-config/vpc_sc_v8.yml
@@ -37,4 +37,4 @@ gcpProjectConfig:
   serviceUsage:
     bigQuery:
       overrideBigQueryDailyUsageQuota: true
-      bigQueryDailyUsageQuotaOverrideValueMebibytes: 43980465111040 # 40 TB
+      bigQueryDailyUsageQuotaOverrideValueBytes: 43980465111040 # should be MiB; 40,000 PB
--- a/src/main/resources/config/resource_schema.yaml
+++ b/src/main/resources/config/resource_schema.yaml
@@ -194,9 +194,9 @@ components:
             and there's no way to generate Optional types.
           type: boolean
           default: false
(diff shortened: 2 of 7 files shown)


zloery (approved):

src/main/resources/config/resource_schema.yaml
```diff
@@ -194,9 +194,9 @@ components:
             and there's no way to generate Optional types.
           type: boolean
           default: false
-        bigQueryDailyUsageQuotaOverrideValueBytes:
+        bigQueryDailyUsageQuotaOverrideValueMebibytes:
```
zloery, line 197: I think changing this key will break the existing v8 config. For compatibility, it seems like you might need to introduce a new key -> deprecate and delete the v8 pool -> remove the old key.

Alternatively I wonder if you could edit the key name in the v8 config? I'm not sure if RBS lets you do that

jaycarlton: I'm trying the latter suggestion, b/c it fails a test if I don't. There's only one current schema, so when it changes, all the configs need to adhere to it.

Another option would be to add a new key set to the old value and leave the v8 schema intact, then remove it once we've removed v8. It's more churn, so if the naive way works, I'm going to stick with it.

src/main/resources/config/dev/resource-config/vpc_sc_v8.yml
```diff
@@ -37,4 +37,4 @@ gcpProjectConfig:
   serviceUsage:
     bigQuery:
       overrideBigQueryDailyUsageQuota: true
-      bigQueryDailyUsageQuotaOverrideValueBytes: 43980465111040 # 40 TB
+      bigQueryDailyUsageQuotaOverrideValueMebibytes: 43980465111040 # 40 TB
```
calbach, line 40: The 40TB comment is now outdated - but also, should this number adjusted back down to actually be 40 TB?

Note: if this change is possible, do we need the v9 pool at all? Personally I don't care if the next ~300 projects from v8 have an incorrect BigQuery quota - we'll probably go through those in a week or so. As long as one project gets created with this new setting, we can verify the fix via cloud console.

jaycarlton: My thought process was that we only need to change the v8 pool config to make it loadable under the new schema. I don't know if it's safe to change the value. I suppose we save a round trip to the other repo to point to v9 if we're sure changing v8 will work.

Zach's approach might be the most fool-proof, and would preserve history in the pools.

calbach: I defer to you and @zloery, I'm fine with either approach and don't have enough RBS experience to say whether changing v8 would work. My only request is either fix the comment or fix the value here before merging.

calbach (approved):

src/main/resources/config/resource_schema.yaml
jaycarlton: Went with the safe route, keeping the old key but deprecating it.

PR #229, opened 2022-03-22
[PF-1415] reactivate de-activated pools

jaycarlton:
As described in [this document](https://docs.google.com/document/d/1aXOF_0T_Qje4O5KbcKqgNPGhnytm97dPSBaGnzNtEUY/edit?usp=sharing&resourcekey=0-Vwd3zNcBNJ7mbbGq0OuflA), implementing pool re-activation after deactivation. Simply add the pool back into the schema file and it should clear the expiration and set the status back to ACTIVE. The scheduler should add flights to return the pool to its stated size.

### DIFF
--- a/src/test/java/bio/terra/buffer/service/pool/PoolServiceTest.java
+++ b/src/test/java/bio/terra/buffer/service/pool/PoolServiceTest.java
@@ -27,6 +27,7 @@ import bio.terra.buffer.generated.model.ResourceInfo;
 import bio.terra.common.exception.BadRequestException;
 import com.google.common.collect.ImmutableList;
 import java.time.Instant;
+import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


src/test/java/bio/terra/buffer/service/pool/PoolServiceTest.java
```diff
@@ -221,6 +222,85 @@ public void updateFromConfig_updateSize_updatePoolSizeSuccess() throws Exception
     assertEquals(bufferDao.retrievePools().get(0), createdPool.toBuilder().size(size + 10).build());
   }
 
+  @Test
```
melissachang, line 201: I like to have minimal test for testing the feature. Can you test one pool instead of two? It makes tests faster, don't have as many flakes to deal with, etc.

melissachang (approved):

src/test/java/bio/terra/buffer/service/pool/PoolServiceTest.java
jaycarlton: These are batch methods, so I wanted a plural number of pools to be sure it was plumbed correctly.

melissachang: Can you add a comment, that's why you want more than one pool? Thanks

jmczerk (approved):
LGTM

src/test/java/bio/terra/buffer/service/pool/PoolServiceTest.java
jaycarlton: done

PR #231, opened 2022-03-30
[RW-7885] Add AoU prod pool config with BQ quota enforcement

calbach:


calbach:
I just realized I should probably include perf here ... please hold

src/main/resources/config/prod/resource-config/vpc_sc_v8.yml
```diff
@@ -0,0 +1,40 @@
...
+    - "cloudtrace.googleapis.com"
+    - "containerregistry.googleapis.com"
+    - "dataflow.googleapis.com"
+    - "dataproc.googleapis.com"
+    - "dns.googleapis.com"
+    - "genomics.googleapis.com"
+    - "lifesciences.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "serviceusage.googleapis.com"
```
zloery, line 27: This API is new compared to v7, is it required for managing quota?

zloery (approved):

src/main/resources/config/prod/resource-config/vpc_sc_v8.yml
calbach: Yes - I noticed this was added in the dev PR. That API should be necessary for updating the BQ quota override when setting up projects, @jaycarlton could confirm this.

calbach:
(PR expanded to include perf as well, sorry @zloery your review was too fast)

zloery (approved):

src/main/resources/config/prod/resource-config/vpc_sc_v8.yml
jaycarlton: Yeah, that's why I added it.

src/main/resources/config/perf/pool_schema.yml
```diff
@@ -19,6 +19,9 @@ poolConfigs:
   - poolId: "vpc_sc_v6"
     size: 500
     resourceConfigName: "vpc_sc_v6"
+  - poolId: "vpc_sc_v7"
```
jaycarlton, line 22: Oh, I thought the convention was that you'd skip versions and this would be v9 to match `dev`. It's good to know that these numbers are local to the environment.

jaycarlton (commented):
You might want to remove v6 and v8 from the `dev` config while you're at it.

jaycarlton (approved):

calbach (commented):
I can send a followup PR for old pool removal. Looks like there's a bunch of stale ones here.

PR #232, opened 2022-03-30
Drop some unused AoU pools

calbach:
I left an extra 2 versions in prod to allow for potential rollback. I also noted some versions in alpha+staging, which are definitely unused by AoU. I'm not comfortable touching these without knowing their purpose, can check with @yonghaoy when he's back. By my estimate they could simply be deleted.

I'm also noticing that all historical pool config yaml files appear to be preserved, so I haven't touched those.


jaycarlton (approved):

zloery (approved):

yonghaoy:
> 

RE staging/alpha AoU pools: Rawls needs a dummy VPC-SC pool exists in each Terra env.

PR #233, opened 2022-03-31
POL-219 plug spring vulnerability

dvoet:
https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement

PR #234, opened 2022-04-12
[PF-1532] Update dependencies and gradle version

zloery:
This bumps the versions for a number of dependencies and also gradle (required by the newest Spring Boot plugin). The gradle changes (gradle-wrapper.jar, gradle-wrapper.properties, gradlew, gradlew.bat) were automatically generated with `gradle wrapper --gradle-version 7.4.1`. 

I've also removed two previous changes specific to older spring versions:

- Removed the `log4j` version pinning, the current version of Spring Boot already pulls in a newer version of log4j-api dependency
- Removed `SetDisallowedBinderFields` added in #233, this was a patch until we could upgrade to the latest version of Spring Boot.

Although it's not strictly necessary, I also bumped the versions of libraries we use for client tests to keep things consistent and made the client test depend on the client library project, rather than the published version.


src/test/java/bio/terra/buffer/common/BaseUnitTest.java
```diff
@@ -9,8 +9,8 @@
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 
 @Tag("unit")
-@ActiveProfiles({"test", "unit"})
+@ActiveProfiles({"test", "unit", "human-readable-logging"})
 @ExtendWith(SpringExtension.class)
 @ContextConfiguration(classes = Main.class)
-@SpringBootTest
+@SpringBootTest(properties = {"spring.cloud.gcp.credentials.location="})
```
yuhuyoyo, line 15: just curious what is this for?

yuhuyoyo (approved):

src/test/java/bio/terra/buffer/common/BaseUnitTest.java
zloery: I took this approach from WSM, but this intentionally nulls out GCP credentials for the test. 

I thought this fixed an error I was getting earlier, but it looks like it's actually not necessary. Removed

PR #239, opened 2022-04-19
[PF-1601] Downgrade kubernetes client via TCL

zloery:
The latest kubernetes client version is causing problems with the currently deployed kubernetes clusters, causing RBS to crashloop. This seems to be fixed by downgrading to 14.0.0, which is still late enough to have the security fixes that drove this update in the first place.

I tested this via developer environment, where it works as expected.


jmczerk (approved):

buffer-clienttests/build.gradle
```diff
@@ -20,7 +20,7 @@ dependencies {
         guava = "31.1-jre"
         jackson = "2.13.2"
         jersey = "2.35"
-        kubernetesClient = "15.0.0"
+        kubernetesClient = "14.0.0"
```
jaycarlton, line 23: Have you contacted the Kubernetes folks about the issue?

jaycarlton (approved):

PR #247, opened 2022-08-12
[PF-1539] Update JDK version to 17

dexamundsen:
- RBS (server only) Update JDK version to 17
- The client remains in version 11, and will be updated separately
- Update github workflows: skip tests when there are changes only in .md files
- Create and use a helper function to consistently use Millisecond precision for timestamps (change made to avoid issues with nanoS used by JDK 17 and MicroS used by postgres DB)

Ref - https://github.com/DataBiosphere/terra-workspace-manager/pull/644


Qi77Qi (approved):
👍

.java-version
```diff
@@ -1 +1 @@
-11.0
+17.0.4
```
ddietterich, line 1: I am curious about this. IIUC, this is a `jenv` artifact that will only work if the user has that tool installed. Is that right?

The consequence I have seen for me is that if I want to work on RBS, I would have to either install 17.0.4 or delete the file and set it to 17.0.2.

WSM does not have this set and for every new clone I make, I always have to set the project to 17. I wonder if we can find the recipe to get the JDK requirement setup without locking the version number.

ddietterich (approved):
LGTM - one question I'd be interested in understanding, but not blocking for the PR

.java-version
dexamundsen: Yes to both - with this file added, the user has to ensure the exact version is installed locally prior to any local development (updated this value in the existing file)

Removing it, as suggested and in line with WSM

PR #257, opened 2022-12-05
[PF-2203] Update dependencies, remove spring dependency management

zloery:
This PR updates a number of dependencies which have known security vulnerabilities. As part of doing this, I removed the Spring dependency management plugin, as it is not compatible with the gradle `constraints` block for versioning transitive dependencies (otherwise, the plugin will ignore constraints and decide versions on its own).


dexamundsen (approved):

ddietterich (approved):
LGTM

PR #265, opened 2023-04-21
[PF-2639] Add new pool for WSM with BigQuery Transfer API enabled

zloery:
Adds a new WSM pool with `bigquerydatatransfer.googleapis.com` enabled and no other changes over v9 for `tools` and `dev` environments, which are the only Broad environments which use WSM-managed GCP projects.

I'll remove the v9 pool in a few days after configuring all instances to use the v10 pool instead.


ddietterich (approved):
THanks!

yonghaoy (approved):

PR #270, opened 2023-06-22
[PF-2844] Add explicit validation for RANDOM_CHAR pattern

zloery:
Currently, RBS only validates prefix length for pool names using the `TWO_WORDS_NUMBER` scheme. However, it will also fail to produce valid names with the `RANDOM_CHAR` scheme if the prefix is too long (currently >22 characters). This adds explicit validation so that pool configuration fails earlier instead of each flight failing individually.

### DIFF
--- a/src/main/java/bio/terra/buffer/service/pool/GcpResourceConfigValidator.java
+++ b/src/main/java/bio/terra/buffer/service/pool/GcpResourceConfigValidator.java
@@ -1,10 +1,13 @@
 package bio.terra.buffer.service.pool;
 
+import static bio.terra.buffer.service.resource.projectid.GcpProjectIdGenerator.MAX_LENGTH_GCP_PROJECT_ID;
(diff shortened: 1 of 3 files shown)


yonghaoy (approved):

src/main/java/bio/terra/buffer/service/pool/GcpResourceConfigValidator.java
```diff
@@ -34,6 +37,13 @@ public void validate(ResourceConfig config) {
           String.format(
               "Project id prefix is too long for TWO_WORDS_NUMBER naming scheme: %s",
               config.getConfigName()));
+    } else if (gcpProjectConfig.getProjectIdSchema().getScheme().equals(SchemeEnum.RANDOM_CHAR)
```
yuhuyoyo, line 40: nit, just if is enough

yuhuyoyo: another nit, for enum, == is enough. and null safe :)

yuhuyoyo (approved):

PR #271, opened 2023-06-26
[PF-2872] Stop enabling genomic api

yonghaoy:
This API blocks RBS creating new project
`Bind permission denied for service: [genomics.googleapis.com](http://genomics.googleapis.com/)\nService [genomics.googleapis.com](http://genomics.googleapis.com/) is not available to this consumer.\nHelp Token`

Ideally, we are supposed to create a new pool with a new config. However, this PR is okay because:

- That API is not used anyway. Fine to have mixed settings.
- Creating projects with the old settings is broken.
- This is quicker as we do not need another change in Rawls/WSM/firecloud-devops

**Note: Forgot if RB blocks us from updating pool config in place**

### DIFF
--- a/src/main/resources/config/perf/pool_schema.yml
+++ b/src/main/resources/config/perf/pool_schema.yml
@@ -1,18 +1,15 @@
 # RBS Pools Schema for running buffer service performance test.
 ---
 poolConfigs:
-  - poolId: "cwb_ws_perf_v6"
+  - poolId: "cwb_ws_perf_v7"
     size: 500
-    resourceConfigName: "cwb_ws_perf_v6"
+    resourceConfigName: "cwb_ws_perf_v7"
   - poolId: "datarepo_v1"
     size: 300
     resourceConfigName: "datarepo_v1"
-  - poolId: "vpc_sc_v7"
+  - poolId: "vpc_sc_v9"
     size: 500
-    resourceConfigName: "vpc_sc_v7"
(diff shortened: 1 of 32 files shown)


src/main/resources/config/perf/pool_schema.yml
```diff
@@ -7,9 +7,6 @@ poolConfigs:
   - poolId: "datarepo_v1"
     size: 300
     resourceConfigName: "datarepo_v1"
-  - poolId: "vpc_sc_v7"
```
zloery, line 10: IMO, it's better to keep this PR very narrowly scoped to just removing `genomics.googleapis.com` since odds are high we'll need to hotfix it to various environments.

zloery (approved):

zloery:
I'd also like to get some confirmation from GCP that this API is intentionally no longer usable (vs. this just being a bug) before this gets to production

yonghaoy:
> 

I vaguely remember that this API is supposed to be removed after: https://github.com/DataBiosphere/terra-resource-buffer/pull/94

yonghaoy:
@zloery unfortunately, the previous one does not work because RB does not allow update pool config in place. 
Have to create new pool and upgrade others. I will prepare Rawls PR to bump pool version. 
can you help to bump WSM pool? 
Thanks

zloery (approved):

PR #275, opened 2023-06-28
[PF-2878] - Add new wsm pool with keepDefaultNetwork

rogerwangcs:
For dataproc clusters to properly create and run in gcp projects, we need to allow inter-vm ingress traffic for the cluster nodes to properly communicate with each other.

Gcp projects come with a default network which includes the default-allow-internal firewall rule that would configure the network to our needs, but currently RBS deletes that network and uses a custom network by default.

This PR will add a new wsm pool for `dev` and `tools` environments with the `keepDefaultNetwork` config to prevent the deletion and enable the default network in provisioned projects.

Followups:
https://github.com/DataBiosphere/terra-workspace-manager/pull/1361
https://github.com/DataBiosphere/terra-resource-buffer/pull/276

### DIFF
--- a/src/main/resources/config/dev/pool_schema.yml
+++ b/src/main/resources/config/dev/pool_schema.yml
@@ -16,3 +16,6 @@ poolConfigs:
   - poolId: "workspace_manager_v11"
     size: 500
     resourceConfigName: "workspace_manager_v11"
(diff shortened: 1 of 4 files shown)


ddietterich:
I don't think we should do this in the Broad environment without checking with their infosec people first.

PR #281, opened 2023-10-05
Create NAT gateway and router 

yuhuyoyo:


src/main/java/bio/terra/buffer/service/resource/flight/CreateRouterNatStep.java
```diff
@@ -0,0 +1,95 @@
...
+  @VisibleForTesting public static final String NAT_IP_ALLOCATION = "AUTO_ONLY";
+
+  public CreateRouterNatStep(GcpProjectConfig config, CloudComputeCow computeCow) {
+    gcpProjectConfig = config;
+    this.computeCow = computeCow;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    if (!blockBatchInternetAccess(gcpProjectConfig) || !enableNatGateway(gcpProjectConfig)) {
```
yonghaoy, line 52: what is the reason to skip this step when `!blockBatchInternetAccess(gcpProjectConfig)`?

yuhuyoyo: if there's no firewall rule to block internet access, we don't need nat

yonghaoy: My understanding before from reading https://docs.google.com/document/d/1-gWVpP1jG-iFnbwqowD8omM2o1bAhMx0NF2_dL0WuYM/edit
When create Mikey VMs, we will create GCE VMs not having public IP, the approach in my mind would be not set `networkInterfaces`. 

After reading your comment `if there's no firewall rule to block internet access, we don't need nat`, Mikey VMs can have public IP, it does not have because of blockBatchInternetAccess?

yonghaoy: > if there's no firewall rule to block internet access, we don't need nat
I was trying to say that even without that firewall rule, VM might still nee Nat if they don't have public IP.

yuhuyoyo: you are right. will remove

src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java
```diff
@@ -356,6 +364,37 @@ public void testCreateGoogleProject_blockedRegions_invalidBlockedRegion() throws
     assertNoSubnetsInBlockedRegions(project, ImmutableList.of(validBlockedRegion));
   }
 
+  @Test
+  public void testCreateGoogleProject_externalTrafficThroughNatWithBlockedRegion()
```
yonghaoy, line 368: My little concern is each new tests would take 15~20 minutes and I can not make them run in pararrel. So I have been very cautious adding new test. 
I would suggest to enable this new config into the existing test above, and drop this new test?

yonghaoy (approved):
LGTM as long as it is working.

src/test/java/bio/terra/buffer/integration/CreateProjectFlightIntegrationTest.java
yuhuyoyo: sure. even rmed another test.

src/main/java/bio/terra/buffer/service/resource/flight/CreateRouterNatStep.java
```diff
@@ -0,0 +1,109 @@
...
+              .setRegion(region)
+              .setNetwork(networkUri(projectId, NETWORK_NAME))
+              .setNats(List.of(natGateway));
+      try {
+        Optional<Operation> insertOperation =
+            createResourceAndIgnoreConflict(
+                () -> computeCow.routers().insert(projectId, region, router).execute());
+        if (insertOperation.isPresent()) {
+          LOGGER.info(
+              "start pulling insert router operation in project {} region {}", projectId, region);
```
jmczerk, line 87: Tiny nit: `s/pulling/polling`

jmczerk (approved):

src/main/java/bio/terra/buffer/service/resource/flight/CreateRouterNatStep.java
yuhuyoyo: i will sneak this in the next PR :)

PR #282, opened 2023-10-09
add firewall rule for ssh-through-iap

yuhuyoyo:


jmczerk (approved):

src/main/java/bio/terra/buffer/service/resource/flight/CreateFirewallRuleStep.java
```diff
@@ -151,6 +158,18 @@ public class CreateFirewallRuleStep implements Step {
               Arrays.asList(
                   new Firewall.Allowed().setIPProtocol("tcp").setPorts(Arrays.asList("443"))));
 
+  @VisibleForTesting
+  public static final Firewall ALLOW_INGRESS_SSH_THROUGH_IAP =
+      new Firewall()
+          .setName(ALLOW_SSH_THROUGH_IAP)
+          .setDirection("INGRESS")
+          // ip addresses that IAP uses for TCP forwarding
```
yonghaoy, line 166: nit: It's still a bit confusing and require a bit knowledge. Can we add more context here, e.g., add this link: 
https://cloud.google.com/iap/docs/using-tcp-forwarding#create-firewall-rule

yonghaoy (approved):

PR #284, opened 2023-11-07
Bump CRL and janitor versions

rtitle:
https://broadworkbench.atlassian.net/browse/TOAZ-336

I updated Janitor in https://github.com/DataBiosphere/terra-resource-janitor/pull/168 (removed some unused Azure fields). Updating CRL/Janitor dependency in buffer to get the new schema.

The CRL/Janitor changes are all Azure-related; so no logic changes needed to Buffer.


rtitle:
Looking into unit test issue..

src/main/resources/static/service_openapi.yaml
```diff
@@ -303,18 +303,6 @@ components:
           description: Name of the resource group.
           type: string
 
-    AzurePublicIp:
```
rtitle, line 306: I didn't realize Buffer duplicated these models from Janitor. 😅 

These types were removed in https://github.com/DataBiosphere/terra-resource-janitor/pull/168, so they need to be removed here as well. A few Azure resource types were added to Janitor as well, but I don't think Buffer needs to be aware of them (Janitor will handle `CloudResourceUid` messages from Buffer without all the Azure resources).

zloery (approved):

PR #285, opened 2023-11-07
WOR-1337 Spring3 etc

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-1337


zloery (approved):

aherbst-broad (commented):
Will we need a javax client for rawls, like we did for BPM?

dvoet:
> Will we need a javax client for rawls, like we did for BPM?

I am hoping not, this does not change so much so I don't think rawls will need to take a new version until it is ready for jakarta

aherbst-broad (approved):

PR #287, opened 2023-12-05
[IA-4525] Enable network API

jdcanas:
Some tests we are developing requires connectivity to VMs directly so we can reduce reliance on selenium in backend testing. This PR enables the google API required to tunnel to VMs protected by fire wall rules regardless of the source IP, which is needed since these tests will soon run on GHA nodes, which it is not secure to add to our VMs allow list. Google has an offering that supports this in a secure way, [IAP tunneling](https://cloud.google.com/iap/docs/using-tcp-forwarding#gcloud_2). Essentially, we will allow list a range of IPs reserved by google in the network that leonardo sets up for VMs. 

I have SSH working locally, since on the VPN my IP is on our VM allow list. However, I cannot use IAP proxying, and the error message points to enabling this API (here I am authenticated as the leo qa service account). 
image

Note that this change is only needed for projects created during automation tests, so we should not need to migrate existing projects. Automation tests create new projects in a bee, which all point to a central `tools` buffer instance as I understand it. 

I believe I've modified the proper config to have changes take effect on the buffer instance that serves bees in this PR, but I'm not positive.


src/main/resources/config/tools/resource-config/cwb_ws_resource_quality_v9.yml
```diff
@@ -0,0 +1,35 @@
...
+    - "containerregistry.googleapis.com"
+    - "dataflow.googleapis.com"
+    - "dataproc.googleapis.com"
+    - "dns.googleapis.com"
+    - "lifesciences.googleapis.com"
+    - "logging.googleapis.com"
+    - "monitoring.googleapis.com"
+    - "storage-api.googleapis.com"
+    - "storage-component.googleapis.com"
+    - "networkmanagement.googleapis.com"
```
jdcanas, line 28: this is the only thing changed besides the `configName` field

src/main/resources/config/tools/pool_schema.yml
```diff
@@ -4,9 +4,9 @@ poolConfigs:
   - poolId: "cwb_ws_fiab_v7"
     size: 500
     resourceConfigName: "cwb_ws_resource_fiab_v7"
-  - poolId: "cwb_ws_quality_v8"
+  - poolId: "cwb_ws_quality_v9"
     size: 5000
-    resourceConfigName: "cwb_ws_resource_quality_v8"
```
yonghaoy, line 9: Usually you can not do inplace update like this. As long as this PR released and v8 is deleted in this file. RBS will start deleting that pool. All clients that are currently using this pool will be broken. 
You need to check Rawls config see if it is using this pool and upgrade Rawls config.
Steps: 1 Add v9 pool. 
2: Update rawls config to use v9 pool
3: Delete v8 pool

yonghaoy (commented):
I am not familar with CWB test. I know there is `fiab` and quality. Do you need to update the other one?

src/main/resources/config/tools/pool_schema.yml
jdcanas: Which rawls config?

jdcanas:
cc @jack-r-warren , will merging this trigger a deploy to the `tools` resource buffer instance?

yonghaoy:
> I think this is right -- it seems BEEs use the quality resource pool in the tools Buffer instance:
> 
> https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/rawls/bee.yaml.gotmpl#L11
> 
> @yonghaoy will PR merges automatically deploy to tools?

`Tools` is auto deployed before, not sure anything changed today. Might need to ask @jack-r-warren ?

jack-r-warren:
@yonghaoy probably manual. For these sorts of questions please ask our team in #dsp-devops-champions

yonghaoy (approved):

PR #292, opened 2024-01-31
DC-814: Increase datarepo tools pool size

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-814
________
We've been seeing a lot of test flakiness around RBS over the last couple of weeks. After other debugging attempts (detailed in [datarepo PR](https://github.com/DataBiosphere/jade-data-repo/pull/1579)), I'd like to try bumping up the pool size. The recent errors have been due to not enough projects being available and the retries did not work, so potentially bumping up the pool size will provide some relief.

### DIFF
--- a/src/main/resources/config/tools/pool_schema.yml
+++ b/src/main/resources/config/tools/pool_schema.yml
@@ -11,7 +11,7 @@ poolConfigs:
     size: 5000
     resourceConfigName: "cwb_ws_resource_quality_v10"
   - poolId: "datarepo_v1"
-    size: 1000
+    size: 1500
     resourceConfigName: "datarepo_v1"
   - poolId: "vpc_sc_qa-fiab_v5"
     size: 100


rtitle (approved):

s-rubenstein (approved):

pshapiro4broad (approved):
This change seems harmless but it is puzzling to me that we'd need to change this now given that it hasn't changed in two years and I don't think our usage has changed recently. If this change doesn't fix the problem, is the plan to later revert it?

FYI it was last changed from 500 to 1000 in https://github.com/DataBiosphere/terra-resource-buffer/pull/240

snf2ye:
> This change seems harmless but it is puzzling to me that we'd need to change this now given that it hasn't changed in two years and I don't think our usage has changed recently. If this change doesn't fix the problem, is the plan to later revert it?

Since we are seeing some failures due to the lack of available projects, I'm not inclined to revert this. Even if it doesn't fix all of the flakiness, I think it should help. 
I agree, I don't understand why we're needing to increase this. My hope is that this either is the root issue OR that it alleviates some of the stress on the service and that we can more accurately pinpoint the real issue.

PR #295, opened 2024-05-08
[WOR-1668] Update pg jdbc driver to pull in cve fix

aherbst-broad:


build.gradle
```diff
@@ -59,7 +59,7 @@ repositories {
 
 dependencies {
     // Terra deps - we get Stairway via TCL
-    implementation group: 'bio.terra', name: 'terra-common-lib', version: '0.1.7-SNAPSHOT'
+    implementation group: 'bio.terra', name: 'terra-common-lib', version: '1.1.6-SNAPSHOT'
```
aherbst-broad, line 62: I needed to upgrade TCL to avoid a conflict with the PG driver version, which also required a bump in the spring boot version above.

cahrens (approved):

build.gradle
```diff
@@ -1,7 +1,7 @@
 // Build Script Classpath
 buildscript {
     ext {
-        springBootVersion = '3.1.2'
+        springBootVersion = '3.2.3'
```
okotsopoulos, line 4: 3.2.5 is the latest, would you like to bump to that one instead?

I'm not sure if this repo uses Spring dependency management plug-in, but 3.2.5 pulls in a non-vulnerable version of the Postgres driver, 42.6.2: https://github.com/spring-projects/spring-boot/blob/v3.2.5/spring-boot-project/spring-boot-dependencies/build.gradle#L1126

It's not the latest version, but I think it would eliminate the need to pin the version below.

build.gradle
okotsopoulos: 1.1.11 is the latest, would you like to bump to that one instead?  It addresses additional security vulnerabilities in the k8s client, not sure if that impacts this repo.

marctalbott (approved):

build.gradle
aherbst-broad: Good call, I'll try that out.

build.gradle
aherbst-broad: I'll give that a shot, thanks.

build.gradle
aherbst-broad: This is a bit confusing; when I pull in TCL 1.1.11 I get 

```
> Could not resolve all files for configuration ':testRuntimeClasspath'.
   > Could not find org.springframework.retry:spring-retry:.
     Required by:
         project : > bio.terra:terra-common-lib:1.1.11-SNAPSHOT:20240506.133839-1
```

Which is strange, since TCL has a dependency [setup](https://github.com/DataBiosphere/terra-common-lib/blob/develop/build.gradle#L53) for spring-retry. @okotsopoulos any thoughts? I might leave this alone for now until we can do a more thorough RBS cleanup next sprint.

build.gradle
aherbst-broad: Done.

build.gradle
okotsopoulos: My instinct is that TCL pulls in that dependency unversioned, with the thought that versioning will be determined by the caller's Spring dependency manager, a BOM, etc.

Tabling this for later makes a lot of sense.

PR #309, opened 2024-07-01
[WOR-1745] Remove lockfile for dependabot, and spotbugs because we use Sonarcloud.

cahrens:
Related to enabling dependabot.

We have removed Spotbugs in multiple other repos (WSM, LZS, etc.) because we have Sonar enabled.


aherbst-broad (commented):
The [README](https://github.com/DataBiosphere/terra-resource-buffer/blob/a776a7fd7a46153f46c605692f2c164e906af263/README.md#L197-L206) should get updated to remove mention of the lockfile.

aherbst-broad (approved):

blakery (approved):

PR #312, opened 2024-07-02
Switch to writing credentials from secrets for integration tests (WOR-1745).

cahrens:
As part of [WOR-1745](https://broadworkbench.atlassian.net/browse/WOR-1745), we need to change how the CI pipeline gets the SA credentials so that dependabot is able to run the tests (related PR is https://github.com/broadinstitute/terraform-ap-deployments/pull/1623).

[WOR-1745]: https://broadworkbench.atlassian.net/browse/WOR-1745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gradle/testing.gradle
+++ b/gradle/testing.gradle
@@ -6,7 +6,7 @@ test {
 
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 // This is the path to the default Google service account for the buffer service to run as.
-def googleCredentialsFile = "${projectDir}/src/test/resources/rendered/sa-account.json"
+def googleCredentialsFile = "${projectDir}/rendered/sa-account.json"
 tasks.withType(Test) {
     environment = [
             'GOOGLE_APPLICATION_CREDENTIALS': "${googleCredentialsFile}"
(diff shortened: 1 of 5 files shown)


gradle/testing.gradle
```diff
@@ -6,7 +6,7 @@ test {
 
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 // This is the path to the default Google service account for the buffer service to run as.
-def googleCredentialsFile = "${projectDir}/src/test/resources/rendered/sa-account.json"
+def googleCredentialsFile = "${projectDir}/rendered/sa-account.json"
```
cahrens, line 9: The `write-credential` action is not embedding the credential files within the test directory, which makes it different from what `local-dev/render-config.sh` does (and what `local-dev/run-local.sh` expects). However, I don't think that's a big deal and I'm not inclined to change how the service is run locally at this time. At some point we should likely follow the repo standards, which will involve creating a standard set of scripts, and we can unify the experience at that time.

FWIW, I did verify that rendering credentials and running the service locally (following the directions in the README) does still work.

blakery (approved):
I think it's fine to merge this with only one approval, since this is only test changes (since everyone else is out)

PR #353, opened 2024-09-26
[WOR-1816] Scheduled DB cleanup job

aherbst-broad:
## Context
The `resource` and `cleanup_record` tables are only ever appended to, and grow without bound. This can cause performance issues as the DB eventually struggles to manage these ever-larger tables.

## This PR
* I've introduced a scheduled job that will find all records in these tables and delete those that are eligible for cleanup, up to a limit. This job is scheduled via Spring's [scheduling](https://spring.io/guides/gs/scheduling-tasks) mechanism. This introduces a divergence from the current pattern in the codebase of [using](https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/service/cleanup/CleanupScheduler.java#L43) a [ScheduledExecutorService](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ScheduledExecutorService.html). I've opted for the spring mechanism primarily for ease-of-use and a more grok-able cron scheduling configuration that plays well with spring's configuration framework. 
    * The `cleanup_record` table does not have any usable timestamps to determine when a record is eligible for cleanup. So, I've added one in [this](https://github.com/DataBiosphere/terra-resource-buffer/pull/353/files#diff-1f5abc82d32af500cac1b822676a7d4755fa2a17ab71aaec487add2d8bbaa9a5) changeset.
* Since the scheduled tasks will run on any node in a k8s cluster, a locking mechanism is required to prevent unwanted concurrent execution of this job. I've leveraged [Shedlock](https://github.com/lukas-krecan/ShedLock/tree/master) to manage this and ensure at most once execution per scheduled tick. 
   * Shedlock requires a table to track locking, which I've added in [this](https://github.com/DataBiosphere/terra-resource-buffer/pull/353/files#diff-c019a2563fe7b3aa6d6fb66aed8f3ffbde2dccc5c93a562efef59b9dac9a6e4a) changeset.
 * The job is configured [off by default](https://github.com/DataBiosphere/terra-resource-buffer/blob/536ef1a0fdbf685dba3007578f7cf1df08a56356/src/main/resources/application.yml#L10), it will need to be enabled via helmfile env var changes as it is rolled out per environment. 
 * I also ran spotless over the codebase to get formatting in line, there are a handful of changes that resulted from that.

### DIFF
--- a/src/main/java/bio/terra/buffer/service/cleanup/JanitorResourceCleanupScheduler.java
+++ b/src/main/java/bio/terra/buffer/service/cleanup/JanitorResourceCleanupScheduler.java
@@ -33,11 +33,11 @@ import org.springframework.stereotype.Component;
 
 /** Scheduler service to publish message to Janitor to cleanup resource. */
 @Component
-public class CleanupScheduler {
+public class JanitorResourceCleanupScheduler {
   // Number of message to publish per scheduler run.
   private static final Integer MESSAGE_TO_PUBLISH_PER_RUN = 100;
 
-  private final Logger logger = LoggerFactory.getLogger(CleanupScheduler.class);
+  private final Logger logger = LoggerFactory.getLogger(JanitorResourceCleanupScheduler.class);
 
   /** Only need as many threads as we have scheduled tasks. */
   private final ScheduledExecutorService executor = new ScheduledThreadPoolExecutor(1);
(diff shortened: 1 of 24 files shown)


src/main/java/bio/terra/buffer/service/cleanup/JanitorResourceCleanupScheduler.java
```diff
@@ -33,11 +33,11 @@
 
 /** Scheduler service to publish message to Janitor to cleanup resource. */
 @Component
-public class CleanupScheduler {
+public class JanitorResourceCleanupScheduler {
```
aherbst-broad, line 36: I've renamed this class to make it more obvious that it deals with scheduling Janitor cleanups of cloud resources.

marctalbott (approved):

PR #365, opened 2024-10-22
CORE-3: upgrade to Postgres 16.3 for unit tests

davidangb:
For unit tests, and unit tests only, upgrade to Postgres 16.3.

16.3 is the version supported by Google: https://cloud.google.com/sql/docs/postgres/db-versions

This change allows us to get some burn-in on Postgres 16 before upgrading any live environments.

_reviewer: the Codacy failures on this PR sure look like false positives to me._

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:12.3
+        image: postgres:16.3
         env:
           POSTGRES_PASSWORD: postgres
         options: >-
--- a/local-dev/local-postgres-init.sql
+++ b/local-dev/local-postgres-init.sql
@@ -3,5 +3,8 @@ CREATE DATABASE testdb_stairway;
 CREATE ROLE dbuser WITH LOGIN ENCRYPTED PASSWORD 'dbpwd';
 CREATE ROLE dbuser_stairway WITH LOGIN ENCRYPTED PASSWORD 'dbpwd_stairway';
 
-ALTER DATABASE testdb OWNER TO dbuser;
-ALTER DATABASE testdb_stairway OWNER TO dbuser_stairway;
+GRANT USAGE ON SCHEMA testdb.public TO dbuser;
+GRANT CREATE ON SCHEMA testdb.public TO dbuser;
+
+GRANT USAGE ON SCHEMA testdb_stairway.public TO dbuser_stairway;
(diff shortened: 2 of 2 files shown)


local-dev/local-postgres-init.sql
```diff
@@ -1,4 +1,5 @@
-CREATE DATABASE testdb;
-CREATE DATABASE testdb_stairway;
 CREATE ROLE dbuser WITH LOGIN ENCRYPTED PASSWORD 'dbpwd';
 CREATE ROLE dbuser_stairway WITH LOGIN ENCRYPTED PASSWORD 'dbpwd_stairway';
+
+CREATE DATABASE testdb OWNER dbuser;
+CREATE DATABASE testdb_stairway OWNER dbuser_stairway;
```
davidangb, line 5: this change is necessary due to permission lockdown as of Postgres 15. https://www.percona.com/blog/public-schema-security-upgrade-in-postgresql-15/ is a good writeup of this change. Without the fix, Liquibase fails to create tables in these dbs, leading to widespread failures of everything else.

This only affects unit tests, which use a blank/empty database each time. In live environments, when we upgrade from Postgres 12 to 16, permissions should be preserved:
> Upgrading a cluster or restoring a database dump will preserve public's existing permissions.

https://www.postgresql.org/docs/release/15.0/

.github/workflows/test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:12.3
+        image: postgres:16.3
```
davidangb, line 24: 16.3 is the version supplied by Google for CloudSQL: https://cloud.google.com/sql/docs/postgres/db-versions

local-dev/local-postgres-init.sql
davidangb, line 5: this change is necessary due to permission lockdown as of Postgres 15. https://www.percona.com/blog/public-schema-security-upgrade-in-postgresql-15/ is a good writeup of this change. Without the fix, Liquibase fails to create tables in these dbs, leading to widespread failures of everything else.

This only affects unit tests, which use a blank/empty database each time. In live environments, when we upgrade from Postgres 12 to 16, permissions should be preserved:
> Upgrading a cluster or restoring a database dump will preserve public's existing permissions.

https://www.postgresql.org/docs/release/15.0/

dvoet (approved):

kevinmarete (approved):

PR #367, opened 2024-10-28
AN-186 Add pool configs with Google Batch enabled

jgainerdewar:
In support of migrating workflow compute from LifesciencesAPI to the new Batch API, create new pools with the Batch API enabled.

### DIFF
--- a/src/main/resources/config/toolsalpha/resource-config/resource_toolsalpha_v9.yml
+++ b/src/main/resources/config/toolsalpha/resource-config/resource_toolsalpha_v9.yml
@@ -1,0 +1,41 @@
+# Resource template for local testing and personal environment on GKE
+---
+configName: "resource_toolsalpha_v9"
+gcpProjectConfig:
+  projectIdSchema:
+    prefix: "terra-toolsalpha"
+    scheme: "RANDOM_CHAR"
+  # RBS Testing
+  parentFolderId: "637867149294"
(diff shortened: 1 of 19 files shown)


src/main/resources/config/toolsalpha/resource-config/resource_toolsalpha_v9.yml
```diff
@@ -0,0 +1,41 @@
+# Resource template for local testing and personal environment on GKE
+---
+configName: "resource_toolsalpha_v9"
```
aednichols, line 3: Isn't Alpha decomissioned?

jgainerdewar: Yes, and I didn't update the `alpha` (or `perf`) environments here, but I wasn't sure whether this one referenced our alpha environment specifically or some other generic "alpha."

aednichols (approved):

src/main/resources/config/toolsalpha/resource-config/resource_toolsalpha_v9.yml
marctalbott: I'm not sure what the `toolsalpha` environment is to be completely honest, but it's not worth worrying about in this PR. The Core Services team can do an audit of the pools another time

marctalbott (approved):

PR #293, opened 2024-04-30
[DCJ-284] Increase tools pool datarepo_v1 from 1500->2000

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DCJ-284

**Background**

TDR uses the `datarepo_v1` pool in RBS' tools namespace in our integration tests.  We periodically run into pool exhaustion when many test runs are running concurrently.   The configuration change we made to cancel earlier active test runs on a PR helps manage this load, but isn’t enough given the high volume of developer activity on TDR these days.

RBS logs showed a spike in related errors yesterday, aligning with many test runs that ultimately failed (we didn’t yet have pool availability metrics exposed in Grafana): https://cloudlogging.app.goo.gl/Lvwz3ytmAb1Vwx3T8

**Changes**

- Increase `datarepo_v1` pool in RBS' tools namespace from 1500 -> 2000
- Include a link to Grafana panel showing READY resource ratios for tools pools, which is getting populated as of this morning.
  - When deciding whether to modify our pool size in the past, we've wanted these metrics to be readily available to inform our decision-making.

PR #374, opened 2024-11-04
DDO-3917-scale-down-buffer-pools

jyang-broad:
Looking at the last few weeks of Buffer activity, it looks like we're quite over provisioned on projects, while technically we're not paying for these, it looks like the number of projects is breaking the billing apis via timeout and thus our ability to track costs across the org.

This PR removes what looks like unused pools as well as scales down the big used pools to ~50% of their peak usage.

https://grafana.dsp-devops.broadinstitute.org/d/dcvlwY2nz/resource-buffer-service?orgId=1&var-cluster=terra-integration&var-namespace=terra-tools&var-project=&var-container=All&viewPanel=4&refresh=10s&from=now-30d&to=now


src/main/resources/config/tools/pool_schema.yml
```diff
@@ -2,30 +2,15 @@
 # READY resource ratio by pool over time: https://grafana.dsp-devops.broadinstitute.org/goto/qHOGMufSk?orgId=1
 ---
 poolConfigs:
-  - poolId: "cwb_ws_fiab_v7"
-    size: 500
-    resourceConfigName: "cwb_ws_resource_fiab_v7"
-  - poolId: "cwb_ws_fiab_v8"
```
calypsomatic, line 8: `cwb_ws_fiab_v8` was just added a few days ago - seems like we might want to keep it around?

src/main/resources/config/tools/pool_schema.yml
```diff
@@ -2,30 +2,15 @@
...
+    size: 1000
     resourceConfigName: "cwb_ws_resource_quality_v11"
   - poolId: "datarepo_v1"
-    size: 2000
+    size: 1000
     resourceConfigName: "datarepo_v1"
-  - poolId: "vpc_sc_qa-fiab_v5"
-    size: 100
-    resourceConfigName: "vpc_sc_qa-fiab_v5"
-  - poolId: "vpc_sc_qa-fiab_v6"
```
calypsomatic, line 26: `vpc_sc_qa-fiab_v6` was also just added a few days ago

src/main/resources/config/tools/pool_schema.yml
pshapiro4broad: Seems likely, the author was @jgainerdewar in PR https://github.com/DataBiosphere/terra-resource-buffer/pull/367

src/main/resources/config/tools/pool_schema.yml
```diff
@@ -2,30 +2,15 @@
...
 poolConfigs:
-  - poolId: "cwb_ws_fiab_v7"
-    size: 500
-    resourceConfigName: "cwb_ws_resource_fiab_v7"
-  - poolId: "cwb_ws_fiab_v8"
-    size: 500
-    resourceConfigName: "cwb_ws_resource_fiab_v8"
-  - poolId: "cwb_ws_quality_v8"
-    size: 5000
-    resourceConfigName: "cwb_ws_resource_quality_v8"
```
pshapiro4broad, line 13: I think each `resourceConfigName` corresponds to a file, should those be deleted too? https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/config/tools/resource-config/cwb_ws_resource_quality_v8.yml

src/main/resources/config/tools/pool_schema.yml
```diff
@@ -2,30 +2,15 @@
...
-    size: 5000
+    size: 1000
     resourceConfigName: "cwb_ws_resource_quality_v10"
   - poolId: "cwb_ws_quality_v11"
-    size: 5000
+    size: 1000
     resourceConfigName: "cwb_ws_resource_quality_v11"
   - poolId: "datarepo_v1"
-    size: 2000
+    size: 1000
```
pshapiro4broad, line 12: This was (somewhat) recently increased from 1500 to 2000. Before changing this to a lower number, we should understand why this was changed in the first place. The PR that changed it had a lot of details: https://github.com/DataBiosphere/terra-resource-buffer/pull/293

jyang-broad: It was mentioned last time that there was movement on the TDR side to remove/deprecate/update the tests that tend to consume a lot of projects, that and the PR volume is down. 

@snf2ye are you able to comment on if this number seems safe?

src/main/resources/config/tools/pool_schema.yml
jyang-broad: added this one back

src/main/resources/config/tools/pool_schema.yml
jyang-broad: added this one back, sorry! got fiab v8 confused w/ quality v8 (which _does_ appear to be unused today)

src/main/resources/config/tools/pool_schema.yml
jyang-broad: I went ahead and removed all the config files, checking around to make sure this is a safe-ish move though.

jyang-broad:
hey folks! can I get some thumbs to push this one through?

pshapiro4broad (approved):
This looks OK.

Is there a way to verify that the schema files are correct, maybe by deploying to a bee?

jyang-broad:
> This looks OK.
> 
> Is there a way to verify that the schema files are correct, maybe by deploying to a bee?

Unfortunately, Buffer never got ported to Bees (we would have had to solve of workspace pooling across multiple buffers as well as buffer orphaned workspaces). I haven't contributed to this repo before, but I expect we can just roll it back

jgainerdewar (approved):

jyang-broad:
just realized I might have answered a different question than you were asking, (do the new pools work vs will buffer crash) @jgainerdewar mentioned she just tested the new pools in a bee

PR #376, opened 2024-11-06
AN-186 Update test to use new pool config

jgainerdewar:
Update test to use pool created in #367


trholdridge (approved):

jgainerdewar:
Will incorporate this change into a different PR that removes the old pools.

PR #377, opened 2024-11-07
[AN-181] Give default compute SA the secretAccessor role

lucymcnatt:
https://broadworkbench.atlassian.net/browse/AN-181

### DIFF
--- a/src/main/java/bio/terra/buffer/service/resource/flight/CreateGkeDefaultSAStep.java
+++ b/src/main/java/bio/terra/buffer/service/resource/flight/CreateGkeDefaultSAStep.java
@@ -39,7 +39,8 @@ public class CreateGkeDefaultSAStep implements Step {
           "roles/logging.logWriter",
           "roles/monitoring.metricWriter",
           "roles/monitoring.viewer",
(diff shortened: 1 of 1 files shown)


davidangb (approved):
+1 for code syntax.

I assume you've run this through appsec and everyone agrees this is a Good Idea?

lucymcnatt:
> I assume you've run this through appsec and everyone agrees this is a Good Idea?

We've submitted a 1-pager about these changes and gotten preliminary approval, but will hold off on merging until there's a final decision
https://broadinstitute.slack.com/archives/CADU7L0SZ/p1730223296197239

marctalbott (commented):
This will only give the SA the secretAccessor role in new Google projects that RBS creates. To make sure all new workspaces get this change, you'll need to set up new project pools

lucymcnatt:
After discussing with the team, we've decided to go ahead and merge this without creating new project pool;
- this change will be propagated to prod with some other Batch changes next week
- this role is only necessary for accessing private dockerhub images with the new batch backend so not something we need urgently in the lower environments.

PR #378, opened 2024-11-07
AN-186 Clean up pools from Google Batch enablement

jgainerdewar:
This is a follow-on to #367 now that the non-prod pools there have been (mostly) put into use.
 * Update test utils to use new pool version
 * Remove the old versions of pools from `pool_schema.yml` files
 * Remove ALL `resource-config` files for unused pools

Questions for reviewers:
 * Do we want to remove `resource-config` files for older pools? Is this desirable cleanup or am I deleting a useful record?
 * In #367 I added new versions of all non-TDR pools in environments I didn't know to be defunct (perf, alpha), because this change should apply to ALL new Terra workspace Google projects. In a couple cases, though, I couldn't find where the pool was used, so I couldn't update that usage to the new pool version. This leaves us with a newer pool with the correct settings, and an older pool that may or may not be in use somewhere, and so can't be removed. What's the best way to handle this? The two pools in this state are `tools/cwb_ws_fiab_v7` and `toolsalpha/resource_toolsalpha_v8`.


dvoet:
I can't find any reference to `tools/cwb_ws_fiab_v7`, or v8 for that matter. Maybe their use died with FIABs. The [comment here](https://github.com/DataBiosphere/terra-resource-buffer/blob/7f8dd0004bfb9c56ba98efa90be7f8a50cc638eb/src/main/resources/config/toolsalpha/pool_schema.yml#L1) says `resource_toolsalpha` is for local testing. Maybe you want another PR for it, but I think you can drop cwb_ws_fiab all together and drop the old version of `resource_toolsalpha`

dvoet (approved):

PR #402, opened 2025-01-09
[DDO-3972] Stop sending perf test notifications to #dsde-qa

choover-broad:
It looks like a recent PR (#398) had the side effect of reenabling Buffer's nightly perf tests, but they no longer work. I'd like to remove notifications to the #dsde-qa channel since it is used for monolith release triage and these are not release-blocking. Notifications will continue to go to other configured Slack channels.


marctalbott (approved):

marctalbott:
I went ahead and disabled the workflow (didn't realize it had started up again so thanks for the head's up!).

This is still worth merging IMO in case we accidentally re-enable it in the future.

PR #446, opened 2025-07-23
[CORE-553] Repair Google Project

samanehsan:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-553

**Background:**
If a Google project has an inactive billing account (e.g. the credit card expired), google will disable APIs. So if this project was associated with a Terra workspace and a user fixes the billing, it may not be enough to get everything working again. 

**Changes:**
Add a new endpoint to RBS that finds the pool that the google project was created in, and then re-enables all of the APIs that it originally set up. The endpoint starts a Stairway job, so this PR also adds endpoints to get a job, and get a job result so that the caller has a way to tell whether or not the operation was successful. 

Screenshot 2025-07-23 at 3 45 10 PM

**Testing:**
1. Change the buffertest pool config to have 1 project (default is 0)
2. Run RBS locally and wait for project creation to finish
4. Request a project
5. View project in GCP console and disable some APIs (I disabled Storage, Cloud DNS and Container Registry API)
6. Call the repair API endpoint on the project (returns a job id)
7. View project in GCP console and confirm that APIs were re-enabled
8. Get job --> status should say "succeeded"
9. Get job result --> returns 200 Ok (no info gets returned)


src/main/java/bio/terra/buffer/service/resource/flight/EnableServicesStep.java
```diff
@@ -36,6 +37,9 @@ public StepResult doStep(FlightContext flightContext) throws RetryException {
       return StepResult.getStepResultSuccess();
     }
     String projectId = flightContext.getWorkingMap().get(GOOGLE_PROJECT_ID, String.class);
+    if (projectId == null) {
```
samanehsan, line 40: The google project id is passed as an input to the repair flight, but the create project flight puts it in the working map.

src/main/resources/static/service_openapi.yaml
```diff
@@ -352,6 +474,48 @@ components:
         azureVirtualMachine:
           $ref: '#/components/schemas/AzureVirtualMachine'
 
+    JobModel:
```
samanehsan, line 477: Copied from TDR

src/test/java/bio/terra/buffer/service/job/JobServiceTest.java
```diff
@@ -0,0 +1,135 @@
...
+import org.mockito.Mockito;
+import org.springframework.http.HttpStatus;
+
+import java.time.Instant;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.*;
+
+class JobServiceTest {
```
samanehsan, line 23: Copilot-generated tests :octocat: I reviewed and added two more (`testRetrieveJobInterruptedException` and `testRetrieveJobResultFailedWithException`)

src/main/java/bio/terra/buffer/app/controller/ResourceApiController.java
```diff
@@ -0,0 +1,38 @@
...
+    ResourceApiController(PoolService poolService, FlightScheduler flightScheduler, JobService jobService) {
+        this.poolService = poolService;
+        this.flightScheduler = flightScheduler;
+        this.jobService = jobService;
+    }
+
+    @Override
+    public ResponseEntity<JobModel> repairResource(String projectId) {
+        GoogleProjectUid googleProjectUid = new GoogleProjectUid().projectId(projectId);
+        Pool pool = poolService.getPoolForGoogleProject(googleProjectUid);
```
samanehsan, line 34: I decided to get the pool outside of the flight so that we can "fail fast" if the `googleProjectUid` doesn't exist. But if it's clearer to put everything inside of the flight I can also make that change.

src/main/java/bio/terra/buffer/service/pool/PoolService.java
```diff
@@ -103,6 +101,17 @@ public PoolInfo getPoolInfo(PoolId poolId) {
...
 
+  public Pool getPoolForGoogleProject(GoogleProjectUid projectId) {
+    CloudResourceUid cloudResourceUid = new CloudResourceUid().googleProjectUid(projectId);
+    Optional<Resource> resource = bufferDao.retrieveResource(cloudResourceUid);
+    if (resource.isEmpty()) {
+      throw new NotFoundException(String.format("Resource id does not exist: %s.", projectId));
+    }
+    PoolId poolId = resource.get().poolId();
+    return bufferDao.retrievePool(poolId)
+            .orElseThrow(() -> new NotFoundException(String.format("Pool for this resource does not exist: %s.", poolId)));
```
samanehsan, line 112: Not sure if it's worth separating these different queries/error messages...I could consolidate this into one db query

davidangb: I like having these as separate errors!

Given that we have cleaned up old pools, e.g. #443 ... it seems like we could hit this error quite often. Does this hint that we should NOT clean up old pools but instead set them to `size: 0`? Should we try to restore the configs for older pools?

kevinmarete (approved):
Other than a small nit comment and failing test, looks good to me!

PR #450, opened 2025-08-06
[CORE-553] Add empty request body to repair POST endpoint

samanehsan:


calypsomatic (approved):
This is fine, but if it doesn't need a body should it be a POST?

samanehsan:
@calypsomatic it's definitely weird to have an empty post request body, but to me it doesn't makes sense as a GET endpoint either 😕.  One alternative is to put the googleProjectId into the post body and have no path parameters, which is a slightly bigger change but not difficult to do.

calypsomatic:
> @calypsomatic it's definitely weird to have an empty post request body, but to me it doesn't makes sense as a GET endpoint either 😕. One alternative is to put the googleProjectId into the post body and have no path parameters, which is a slightly bigger change but not difficult to do.

Also seems weird to put the id into the postbody when it doesn't need to be.  🤷 May as well leave it as is

PR #460, opened 2025-09-30
CORE-69: ignore jackson 2.19.2 for buildScript

davidangb:
Jackson 2.19.2 is only specified for the `buildScript` classpath.
As of this writing, Jib requires this specific version, and we want Dependabot to leave it alone.

This PR adds an ignore clause to `.github/dependabot.yml` to ignore the 2.20.0 update, which causes problems seen in #461. This will only be effective until jackson releases a higher version. I am struggling to come up with a better way to handle NOT updating the Jib-specific dependency.

### DIFF
--- a/buffer-clienttests/build.gradle
+++ b/buffer-clienttests/build.gradle
@@ -28,7 +28,7 @@ dependencies {
         googleOauth2 = "1.39.1"
 
         bufferServiceClient = "0.4.3-SNAPSHOT"
-        testRunner = "0.2.1-SNAPSHOT"
+        testRunner = "0.2.2-SNAPSHOT"
     }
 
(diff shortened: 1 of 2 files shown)


buffer-clienttests/build.gradle
```diff
@@ -28,7 +28,7 @@ dependencies {
         googleOauth2 = "1.39.1"
 
         bufferServiceClient = "0.4.3-SNAPSHOT"
-        testRunner = "0.2.1-SNAPSHOT"
+        testRunner = "0.2.2-SNAPSHOT"
```
davidangb, line 31: 0.2.1 is not available in GAR, but 0.2.2 is.

calypsomatic (approved):

samanehsan (approved):

PR #487, opened 2026-03-16
PROD-1042: Upgrade to latest TCL

jgainerdewar:
RBS is currently unable to deploy because of a mismatch between the Java client k8s version and the k8s cluster version, which upgraded over the weekend. This should resolve it by pulling in a version of TCL that has the upgraded java client for k8s.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ repositories {
 
 dependencies {
     // Terra deps - we get Stairway via TCL
-    implementation group: 'bio.terra', name: 'terra-common-lib', version: '1.1.69-SNAPSHOT'
+    implementation group: 'bio.terra', name: 'terra-common-lib', version: '1.1.70-SNAPSHOT'
     implementation(group: 'bio.terra', name: 'terra-cloud-resource-lib', version: '1.2.42-SNAPSHOT') {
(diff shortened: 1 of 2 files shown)


JoshEvans33 (approved):

snf2ye (approved):

marctalbott (approved):

jgainerdewar:
Why the new nullable check?

The k8s client 25.0.0 bumped OkHttp from 4.12.0 to 5.1.0, and since Gradle resolves to the highest version, that upgrade wins across the entire test classpath — including for the
  terra-cloud-resource-lib (CRL) which uses OkHttp 4.10.0.

  OkHttp 5.x is a significant upgrade. It could change how HTTP responses are read/buffered, which could affect how Google's AbstractGoogleJsonClient parses JSON responses. If a response body is read slightly
  differently (e.g., an empty JSON array [] getting dropped vs preserved), getAccounts() could return null where it previously returned an empty list.

  Summary: The k8s client jump from 23 to 25 dragged OkHttp from 4.x to 5.x across the whole test runtime. That likely changed the HTTP response handling for the IAM client, causing getAccounts() to return null
  instead of an empty list in certain cases.

